PR #1, opened 2025-09-08
feat/LEI-20: integrate the HL7.FHIRPath.Validator

doug-helios:


.github/workflows/ci.yml
```diff
@@ -152,6 +152,159 @@ jobs:
           fail_ci_if_error: true
           token: ${{ secrets.CODECOV_TOKEN }}
 
+  fhirpath-validation:
+    name: FhirPath Validation
+    runs-on: [self-hosted, Linux]
+    needs: [test, lint]
```
smunini, line 158: I think we can safely remove this needs: line - It's ok if the fhirpath-validation runs in early parallel to the other tests.

PR #61, opened 2026-04-10
feat(subscriptions): implement FHIR topic-based Subscriptions — Phase 1   (rest-hook + core infrastructure)

aacruzgon:
## Summary

  Implements Phase 1 of the FHIR topic-based Subscriptions Framework as
  described in [Discussion
  #59](https://github.com/HeliosSoftware/hfs/discussions/59).

  This introduces a new `helios-subscriptions` crate and integrates it into the
  REST layer as an optional feature. The engine fires asynchronously after every
  resource write (fire-and-forget via `tokio::spawn`), mirroring the existing
  audit middleware pattern.

  R4 and R4B use the [Subscriptions R5 Backport
  IG](https://build.fhir.org/ig/HL7/fhir-subscription-backport-ig/);
  R5 and R6 use the native Subscriptions Framework.

  ---

  ## What's in this PR

  ### New crate: `helios-subscriptions`

  The engine decomposes into five concerns:

  **Topic Registry** (`src/topics/`)
  - `InMemoryTopicRegistry` backed by `RwLock<HashMap>` for concurrent read
  access
  - Parses `SubscriptionTopic` JSON (R5/R6 native) and `Basic` resources with
  backport extensions (R4)
  - Matches resource write events against topics by resource type and
  interaction type

  **Subscription Manager** (`src/manager/`)
  - `SubscriptionManager` backed by `DashMap<(tenant_id, subscription_id),
  ActiveSubscription>` for lock-free concurrent access
  - `SubscriptionStatusCode` state machine: `Requested → Active → Error → Off`
  with validated transitions
  - Version-aware JSON parsing: R4 reads the backport extension set
  (`backport-topic-canonical`, `backport-payload-content`,
  `backport-channel-type`, `backport-filter-criteria`); R5/R6 read native
  `topic`, `channelType`, `filterBy` fields

  **Event Evaluator** (`src/evaluator/`)
  - Matches `ResourceEvent` against active subscriptions via topic triggers and
  subscription filter criteria
  - Filter matching supports `code`, `category`, `patient`/`subject`,
  `identifier`, and direct field lookup
  - All maps are keyed by `(tenant_id, ...)` — subscriptions in different
  tenants never interact

  **Notification Builder** (`src/notification/`)
  - R4: `Bundle(type=history)` with `SubscriptionStatus` as a `Parameters`
  resource (backport IG profile)
  - R4B/R5/R6: `Bundle` with native `SubscriptionStatus` resource; R5/R6 use
  `type=subscription-notification`
[...]

PR #62, opened 2026-04-12
feat(subscriptions): implement WebSocket channel — Phase 2

aacruzgon:
## Summary

Implements Phase 2 of the FHIR Topic-Based Subscriptions roadmap: the **WebSocket channel**. Builds on top of the Phase 1 rest-hook infrastructure without any breaking changes.

Clients subscribe with `channel.type = "websocket"`, obtain a short-lived binding token from `$get-ws-binding-token`, connect to `/ws/subscriptions/bind?token=<token>`, and receive notification bundles as JSON text frames. The protocol is unidirectional (server → client).

Related: discussion #59, Phase 1 PR #61.

### `helios-subscriptions` crate

**New files:**
- `src/channels/ws_token.rs` — `WsBindingTokenManager`: generates UUID v4 tokens with configurable expiry (default 30 s), single-use via atomic `DashMap::remove()`
- `src/channels/ws_manager.rs` — `WebSocketManager`: tracks `(tenant_id, subscription_id) → Vec<(client_id, UnboundedSender)>`; broadcasts notifications, prunes closed channels, removes all clients on subscription deletion
- `src/channels/websocket.rs` — `WebSocketChannel` implementing `ChannelDispatcher`: dispatches by broadcasting to connected clients (best-effort, always `Success`); handshake is a no-op (activation is immediate)
- `tests/websocket_integration.rs` — 8 integration tests: activation, delivery to single/multiple clients, zero-client no-panic, disconnected-client cleanup, binding token lifecycle, subscription-delete closes clients, tenant isolation

**Modified files:**
- `src/channels/mod.rs` — declare the three new modules
- `src/config.rs` — add `ws_token_lifetime_secs: i64` (default `30`)
- `src/engine/mod.rs` — add `ws_manager`, `ws_channel`, `ws_token_manager` fields; expose `ws_manager()` / `ws_token_manager()` accessors; add `ChannelType::Websocket` arms to `activate_subscription()` and `dispatch_with_retry()`; call `ws_manager.remove_all_clients()` when a Subscription is deleted
- `src/lib.rs` — re-export `WebSocketManager` and `WsBindingTokenManager` for use by the REST layer

### `helios-rest` crate

**New files:**
- `src/handlers/ws.rs` — `ws_bind_handler`: validates the binding token, upgrades the HTTP connection, sends the handshake bundle as the first frame, then loops `tokio::select!` over the notification channel and the socket (for close detection); cleans up via `ws_manager.remove_client()` on exit

**Modified files:**
- `Cargo.toml` — add `"ws"` to axum features (pulls in `tokio-tungstenite` automatically)
- `src/handlers/subscriptions.rs` — add `get_ws_binding_token_handler`: validates channel type is `websocket`, generates token, returns `Parameters` with `token` / `expiration` / `websocket-url`
- `src/handlers/mod.rs` — register `ws` module behind `#[cfg(feature = "subscriptions")]`
- `src/routing/fhir_routes.rs` — add `/{resource_type}/{id}/$get-ws-binding-token` and `/ws/subscriptions/bind` routes
- `src/lib.rs` — add `"websocket"` to `supported_channel_types` in the engine initializer

## Design decisions

**WebSocket dispatch is best-effort.** `dispatch()` always returns `DispatchResult::Success` even when zero clients are connected. Treating zero clients as an error would incorrectly drive the subscription into `error` / `off` status — a WebSocket subscription is valid before any client connects.

**Handshake is deferred to connect time.** Unlike rest-hook (which POSTs a handshake to validate the endpoint), WebSocket subscriptions activate immediately. The handshake bundle is sent as the first WebSocket frame when a client connects and binds.

**Binding tokens are single-use.** `DashMap::remove()` is atomic — a token can only be consumed once even under concurrent connection attempts.

**Subscription deletion closes clients.** When a `Subscription` resource is deleted, `ws_manager.remove_all_clients()` drops all sender halves. This causes the `rx.recv()` calls in the WebSocket handlers to return `None`, triggering graceful close without an explicit close frame.

**No retry for WebSocket.** The retry loop in `dispatch_with_retry` still runs, but since `WebSocketChannel.dispatch()` always returns `Success`, it exits on the first call. Retries are meaningless for in-process channel broadcasting.

## Test plan

- [x] `cargo test -p helios-subscriptions` — 116 tests pass (106 unit + 2 rest-hook integration + 8 WebSocket integration)
- [x] `cargo test -p helios-rest` — 170 tests pass
- [x] `cargo clippy --all-targets --all-features -- -D warnings …` — zero warnings
- [x] `cargo fmt --all` — clean
- [x] Manual smoke test: `HFS_SUBSCRIPTIONS_ENABLED=true cargo run --bin hfs --features subscriptions`, create a websocket subscription, call `$get-ws-binding-token`, connect with `websocat "ws://localhost:8080/ws/subscriptions/bind?token=<token>"`, POST a matching resource, verify notification bundle arrives


aacruzgon:
Additional fixes in the subscriptions smoke regression **and** closes the R4 backport gaps that caused it.

  #### 1) R4 backport topic handling (`Basic`) is now implemented
  - Added strict parsing for R4 backport topics represented as `Basic` resources.
  - Topic detection requires:
    - `resourceType = Basic`
    - `code.coding` includes `system=http://hl7.org/fhir/fhir-types` + `code=SubscriptionTopic`
    - required core topic extensions (canonical URL + resource trigger)
  - Added parsing for trigger/filter/notification-shape backport extensions into internal `TopicDefinition`.

  **Why:**
  Our smoke/job was failing at `POST /SubscriptionTopic` on R4 builds because native `SubscriptionTopic` is not valid in R4.
  Backport topics in R4 are represented through `Basic`, so we now support that path directly and safely (strict markers prevent accidental misclassification of arbitrary `Basic` resources).

  ---

  #### 2) Engine lifecycle now recognizes R4 `Basic` topic events
  - Extended subscription engine event routing to process R4 `Basic` topic create/update/delete lifecycle events.
  - Non-topic `Basic` resources still flow through normal evaluation (no behavior change there).

  **Why:**
  Even with parser support, topics were never registered unless event type was native `SubscriptionTopic`. This blocked end-to-end R4 topic registration.

  ---

  #### 3) Backport payload-content extraction improved for R4 `Subscription`
  - R4 channel parsing now prefers `channel._payload.extension(backport-payload-content)` and falls back to the previous root extension path for compatibility.

  **Why:**
  This aligns with backport-style payload extension placement while preserving compatibility with older payload shapes.

  ---

  #### 4) WebSocket binding protocol switched to `bind-with-token`
  - `/ws/subscriptions/bind` now requires first WS message: `bind-with-token <token>`.
  - Token validation and subscription binding happen after upgrade on that message.
  - Enforced single bind per connection; additional bind attempts are rejected.

  **Why:**
  This aligns runtime behavior with the backport websocket interaction model and removes ambiguity from query-string token binding.

  ---

  #### 5) External subscriptions smoke test updated for true R4 backport flow
  - Topic creation changed from `POST /SubscriptionTopic` to `POST /Basic` with backport topic extensions.
  - R4 `Subscription` payloads now include backport profile and payload-content extension.
  - WebSocket smoke flow now:
    1. calls `$get-ws-binding-token`
    2. connects to `websocket-url`
    3. sends `bind-with-token <token>` over the socket
  - Kept existing handshake/event assertions (`Bundle.type=history`, `Parameters.type` checks).

  **Why:**
  The previous smoke was testing native-topic behavior against an R4 build, causing immediate failure before notification logic executed. This now validates the intended R4 backport contract end-to-end.

  ---

  #### 6) Documentation updated
  - README websocket binding docs now describe the bind-message flow instead of `?token=` URL binding.

  ---

  ### Validation Performed
  - `cargo test -p helios-subscriptions` (full suite passed)
  - Targeted/new tests for:
    - R4 `Basic` topic parsing
    - R4 `Basic` topic engine registration
    - R4 payload-content via `channel._payload` extension
    - websocket bind message parsing
  - `cargo check -p helios-rest --features subscriptions`
  - smoke script syntax check (`bash -n`)

  ---

  ### Net effect
This  makes the R4 path actually backport-compliant in the tested areas, unblocks the external subscriptions smoke workflow, and removes protocol drift between documented and implemented websocket binding behavior.

aacruzgon:
Implemented the non-R4 subscriptions to make R4B/R5/R6 build, run, and test like first-class paths (while preserving R4 backport behavior).

  ### What we changed

  - Feature wiring across crates so version flags propagate correctly:
      - helios-subscriptions -> helios-persistence -> helios-fhirpath
      - Disabled implicit default-version activation where needed (default-features = false on direct helios-fhir/helios-fhirpath deps).
  - Implemented real SubscriptionTopic delete behavior in the engine:
      - Added topic canonical tracking by (tenant, resource_type, resource_id) so delete events without payload can still remove registry entries.
      - Applied for both native SubscriptionTopic and R4 backport Basic topic resources.
  - Fixed delete-event emission in REST:
      - Delete now emits with resolved fhir_version and previous resource payload when available (instead of hardcoded default version).
  - Completed native non-R4 behavior in subscription parsing/notification:
      - filterBy.comparator is preserved in filter conversion.
      - Native SubscriptionStatus fields eventsSinceSubscriptionStart and eventNumber are emitted as integers.
      - R4 remains backport (Parameters); R4B/R5/R6 use native SubscriptionStatus.
  - Made integration tests version-aware for both channels:
      - rest-hook and websocket tests now run correctly for R4 and non-R4 builds.
  - Expanded CI coverage:
      - Added FHIR-version matrix to subscriptions channel integration workflow.
      - Added version matrix to external smoke workflow/build artifacts.
  - Updated external smoke script:
      - Added FHIR_VERSION support (R4, R4B, R5, R6).
      - Version-specific topic/subscription payload generation and assertions.
      - Bundle/type assertions now adapt to backport vs native notification shapes.
  - Updated subscriptions README:
      - Aligned docs with actual phase-2 behavior and version split (R4 backport vs native R4B/R5/R6).

  ### Why

  The prior build failures were caused by two coupled issues:

  1. Build-time version mismatch from incomplete feature propagation (non-R4 builds unintentionally pulled R4-only defaults in dependency chains).
  2. Runtime/test assumptions were still R4-centric (topic lifecycle delete gaps, payload shape assertions, and smoke payloads hardcoded to backport format).

  This set of changes removes both blockers so non-R4 paths are not only compilable, but behaviorally covered for rest-hook and websocket.

  ### Validation run

  - cargo check -p helios-subscriptions --no-default-features --features R4|R4B|R5|R6
  - Channel integration tests passed for non-R4:
      - rest_hook_integration and websocket_integration with R4B, R5, R6
  - HFS compile checks passed with non-R4 feature sets used by smoke builds.
  - Smoke script syntax validated (bash -n) and workflow matrices updated.

  ———

crates/fhirpath/src/cli.rs
```diff
@@ -82,6 +82,33 @@ use crate::parse_debug::{expression_to_debug_tree, generate_parse_debug};
 use crate::{EvaluationResult, evaluate_expression};
 use helios_fhir::{FhirResource, FhirVersion};
 
+#[allow(unreachable_code)]
+fn default_cli_fhir_version() -> FhirVersion {
```
smunini: This new method might be a problem.  This was likely added because you wanted to allow compilation of the codebase for only a specific version of FHIR, for example, only R4B, which would allow for a very specific and minimal/optimized build of the code, which we should permit.  In most cases (and the binaries we ship is this case), we compile with all feature flags, and as I read this new method, I'm unsure what the default fhir version will be.  I think we need to add a cfg arm for the case where all fhir version features are present, and it should default to R4.

crates/fhirpath/src/models.rs
```diff
@@ -168,6 +168,7 @@ pub fn extract_parameters(params: FhirPathParameters) -> Result<ExtractedParamet
     let mut extracted = ExtractedParameters::default();
 
     // Process parameters based on version
+    #[allow(unreachable_patterns)]
```
smunini: ? In what situation is this unreachable?

crates/fhirpath/src/cli.rs
```diff
@@ -82,6 +82,33 @@ use crate::parse_debug::{expression_to_debug_tree, generate_parse_debug};
 use crate::{EvaluationResult, evaluate_expression};
 use helios_fhir::{FhirResource, FhirVersion};
 
+#[allow(unreachable_code)]
```
smunini: In what scenario is this unreachable?  I think this can be removed.

crates/fhirpath/src/handlers.rs
```diff
@@ -547,6 +547,7 @@ fn detect_fhir_version(resource: &Value) -> FhirVersion {
 
 /// Parse FHIR resource based on version
 fn parse_fhir_resource(json: Value, version: FhirVersion) -> FhirPathResult<FhirResource> {
+    #[allow(unreachable_patterns)]
```
smunini: Same here

crates/fhirpath/src/models.rs
```diff
@@ -185,6 +186,12 @@ pub fn extract_parameters(params: FhirPathParameters) -> Result<ExtractedParamet
         FhirPathParameters::R6(parameters) => {
             extract_parameters_from_r6(parameters, &mut extracted)?;
         }
+        _ => {
```
smunini, line 189: remove

crates/fhirpath/src/cli.rs
smunini: @claude I think we can safely revert all of the changes in the fhirpath crate - they are not necessary.

crates/hfs/Cargo.toml
```diff
@@ -18,10 +18,10 @@ path = "src/main.rs"
 default = ["R4", "sqlite"]
 
 # FHIR version features
-R4 = ["helios-fhir/R4", "helios-rest/R4"]
-R4B = ["helios-fhir/R4B", "helios-rest/R4B"]
-R5 = ["helios-fhir/R5", "helios-rest/R5"]
-R6 = ["helios-fhir/R6", "helios-rest/R6"]
+R4 = ["helios-fhir/R4", "helios-rest/R4", "helios-subscriptions?/R4"]
```
smunini: @claude - what is the questionmark syntax on helios-subscriptions and why is it different than the other crates?

crates/hfs/Cargo.toml
```diff
@@ -46,12 +46,12 @@ cloudwatch = ["helios-audit/cloudwatch"]
 reqwest = { version = "0.12", features = ["blocking"] }
 
 [dependencies]
-helios-fhir = { path = "../fhir", version = "0.1.47" }
+helios-fhir = { path = "../fhir", version = "0.1.47", default-features = false }
```
smunini, line 49: @claude - why are default-features = false now?

crates/hfs/Cargo.toml
```diff
@@ -46,12 +46,12 @@ cloudwatch = ["helios-audit/cloudwatch"]
...
 helios-rest = { path = "../rest", version = "0.1.47", default-features = false }
 helios-persistence = { path = "../persistence", version = "0.1.47", default-features = false, features = ["audit"] }
 helios-auth = { path = "../auth", version = "0.1.47" }
 helios-audit = { path = "../audit", version = "0.1.47" }
-helios-subscriptions = { path = "../subscriptions", version = "0.1.47", optional = true }
+helios-subscriptions = { path = "../subscriptions", version = "0.1.47", optional = true, default-features = false }
```
smunini, line 54: Same here - why false?

crates/persistence/src/composite/storage.rs
```diff
@@ -641,8 +641,16 @@ impl CompositeStorage {
                 let fhir_version = resource_json
                     .get("meta")
                     .and_then(|m| m.get("profile"))
-                    .map(|_| FhirVersion::default())
-                    .unwrap_or_default();
+                    .and_then(|v| v.as_array())
```
smunini: This doesn't look correct.  I think the LLM made this change because there was a compilation without a FHIR Version enabled.  I think we should undo this

crates/persistence/src/search/loader.rs
```diff
@@ -617,7 +617,12 @@ impl SearchParameterLoader {
 
 impl Default for SearchParameterLoader {
     fn default() -> Self {
-        Self::new(FhirVersion::R4)
+        Self::new(
```
smunini: Same here - undo

crates/persistence/src/types/stored_resource.rs
```diff
@@ -422,7 +422,12 @@ impl StoredResourceBuilder {
             id: self.id.expect("id is required"),
             version_id,
             tenant_id: self.tenant_id.expect("tenant_id is required"),
-            fhir_version: self.fhir_version.unwrap_or_default(),
+            fhir_version: self.fhir_version.unwrap_or_else(|| {
```
smunini: Undo

crates/hfs/Cargo.toml
```diff
@@ -18,10 +18,10 @@ path = "src/main.rs"
...
 
 # FHIR version features
-R4 = ["helios-fhir/R4", "helios-rest/R4"]
-R4B = ["helios-fhir/R4B", "helios-rest/R4B"]
-R5 = ["helios-fhir/R5", "helios-rest/R5"]
-R6 = ["helios-fhir/R6", "helios-rest/R6"]
+R4 = ["helios-fhir/R4", "helios-rest/R4", "helios-subscriptions?/R4"]
+R4B = ["helios-fhir/R4B", "helios-rest/R4B", "helios-subscriptions?/R4B"]
+R5 = ["helios-fhir/R5", "helios-rest/R5", "helios-subscriptions?/R5"]
+R6 = ["helios-fhir/R6", "helios-rest/R6", "helios-subscriptions?/R6"]
```
smunini: Also, I'm unsure we need all three.  I think we should rely on one and only one FHIR Version feature definition - I think the helios-fhir crate should be the source of truth because it is foundational, so all of these in the codebase should be R5 = ["helios-fhir/R5"] for example - in all Cargo.toml files where fhir version feature flags are needed.

crates/rest/src/handlers/compartment.rs
```diff
@@ -54,6 +54,8 @@ fn get_compartment_params_for_version(
         FhirVersion::R5 => helios_fhir::r5::get_compartment_params(compartment_type, resource_type),
         #[cfg(feature = "R6")]
         FhirVersion::R6 => helios_fhir::r6::get_compartment_params(compartment_type, resource_type),
+        #[allow(unreachable_patterns)]
```
smunini, line 67: Remove

crates/persistence/Cargo.toml
smunini, line 1: re-work as described above

crates/fhirpath/Cargo.toml
```diff
@@ -18,7 +18,7 @@ R5 = ["helios-fhir/R5"]
 R6 = ["helios-fhir/R6"]
 
 [dependencies]
-helios-fhir = { path = "../fhir", version = "0.1.47" }
+helios-fhir = { path = "../fhir", version = "0.1.47", default-features = false }
```
smunini, line 21: ?

crates/fhirpath/src/not_function.rs
```diff
@@ -73,13 +73,12 @@ pub fn not_function(
     // Convert invocation_base to its 3-valued logic boolean form.
     // This handles singletons (Boolean, Integer, String, etc.) and empty/singleton collections.
     // Pass R4 compatibility flag based on FHIR version
-    use helios_fhir::FhirVersion;
+    #[allow(unreachable_patterns)]
```
smunini: Remove

crates/rest/src/handlers/subscriptions.rs
```diff
@@ -83,15 +83,21 @@ where
         })?;
 
     // Return a Bundle with a SubscriptionStatus indicating query-status
+    let bundle_type = if matches!(sub.fhir_version.as_str(), "R5" | "R6") {
```
smunini: We should not hard code "R5" or "R6"  Use the FhirVersion enum in the fhir crate instead.

crates/rest/Cargo.toml
```diff
@@ -38,19 +38,19 @@ subscriptions = ["dep:helios-subscriptions"]
 
 [dependencies]
 # Core dependencies
-helios-fhir = { path = "../fhir", version = "0.1.47" }
+helios-fhir = { path = "../fhir", version = "0.1.47", default-features = false }
```
smunini, line 41: remove

crates/rest/Cargo.toml
```diff
@@ -38,19 +38,19 @@ subscriptions = ["dep:helios-subscriptions"]
...
 helios-persistence = { path = "../persistence", version = "0.1.47", default-features = false }
 helios-serde = { path = "../serde", version = "0.1.47", default-features = false, optional = true }
 helios-auth = { path = "../auth", version = "0.1.47" }
 helios-audit = { path = "../audit", version = "0.1.47" }
-helios-subscriptions = { path = "../subscriptions", version = "0.1.47", optional = true }
+helios-subscriptions = { path = "../subscriptions", version = "0.1.47", optional = true, default-features = false }
```
smunini, line 46: remove

crates/subscriptions/Cargo.toml
```diff
@@ -10,14 +10,14 @@ description = "FHIR topic-based Subscriptions engine for HFS"
 
 [features]
 default = ["R4"]
-R4 = ["helios-fhir/R4"]
-R4B = ["helios-fhir/R4B"]
-R5 = ["helios-fhir/R5"]
-R6 = ["helios-fhir/R6"]
+R4 = ["helios-fhir/R4", "helios-persistence/R4"]
```
smunini, line 13: remove

crates/subscriptions/Cargo.toml
```diff
@@ -10,14 +10,14 @@ description = "FHIR topic-based Subscriptions engine for HFS"
...
+R4B = ["helios-fhir/R4B", "helios-persistence/R4B"]
```
smunini, line 14: remove

crates/subscriptions/Cargo.toml
```diff
@@ -10,14 +10,14 @@ description = "FHIR topic-based Subscriptions engine for HFS"
...
+R5 = ["helios-fhir/R5", "helios-persistence/R5"]
+R6 = ["helios-fhir/R6", "helios-persistence/R6"]
 
 [dependencies]
 # FHIR models
-helios-fhir = { path = "../fhir" }
+helios-fhir = { path = "../fhir", default-features = false }
```
smunini, line 20: remove

smunini (changes requested):
Overall a good PR, however, I think there was a compilation without any fhir version feature and it caused a bunch of changes that are unnecessary. I think we can revert all of the changes to the fhirpath crate for example. It's ok to assume we will always have at least one fhirversion feature turned on (or all of them) during a compile.  Perhaps something to add to CLAUDE.md.

smunini (approved):

smunini (dismissed):

smunini (approved):

PR #89, opened 2026-04-28
Feature/subscriptions Phase 4: FHIR messaging

aacruzgon:
---

  ## Summary

  - Phase 4 of the subscriptions effort: adds the FHIR Messaging channel
  (`Bundle.type = "message"` with a `MessageHeader` first entry) per the
  [Backport IG channels page](https://build.fhir.org/ig/HL7/fhir-subscription-backport-ig/channels.html). `ChannelType::Message` was already an enum variant
  - Engine wiring via `SubscriptionEngine::with_outbound_auth`;
  `SubscriptionEngine::new` preserves the existing no-op outbound behavior so
  unrelated callers don't change.
  - Smoke + channel-matrix CI extended; messaging integration test mirrors
  `rest_hook_integration.rs`.

  ## Why

  Continuation of [Discussion#59](https://github.com/HeliosSoftware/hfs/discussions/59) — Phases 1–3 (rest-hook, websocket, email) shipped. Phase 4 was the last remaining channel from the IG.

  The security ask was the harder half. Investigation of `crates/auth/`
  confirmed the crate is purely inbound: JWKS bearer validation, SMART scope
  policy, audit. Nothing was attaching server identity to outbound requests, so
  a server-to-server FHIR Messaging channel that "imposes auth requirements"
  needed a new abstraction. Rather than hard-code bearer-token logic in the
  messaging dispatcher, `OutboundAuthProvider` lives in `helios-auth` so a
  future `JwtAssertionOutboundAuthProvider` (SMART Backend Services–style
  client-credentials assertion) drops in without touching channel code.

  SSRF guard is on by default because messaging endpoints are operator-supplied
  URLs — production posture should reject
  `127.0.0.1`/`10.0.0.0/8`/`fc00::/7`/etc. Local dev opts in via
  `HFS_SUBSCRIPTION_ALLOW_PRIVATE_ENDPOINTS=true`; the smoke workflow sets it
  explicitly.

  ## Approach

  **`OutboundAuthProvider` (helios-auth)**

  - `async fn authorize(&self, RequestBuilder, audience: &str) ->
  Result<RequestBuilder, AuthError>`. The `audience` arg is unused by the static
   impl but is wired through so a future signer can scope minted tokens per
  receiver.
  - `AuthConfig::outbound_provider()` returns the right `Arc<dyn
  OutboundAuthProvider>` based on `HFS_OUTBOUND_BEARER_TOKEN` presence.
  - Built before `auth_config` is moved into `AppState` so the subscription
  engine can consume it.

  **Notification wrapping (`notification/messaging.rs`)**

  - `wrap_as_message_bundle(notification_bundle, subscription, source_endpoint,
  notification_type)` — builds a fresh `Bundle.type=message` whose `entry[0]` is
   a new `MessageHeader` and whose remaining entries are the original
  notification bundle's entries verbatim (SubscriptionStatus or backport
  Parameters + payload). Not nested Bundle-in-Bundle.
  - `MessageHeader.source.endpoint = HFS_BASE_URL`, `.destination[0].endpoint =
  subscription.channel.endpoint`, `.focus[0].reference = Subscription/{id}`.
  - Event identifier is the topic canonical URL — preserves topic semantics
  across channels. R4B+ also sets `definition` because `eventCoding.code`
  carries the notification kind, not the topic.

  **`MessagingChannel` (`channels/messaging.rs`)**

  - `dispatch()` flow: resolve endpoint → TLS check → SSRF check → wrap → POST
  with `Content-Type: application/fhir+json` → forward subscription headers → if
   no subscription `Authorization`, apply outbound provider → status-code
  mapping identical to rest-hook (2xx Success / 4xx PermanentError / 5xx+timeout
   RetryableError).
  - The `ChannelDispatcher` trait passes the *prebuilt* notification bundle, so
  the dispatcher needs the `NotificationType` for the MessageHeader event
  coding. New helper `notification::notification_type_from_bundle` inspects the
  status entry (R4 Parameters or R4B+ SubscriptionStatus) — no trait change, no
  impact on rest-hook/websocket/email.
  - SSRF guard checks the URL's literal IP via `std::net::IpAddr` predicates.
  DNS-resolved hosts are passed through — that's a network-policy concern, not a
   string check.

  **Engine wiring**

  - `SubscriptionEngine::with_outbound_auth(config, base_url, Arc<dyn
  OutboundAuthProvider>)` constructs the messaging channel when
  `config.messaging` is `Some(MessagingSettings { source_endpoint,
  allow_private_endpoints })`.
  - New match arms in `dispatch_with_retry` and `activate_subscription`
  (handshake path); `_` arm unchanged.
  - Env vars consumed by
  `crates/rest/src/lib.rs::build_messaging_settings_from_env`:
  `HFS_SUBSCRIPTION_MESSAGING_ENABLED`,
  `HFS_SUBSCRIPTION_MESSAGE_SOURCE_ENDPOINT` (defaults to `HFS_BASE_URL`),
  `HFS_SUBSCRIPTION_ALLOW_PRIVATE_ENDPOINTS`, `HFS_OUTBOUND_BEARER_TOKEN`.

  ## Test plan

  - [x] 6 new `helios-auth` outbound tests (NoOp, static bearer presence,
  token).
  - [x] 16 new `notification::messaging` unit tests across R4/R4B/R5/R6 — Bundle
   shape, MessageHeader source/destination/focus, eventUri vs eventCoding per
  version, payload entry preservation.
  - [x] 14 new `channels::messaging` unit tests — success, server/client error
  mapping, timeout, headers, outbound auth, subscription precedence,
  body-is-message-Bundle, TLS, missing endpoint, SSRF allow/reject (loopback +
  10/8 + DNS-passthrough), handshake event coding.
  - [x] 4 new `messaging_integration.rs` cases × 4 FHIR versions
  (`HFS_TEST_FHIR_VERSION`) = 16 runs — full lifecycle (handshake +
  event-notification), header precedence on the wire, channel-not-supported
  rejection, multi-tenant isolation.
  - [x] All 38 `helios-auth` + 150 `helios-subscriptions` lib tests pass.
  - [x] All 388 `helios-rest` tests pass with `--features subscriptions`.
  - [x] `cargo fmt --all` clean.
  - [x] Smoke shell-script messaging block (rest-hook webhook capture + jq
  filter on `.type=="message"` + MessageHeader assertions).
  - [x] `subscriptions-channels.yml`: new `messaging-integration` matrix job (15
   build configs × FHIR-version subset, mirrors rest-hook).
  - [x] `subscriptions-smoke.yml`: env vars set so engine activates messaging
  and the SSRF guard permits the in-cluster `127.0.0.1` webhook capture.
  - [ ] Real-world receiver smoke against an external `$process-message`
  endpoint — deferred until a target is available.

  ## Follow-ups

  - CapabilityStatement does not currently advertise per-channel support for any
   channel (rest-hook/ws/email/messaging). Adding channel advertisement is a
  separate refactor that should cover all four at once; gating messaging on it
  would block this PR for an unrelated cleanup.
  - `JwtAssertionOutboundAuthProvider` (SMART Backend Services outbound) is the
  natural extension of the new trait — explicitly out of scope here.

### DIFF
--- a/.github/workflows/subscriptions-smoke.yml
+++ b/.github/workflows/subscriptions-smoke.yml
@@ -6,6 +6,8 @@ name: HFS Subscriptions External Smoke
 #                        except R6 (R6 is ballot), crossed with 5 backends.
 # R6 is compiled and tested by subscriptions-channels.yml (integration + unit
 # tests) but not exercised at runtime by the external smoke script.
+# Each smoke job exercises rest-hook + websocket + email (when MAILPIT_HTTP_URL
+# is set) + FHIR Messaging channels.
 
 on:
   workflow_dispatch:
@@ -113,6 +115,10 @@ jobs:
         include: ${{ fromJSON(needs.matrix-setup.outputs.smoke-matrix) }}
     env:
       RESULTS_DIR: subscriptions-smoke-results/${{ matrix.build_key }}/${{ matrix.fhir_version }}/${{ matrix.backend }}
+      # Enable the FHIR Messaging channel and let it post to 127.0.0.1
+      # webhook capture (off by default outside CI).
+      HFS_SUBSCRIPTION_MESSAGING_ENABLED: "true"
+      HFS_SUBSCRIPTION_ALLOW_PRIVATE_ENDPOINTS: "true"
     steps:
       - name: Checkout code
         uses: actions/checkout@v5
--- a/crates/rest/src/handlers/create.rs
+++ b/crates/rest/src/handlers/create.rs
@@ -210,8 +210,7 @@ where
             &stored,
             fhir_version,
             helios_subscriptions::ResourceEventType::Create,
-        )
-        .await;
+        );
     }
 
     build_create_response(
--- a/crates/rest/src/handlers/create.rs
+++ b/crates/rest/src/handlers/create.rs
@@ -210,7 +210,8 @@ where
             &stored,
             fhir_version,
             helios_subscriptions::ResourceEventType::Create,
-        );
+        )
+        .await;
     }
 
     build_create_response(
--- a/crates/rest/src/handlers/delete.rs
+++ b/crates/rest/src/handlers/delete.rs
@@ -94,7 +94,8 @@ where
             &id,
             fhir_version,
             existing_resource.map(|stored| stored.content().clone()),
-        );
+        )
+        .await;
     }
 
     // Return 204 No Content (or 200 with OperationOutcome)
--- a/crates/rest/src/handlers/delete.rs
+++ b/crates/rest/src/handlers/delete.rs
@@ -94,8 +94,7 @@ where
             &id,
             fhir_version,
             existing_resource.map(|stored| stored.content().clone()),
-        )
-        .await;
+        );
     }
 
     // Return 204 No Content (or 200 with OperationOutcome)
--- a/crates/rest/src/handlers/patch.rs
+++ b/crates/rest/src/handlers/patch.rs
@@ -141,7 +141,8 @@ where
             &stored,
             stored.fhir_version(),
             helios_subscriptions::ResourceEventType::Update,
-        );
+        )
+        .await;
     }
 
     build_patch_response(&stored, headers, &prefer).map(|mut response| {
--- a/crates/rest/src/handlers/patch.rs
+++ b/crates/rest/src/handlers/patch.rs
@@ -141,8 +141,7 @@ where
             &stored,
             stored.fhir_version(),
             helios_subscriptions::ResourceEventType::Update,
-        )
-        .await;
+        );
     }
 
     build_patch_response(&stored, headers, &prefer).map(|mut response| {
--- a/crates/rest/src/handlers/subscription_event.rs
+++ b/crates/rest/src/handlers/subscription_event.rs
@@ -120,7 +120,10 @@ mod tests {
 
     #[test]
     fn subscriptions_are_not_inline() {
-        assert!(!should_handle_inline("Subscription", FhirVersion::default()));
+        assert!(!should_handle_inline(
+            "Subscription",
+            FhirVersion::default()
+        ));
     }
 
     #[test]
--- a/crates/rest/src/handlers/subscription_event.rs
+++ b/crates/rest/src/handlers/subscription_event.rs
@@ -1,8 +1,7 @@
 //! Subscription event emission helper.
 //!
-//! Constructs a `ResourceEvent` from handler context and spawns the
-//! subscription engine's event handler asynchronously, mirroring the
-//! fire-and-forget pattern used by the audit middleware.
+//! Constructs a `ResourceEvent` from handler context and dispatches it to the
+//! subscription engine.
 
 use std::sync::Arc;
 
@@ -15,9 +14,11 @@ use tracing::debug;
 /// Emits a subscription event for a successful resource write.
 ///
 /// This function constructs a `ResourceEvent` from the handler context and
-/// spawns the subscription engine's `on_resource_event` asynchronously.
+/// dispatches it to the subscription engine. Subscription lifecycle resources
+/// are handled inline so their in-memory state is visible before the HTTP
+/// response returns; ordinary data-resource events remain fire-and-forget.
 /// It is a no-op if the subscription engine is not configured.
-pub fn emit_subscription_event(
+pub async fn emit_subscription_event(
     engine: &Arc<SubscriptionEngine>,
     tenant: &TenantContext,
     stored: &StoredResource,
@@ -45,6 +46,11 @@ pub fn emit_subscription_event(
         "Emitting subscription event"
     );
 
+    if should_handle_inline(&event.resource_type, event.fhir_version) {
+        engine.on_resource_event(event).await;
+        return;
+    }
+
     tokio::spawn(async move {
         engine.on_resource_event(event).await;
     });
@@ -53,7 +59,7 @@ pub fn emit_subscription_event(
 /// Emits a subscription event for a resource delete.
 ///
 /// Delete events carry the resource type and ID but no resource content.
-pub fn emit_delete_event(
+pub async fn emit_delete_event(
     engine: &Arc<SubscriptionEngine>,
     tenant: &TenantContext,
     resource_type: &str,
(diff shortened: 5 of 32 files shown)


.github/workflows/subscriptions-smoke.yml
```diff
@@ -113,6 +115,10 @@ jobs:
         include: ${{ fromJSON(needs.matrix-setup.outputs.smoke-matrix) }}
     env:
       RESULTS_DIR: subscriptions-smoke-results/${{ matrix.build_key }}/${{ matrix.fhir_version }}/${{ matrix.backend }}
+      # Enable the FHIR Messaging channel and let it post to 127.0.0.1
+      # webhook capture (off by default outside CI).
+      HFS_SUBSCRIPTION_MESSAGING_ENABLED: "true"
```
smunini, line 120: These new environment variables are not documented in the README.md

smunini (changes requested):

.github/workflows/subscriptions-smoke.yml
aacruzgon: Documented these in 2a5ada1f. I added the messaging channel variables to the root README environment table and mirrored the channel-specific details in crates/subscriptions/README.md.

crates/rest/src/handlers/create.rs
```diff
@@ -210,7 +210,8 @@ where
             &stored,
             fhir_version,
             helios_subscriptions::ResourceEventType::Create,
-        );
+        )
+        .await;
```
smunini: Let's remove these awaits - I suspect they were added debugging another issue elsewhere.

crates/rest/src/handlers/delete.rs
```diff
@@ -94,7 +94,8 @@ where
             &id,
             fhir_version,
             existing_resource.map(|stored| stored.content().clone()),
-        );
```
smunini, line 97: Let's remove these awaits - I suspect they were added debugging another issue elsewhere.

crates/rest/src/handlers/patch.rs
```diff
@@ -141,7 +141,8 @@ where
             &stored,
             stored.fhir_version(),
             helios_subscriptions::ResourceEventType::Update,
-        );
+        )
```
smunini: Let's remove these awaits - I suspect they were added debugging another issue elsewhere.

crates/rest/src/handlers/update.rs
```diff
@@ -198,7 +198,8 @@ where
             &stored,
             fhir_version,
             event_type,
-        );
+        )
```
smunini: Let's remove these awaits - I suspect they were added debugging another issue elsewhere.

crates/rest/src/handlers/subscription_event.rs
```diff
@@ -15,9 +14,12 @@ use tracing::debug;
...
 ///
 /// This function constructs a `ResourceEvent` from the handler context and
-/// spawns the subscription engine's `on_resource_event` asynchronously.
+/// dispatches it to the subscription engine. Topic lifecycle resources are
+/// handled inline so their in-memory state is visible before the HTTP response
+/// returns; subscriptions and ordinary data-resource events remain
+/// fire-and-forget.
 /// It is a no-op if the subscription engine is not configured.
-pub fn emit_subscription_event(
+pub async fn emit_subscription_event(
```
smunini: remove async

crates/rest/src/handlers/subscription_event.rs
```diff
@@ -53,7 +60,7 @@ pub fn emit_subscription_event(
 /// Emits a subscription event for a resource delete.
 ///
 /// Delete events carry the resource type and ID but no resource content.
-pub fn emit_delete_event(
+pub async fn emit_delete_event(
```
smunini: remove async

aacruzgon:
## Inferno Subscription Test Suite Stabilization Summary

  This Additional changes in the PR fixes the flaky Inferno Subscription Test Suite failures where some
  backends intermittently failed with:

  > No notifications were received from the server related to Subscription ...

  ### Root Cause

  The failures were not caused by workflow port conflicts or backend startup
  issues. The shared Inferno stack, HFS server, topic seeding, session creation,
  and event trigger were all completing successfully.

  The actual race was in subscription activation:

  1. Inferno created a `Subscription` in HFS.
  2. HFS accepted the resource and asynchronously sent the rest-hook handshake.
  3. In some runs, HFS sent the handshake before Inferno had fully associated
  the notification token with the active test run.
  4. Inferno returned `HTTP 500` from the notification listener with an error
  like:

     ```text
     Unable to find test run with identifier 'notification <token>'

  5. HFS treated the failed handshake as an activation failure.
  6. The subscription stayed inactive/error, so the later Encounter event
     matched 0 active subscriptions.
  7. Inferno then reported that no notifications were received.

  ### What Changed

  - Kept topic lifecycle processing synchronous where ordering matters:
      - SubscriptionTopic
      - R4 Basic backport topic resources
  - Kept Subscription activation asynchronous so Inferno can finish recording
    the subscription creation response before callbacks arrive.
  - Added subscription engine diagnostics at info level.
  - Updated runtime logging so helios_subscriptions=info appears in CI hfs.log.
  - Added Inferno container/server log capture to the workflow artifacts.
  - Used those logs to confirm Inferno was returning 500 because the
    notification token/test run was not ready yet.
  - Added configurable activation handshake controls:
      - HFS_SUBSCRIPTION_HANDSHAKE_INITIAL_DELAY_MS
      - HFS_SUBSCRIPTION_HANDSHAKE_MAX_ATTEMPTS
      - HFS_SUBSCRIPTION_HANDSHAKE_RETRY_BASE_MS
      - HFS_SUBSCRIPTION_HANDSHAKE_RETRY_MAX_MS
  - Updated HFS activation behavior:
      - keep subscriptions in requested while retryable handshake failures are
        retried
      - transition to active only after a successful handshake
      - transition to error only after retry attempts are exhausted or a
        permanent error occurs
      - continue preventing event notifications until activation succeeds
  - Configured the Inferno Subscription workflow with a short initial handshake
    delay and retry window to avoid the Inferno listener readiness race.

  ### Validation

  The focused Rust checks pass:

  cargo test -p helios-subscriptions --no-default-features --features R4
  cargo check -p helios-rest --no-default-features --features
  R4,subscriptions,sqlite
  cargo test -p helios-rest --no-default-features --features
  R4,subscriptions,sqlite

  The Inferno Subscription Test Suite has stabilized across the backend matrix
  after these changes.

smunini (changes requested):

smunini (approved):

PR #113, opened 2026-05-22
fix(ci): pin mongo to 7.0 — MongoDB 8.x crashes on CI host kernel 6.19

smunini:
## Summary

The Inferno US Core **MongoDB** matrix jobs were failing while every other storage backend passed.

**Root cause:** the CI Docker host (a Proxmox LXC) shares a Linux kernel **≥ 6.19**, which is incompatible with MongoDB 8.0+. MongoDB's vendored TCMalloc violates the kernel's `rseq` ABI, so **mongod SIGSEGV-crashes ~25–30 s after startup** ([SERVER-121912](https://jira.mongodb.org/browse/SERVER-121912)). A diagnostic step added in this PR confirmed it: `docker inspect` reports `OOMKilled=false ExitCode=139` (a crash, not an OOM kill) on a host with 92 GiB RAM / 58 GiB free — not a memory problem. Full write-up and sources are in [this comment](https://github.com/HeliosSoftware/hfs/pull/113#issuecomment-4521680193).

(The commit history shows an earlier, incorrect OOM/swap-thrash diagnosis — the diagnostic step is what corrected it. The net diff reflects the real fix.)

## Changes

- **Pin the mongo image to `mongo:7.0`** in all 5 workflows that start mongod (`inferno-us-core`, `inferno-subscription`, `audit-events`, `subscriptions-smoke`, `bulk-export-smoke`). MongoDB 7.x is unaffected by SERVER-121912 — this is what unblocks CI.
- Add an `if: failure()` **diagnostic step** to the mongodb jobs: captures `docker inspect` (OOMKilled / exit code), mongod logs, host memory and the kernel OOM log — so future mongo failures are diagnosable without shell access to the host.
- Reuse a single cached MongoDB client / connection pool instead of creating one per request (avoids a connection storm against mongod); build the Mongo config during HFS startup.
- Bump the `mongodb` crate 3.6 → 3.7; set `server_selection_timeout` / `max_idle_time` on the client so requests fail fast against an unavailable server.

## The real fix is host-side

`mongo:7.0` unblocks CI, but the proper fix is to **downgrade/pin the Proxmox host kernel to ≤ 6.18** (or run the Docker host in a VM with its own older kernel). MongoDB 8.x stays unusable on that host until then, or until MongoDB ships a patched TCMalloc.

## Test plan

- [ ] `inferno-us-core` workflow green for the `mongodb` matrix jobs
- [ ] other backends (sqlite / postgres / elasticsearch / s3) still pass
- [ ] `cargo test -p helios-persistence --features mongodb` passes

### DIFF
--- a/.github/workflows/audit-events.yml
+++ b/.github/workflows/audit-events.yml
@@ -292,7 +292,8 @@ jobs:
         run: |
           MONGO_CONTAINER="mongo-audit-${{ github.run_id }}-${{ github.run_attempt }}"
           docker rm -f "$MONGO_CONTAINER" 2>/dev/null || true
-          docker run -d --name "$MONGO_CONTAINER" -p 0:27017 mongo:8.0 --replSet rs0 --bind_ip_all
+          docker run -d --name "$MONGO_CONTAINER" --memory=2g -p 0:27017 mongo:8.3 \
+            --replSet rs0 --bind_ip_all --wiredTigerCacheSizeGB 0.5
 
           echo "MONGO_CONTAINER=$MONGO_CONTAINER" >> $GITHUB_ENV
 
@@ -863,7 +864,8 @@ jobs:
         run: |
           MONGO_CONTAINER="mongo-audit-db-${{ matrix.mode }}-${{ github.run_id }}-${{ github.run_attempt }}"
           docker rm -f "$MONGO_CONTAINER" 2>/dev/null || true
-          docker run -d --name "$MONGO_CONTAINER" -p 0:27017 mongo:8.0 --replSet rs0 --bind_ip_all
+          docker run -d --name "$MONGO_CONTAINER" --memory=2g -p 0:27017 mongo:8.3 \
+            --replSet rs0 --bind_ip_all --wiredTigerCacheSizeGB 0.5
 
           echo "MONGO_CONTAINER=$MONGO_CONTAINER" >> $GITHUB_ENV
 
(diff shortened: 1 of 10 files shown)


smunini:
## Why `mongo:7.0`

MongoDB 8.x SIGSEGV-crashes ~25–30 s after startup on the CI Docker host. The diagnostic step confirmed `OOMKilled=false ExitCode=139` (a crash, not an OOM kill) on a host with **92 GiB RAM / 58 GiB free** — not a memory problem.

Root cause: the CI host (a Proxmox LXC) shares a Linux kernel **≥ 6.19**, which is incompatible with MongoDB 8.0+ — the vendored TCMalloc violates the kernel's `rseq` ABI (**MongoDB SERVER-121912**). `mongo:8.3` refuses to start outright; `mongo:8.0` starts and then crashes. MongoDB 7.x is unaffected — hence the pin to `mongo:7.0`.

The definitive fix is to downgrade/pin the Proxmox host kernel to ≤ 6.18 (or run the Docker host in a VM with its own older kernel); `mongo:7.0` unblocks CI in the meantime.

Sources:
- [MongoDB 8.x and Linux Kernel 6.19 — MongoDB Community](https://www.mongodb.com/community/forums/t/mongodb-8-x-and-linux-kernel-6-19/337547)
- [SERVER-121912](https://jira.mongodb.org/browse/SERVER-121912)
- [MongoDB broken on Linux kernel 6.19 — stoatchat/self-hosted #268](https://github.com/stoatchat/self-hosted/issues/268)

PR #138, opened 2026-06-09
feat(hts): import LOINC language variants at bootstrap and CLI import

darthcav:
## Summary

Imports LOINC linguistic-variant (language translation) files as FHIR `concept.designation` entries, at both server bootstrap and CLI import.

LOINC ships translations inside the main release ZIP under `AccessoryFiles/LinguisticVariants/`: an index `LinguisticVariants.csv` plus per-language files like `frFR28LinguisticVariant.csv`. Both `hts import` and server bootstrap funnel LOINC ZIPs through `import_loinc_csv`, so extending that single parser covers **both** paths — no new flag or file is required.

## Changes

- `find_loinc_paths` now also locates the linguistic-variant index and per-language files (returns a `LoincArchivePaths` struct); the plural index is correctly distinguished from per-variant files.
- New parsers `parse_linguistic_index` and `parse_linguistic_variant`, plus `derive_language_tag` producing BCP-47 tags (`fr-FR`, `de-DE`, `zh-CN`) — index-preferred, filename fallback.
- Translations attach as `concept.designation` entries via the existing `BuilderDesignation` → `concept_designations` path. The English `display` is left untouched. **No DB schema change.**
- Per-language designation counts reported in live and dry-run output, e.g. `imported 3 translated designations (de-DE: 1, fr-FR: 2)`.
- Malformed index, missing columns, unknown LOINC codes, and undetectable languages are recorded as non-fatal parse errors rather than aborting the import.

## Testing

- 10 new tests covering index parsing, language-tag derivation (index + filename fallback), variant CSV parsing, archive file discovery, end-to-end designation writes (verifying English display preserved), dry-run counting, and a regression test that variant-free ZIPs still import cleanly.
- All 23 LOINC tests pass (`cargo test -p helios-hts --features sqlite loinc`).
- `cargo fmt` and `cargo clippy` (CI flags) clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -577,6 +577,10 @@ cargo run --bin hts -- import ./hl7.terminology.r4-6.0.0.tgz
 cargo run --bin hts -- import ./SnomedCT_InternationalRF2_*.zip --format snomed-rf2
 
 # LOINC CSV ZIP  ⚠️  requires free Regenstrief registration at loinc.org
+# Language translations under AccessoryFiles/LinguisticVariants/ (e.g.
+# frFR28LinguisticVariant.csv) are imported automatically as FHIR
+# concept.designation entries tagged with the BCP-47 language (fr-FR, de-DE, …).
+# No extra flag needed — works for both `hts import` and server bootstrap.
 cargo run --bin hts -- import ./Loinc_*.zip --format loinc
 
 # ICD-10-CM tabular XML (free, no license needed — from cms.gov)
--- a/crates/hts/src/import/loinc_csv.rs
+++ b/crates/hts/src/import/loinc_csv.rs
@@ -437,9 +437,10 @@ fn find_loinc_paths(path: &Path) -> Result<LoincArchivePaths, HtsError> {
 
 /// Parse `LinguisticVariants.csv` into `variant_id → (iso_language, iso_country)`.
 ///
-/// Columns: `LinguisticVariantId, IsoLanguage, IsoCountry, LanguageName`. A
-/// missing or malformed index is non-fatal — callers fall back to deriving the
-/// language tag from the per-variant filename.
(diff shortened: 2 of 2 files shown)


darthcav:
## Follow-up fix: axis-only translations (es-ES, it-IT) were silently dropped

Testing against the real `Loinc_2.82` release surfaced two bugs in this PR:

**1. es-ES and it-IT imported zero designations.** These variant files translate only the individual LOINC axis columns (`COMPONENT`, `SYSTEM`, …) and leave `LONG_COMMON_NAME`, `SHORTNAME` *and* `LinguisticVariantDisplayName` empty. The parser only read `LONG_COMMON_NAME` → `SHORTNAME` and skipped rows where both were blank, so all 66k es-ES / 95k it-IT rows were discarded.

Fix: extended the value fallback to `LONG_COMMON_NAME` → `SHORTNAME` → `LinguisticVariantDisplayName` → a synthesized fully-specified name (`COMPONENT:PROPERTY:TIME_ASPCT:SYSTEM:SCALE_TYP:METHOD_TYP` of the populated axes, trailing empty segments dropped).

**2. The index file never parsed against real data.** `parse_linguistic_index` looked for `LinguisticVariantId`/`IsoLanguage`/`IsoCountry`, but real releases use `ID`/`ISO_LANGUAGE`/`ISO_COUNTRY`. The index always failed (logging an error) and tags fell back to filename derivation; the original tests used the invented column names so they passed.

Fix: accept both spellings; updated the index test fixture to real column names.

Added regression tests for axis-only synthesis and real index columns. Verified with a dry-run against `Loinc_2.82`:

```
es-ES: 66058, it-IT: 95394  (both previously 0)  — all 21 languages now present
```

Pushed as `b8a49fb8`.

smunini (approved):

smunini:
Thank you @darthcav !  Great addition.

PR #141, opened 2026-06-09
feat(hts): sync bootstrap directory on every startup

darthcav:
## Problem

The HTS bootstrap importer is first-run-only. On startup it skips the entire `HTS_BOOTSTRAP_DIR` import whenever the database already contains any CodeSystem:

```rust
let n: i64 = "SELECT COUNT(*) FROM code_systems";
if count > 0 { /* bootstrap skipped */ return Ok(()); }
```

As a result, once the server has booted once:

- **A new terminology file added** to the bootstrap directory is never imported.
- **An updated terminology release** replacing an existing file is never re-imported.

## Change

Replace the all-or-nothing guard with a per-file content sync that runs on **every** startup:

- New `bootstrap_imports` ledger table (SQLite + PostgreSQL schemas), keyed on file name, storing the **SHA-256 content hash** and size.
- Each recognized file is hashed and imported only when its hash is **absent from or differs from** the ledger; the new signature is recorded after a successful import. Unchanged files are skipped without re-parsing, so startup stays fast.
- Re-importing is safe — the underlying import upserts on canonical `(url, version)`, so a new version coexists with the old and a same-version re-import refreshes metadata in place.
- RxNorm RRF **directories** are signed by their immediate children's `(name, size)` instead of full-content hashing, to avoid re-reading multi-GB folders on every boot (documented caveat).

The CLI `hts import <dir>` path passes no tracker and keeps its unconditional import behavior unchanged.

## Tests

- `compute_signature` — determinism and content-sensitivity.
- Bootstrap gating end-to-end on an SQLite backend: imports on first run, skips while unchanged, re-imports on a version bump, and picks up a newly added file without re-importing unchanged ones.
- CLI path (no tracker) still imports unconditionally.

All existing `helios-hts` tests pass; `cargo fmt` and `cargo clippy` (CI flags) are clean for both the default (sqlite) and `postgres` feature sets.

## Docs

Documents `HTS_BOOTSTRAP_DIR` (previously undocumented) and the new sync behavior in `CLAUDE.md`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -523,8 +523,27 @@ HTS_DATABASE_URL=./my-terminology.db HTS_SERVER_PORT=9090 cargo run --bin hts
 | `HTS_LOG_LEVEL` | info | Log level (error, warn, info, debug, trace) |
 | `HTS_DATABASE_URL` | ./data/hts.db | Database location — SQLite file path, or `postgresql://user:pass@host/db` for Postgres |
 | `HTS_STORAGE_BACKEND` | sqlite | Storage backend (`sqlite` default; `postgres` when built with `--features postgres`) |
+| `HTS_BOOTSTRAP_DIR` | (none) | Directory of terminology files imported on startup. The Docker image sets this to `/app/terminology-data`. See bootstrap sync below. |
 | `HTS_ENABLE_CORS` | true | Enable CORS |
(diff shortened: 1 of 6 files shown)


smunini:
Thank you @darthcav !  Excellent addition

smunini (approved):

PR #142, opened 2026-06-10
feat(http): compressed request/response support and uniform body-size limits

darthcav:
## Summary

Adds HTTP compression support to all three server binaries (`hfs` via `crates/rest`, `sof-server`, `hts`) and makes the request body-size limit uniform across them.

### Request decompression

- Request bodies sent with `Content-Encoding: gzip` / `deflate` / `br` / `zstd` are decompressed transparently before parsing (`tower_http::decompression::RequestDecompressionLayer`).
- Unsupported encodings are rejected with `415 Unsupported Media Type`; invalid compressed streams produce a 4xx instead of a parse failure deeper in the stack.
- Uncompressed requests are fully backward compatible.

### Response compression

- Responses are compressed when the client advertises support via `Accept-Encoding`, with `Content-Encoding` and `Vary: Accept-Encoding` set (`tower_http::compression::CompressionLayer`).
- SOF excludes `application/parquet` and `application/zip` output from compression — both are already compressed, so HTTP-level gzip would only burn CPU.

### Body-size limits

- The decompression layer runs before extractors, so `HFS_MAX_BODY_SIZE` / `SOF_MAX_BODY_SIZE` / `HTS_MAX_BODY_SIZE` are enforced on the **decompressed** body — a small highly-compressed payload cannot bypass the limit.
- New `HTS_MAX_BODY_SIZE` env var / `--max-body-size` flag (bytes, default `10485760`) gives HTS its first configurable limit, mirroring the existing HFS/SOF variables. HTS was previously stuck at axum's implicit 2 MiB default.

### Adjacent fixes

- **REST:** exceeding the body limit now returns `413` with a `too-long` OperationOutcome (new `RestError::PayloadTooLarge`). Previously the rejection collapsed into a generic `400 "Invalid JSON"`, for uncompressed requests too.
- **HTS CORS:** the explicit-origin path only set `allow-origin`, so every browser preflight failed. It now sets allow-methods and allow-headers (including `Content-Encoding`).
- `Content-Encoding` added to the default CORS allow-header lists of rest and sof so browser clients can send compressed bodies through preflight.

## Tests

25 new integration tests across the three crates (`crates/rest/tests/http_compression.rs`, `crates/hts/tests/http_compression.rs`, inline in `crates/sof/src/server.rs`) covering: gzip/deflate request bodies, invalid gzip → 4xx, unsupported encoding → 415, response compression + `Vary`, identity responses without `Accept-Encoding`, parquet exclusion, decompressed-size limit enforcement (413), and CORS preflight for `content-encoding`.

Validated with `cargo fmt --all`, clippy with the CI flag set (clean), and full `cargo test` runs for `helios-rest`, `helios-hts`, and `helios-sof`.

## Docs

Updated env-var tables and added compression notes in `CLAUDE.md`, the root `README.md`, the `hfs`/`rest`/`sof`/`hts` crate READMEs, and the mdBook pages (`configuration/environment-variables.md`, `ch06-sql-on-fhir.md`, `components/sql-on-fhir.md`, `appendix-a-cli.md`).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -216,10 +216,21 @@ HFS_SERVER_PORT=3000 HFS_LOG_LEVEL=debug cargo run --bin hfs
 #### Limits
 | Variable | Default | Description |
 |----------|---------|-------------|
-| `HFS_MAX_BODY_SIZE` | 10485760 | Max request body size (bytes) |
+| `HFS_MAX_BODY_SIZE` | 10485760 | Max request body size (bytes; applies to the decompressed body for compressed requests) |
 | `HFS_REQUEST_TIMEOUT` | 30 | Request timeout (seconds) |
 | `HFS_DEFAULT_PAGE_SIZE` | 20 | Default search result page size |
 | `HFS_MAX_PAGE_SIZE` | 1000 | Maximum search result page size |
+
+#### HTTP compression
+
+All three HTTP servers (`hfs`, `sof-server`, `hts`) accept request bodies sent
+with `Content-Encoding: gzip` (also `deflate`, `br`, `zstd`); unsupported
+encodings are rejected with `415`. Responses are compressed when the client
+sends `Accept-Encoding` (with `Content-Encoding` and `Vary: Accept-Encoding`
+set). SOF never re-compresses `application/parquet` / `application/zip`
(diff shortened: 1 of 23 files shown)


smunini (approved):

smunini:
Thank you @darthcav - excellent addition!  I'm also going to make this change to the fhirpath-server in a separate branch.

PR #148, opened 2026-06-11
feat(hts): multilingual SNOMED CT support with Accept-Language handling

darthcav:
## Summary

The SNOMED CT RF2 importer previously discarded every non-English description, so imported concepts carried no `concept.designation` entries and `displayLanguage` requests against `http://snomed.info/sct` could not resolve translated terms. This PR adds full multilingual SNOMED support, following the approach already used by the LOINC linguistic-variant importer, and completes `Accept-Language` header handling across the terminology operations.

## Importer (`snomed_rf2.rs`)

- **All Description files are read** — national editions ship one file per language (e.g. `sct2_Description_Snapshot-de_*.txt`) — with last-state-wins semantics per description id.
- **Every active description of an active concept becomes a FHIR `concept.designation`**, tagged with the RF2 `languageCode` and a `use` Coding carrying the SNOMED description type (FSN / synonym).
- **Language reference sets are now parsed** and drive:
  - display selection — US-preferred English synonym, then GB-preferred, then any preferred, then the previous first-synonym/FSN fallback chain;
  - designation ordering — synonyms before FSNs, refset-preferred terms first, so language-keyed first-match lookups resolve to the preferred term;
  - dialect-tagged `en-US` / `en-GB` `preferredForLanguage` designations for refset-preferred synonyms.
- **Full/Snapshot separation fix:** production RF2 ZIPs ship `Full/` and `Snapshot/` side by side; the Concept, Relationship, and association-refset selection previously took the last matching ZIP entry, which could pick a Full file and ingest historical duplicates or resurrect inactivated relationships. All five file selections now prefer Snapshot variants with deterministic ordering.

The designation storage and the `displayLanguage` handling in `$lookup` / `$expand` / `$validate-code` already supported designations end-to-end on both SQLite and PostgreSQL, so they pick these up without schema changes.

## Accept-Language

- The header is now injected as `displayLanguage` (only when no explicit parameter is given) on all four `$lookup` handlers and the GET and instance-level `$validate-code` handlers, matching the behavior `$expand` and the POST `$validate-code` handlers already had.
- `Accept-Language` was added to the CORS `allow_headers` list used when explicit origins are configured (the permissive `*` default already mirrored it), so browser clients can send the header cross-origin.

## Cross-version designation fallback in `$lookup`

When several versions of one canonical URL coexist — e.g. the SNOMED International release alongside a national edition imported via the bootstrap directory — unversioned requests resolve to the highest version string. `$expand` and `$validate-code` already match designations URL-wide across versions, but `$lookup` read designations only from the resolved version's concept row, so a German term carried by an older national edition was invisible once a newer international release was imported.

`$lookup` (SQLite and PostgreSQL) now falls back when no version is pinned and the resolved version has no designation in the requested language: matching designations are pulled from the newest other stored version of the same URL, preserving insertion order so refset-preferred terms stay first. An explicit `version` pin disables the fallback. Lookup response caches already key on `(version, displayLanguage)` and are cleared on import, so caching stays correct.

## Example

```bash
hts import ./SnomedCT_GermanyEdition.zip --format snomed-rf2

curl -X POST http://localhost:8090/CodeSystem/\$lookup \
  -H "Content-Type: application/fhir+json" \
  -H "Accept-Language: de" \
  -d '{"resourceType":"Parameters","parameter":[{"name":"system","valueUri":"http://snomed.info/sct"},{"name":"code","valueCode":"22298006"}]}'
```

## Testing

- 600 `helios-hts` lib tests pass, including new coverage for: multilingual description parsing, language-refset display preference, designation ordering, Full+Snapshot archive handling, end-to-end import + `displayLanguage` lookup, Accept-Language header injection (POST/GET, explicit-parameter precedence), and the cross-version fallback (fallback hit, version-pin respected, newest-other-version-wins).
- PostgreSQL integration test for the cross-version fallback runs green against a testcontainers Postgres.
- `cargo fmt` and `cargo clippy` (CI flags) clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -608,6 +608,12 @@ Load terminology packages directly from the filesystem without going through the
 cargo run --bin hts -- import ./hl7.terminology.r4-6.0.0.tgz
 
 # SNOMED CT RF2 ZIP  ⚠️  requires NRC license
+# Descriptions in all languages (incl. per-language Description files from
+# national extensions) are imported as FHIR concept.designation entries
+# tagged with the RF2 language code; language refsets pick the preferred
+# synonym for display (en-US, then en-GB) and add dialect-tagged
+# preferredForLanguage designations. Query via displayLanguage or the
+# Accept-Language header on $lookup / $expand / $validate-code.
 cargo run --bin hts -- import ./SnomedCT_InternationalRF2_*.zip --format snomed-rf2
 
 # LOINC CSV ZIP  ⚠️  requires free Regenstrief registration at loinc.org
--- a/crates/hts/README.md
+++ b/crates/hts/README.md
@@ -768,6 +768,16 @@ curl -X POST http://localhost:8090/CodeSystem/\$lookup \
   -d '{"resourceType":"Parameters","parameter":[{"name":"system","valueUri":"http://snomed.info/sct"},{"name":"code","valueCode":"22298006"}]}'
 ```
 
+Multiple editions of the same code system (e.g. the SNOMED International
+release plus a national edition) coexist as separate versions keyed on
+`(url, version)`; unversioned requests resolve to the highest version.
+When the requested `displayLanguage` has no designation in that version,
+`$lookup` falls back to the newest *other* stored version that carries the
+language (`$expand` and `$validate-code` match designations across versions
+of the canonical URL by design) — so a German term from a national edition
+is still found when a newer international release is the default. Pinning
+`version` disables the fallback.
+
 ---
 
 ### LOINC
--- a/crates/hts/README.md
+++ b/crates/hts/README.md
@@ -750,6 +750,24 @@ hts import ./SnomedCT_InternationalRF2_PRODUCTION_20250901T120000Z.zip --format 
 
 For large imports, add `--batch-size 200 --verbose` to monitor progress.
 
+Descriptions in **all languages** present in the archive (including the
+per-language Description files shipped by national extensions, e.g.
+`sct2_Description_Snapshot-de_*.txt`) are imported as FHIR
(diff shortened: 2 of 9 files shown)


smunini (approved):

smunini:
Thank you for this contribution, @darthcav — the multilingual import, the Full/Snapshot separation fix, and the cross-version designation fallback are all excellent work.

We've landed your three commits exactly as authored (re-signed to satisfy our commit-signing policy) in #149, together with a follow-up commit that broadens the language support beyond English and German to the rest of the languages SNOMED supports:

1. **BCP-47 / RFC 4647 fallback matching** for `displayLanguage` across `$lookup` / `$expand` / `$validate-code`: a browser sending `Accept-Language: de-DE,de;q=0.9` now resolves designations stored under the bare RF2 tag `de` (and conversely `fr` accepts `fr-CA`). Previously only `en-US`/`en-GB` region-qualified requests resolved, since only those dialect designations were emitted.
2. **`LANG_REFSET_DIALECTS` extended** to the published national-edition language refsets (da-DK, en-AU, en-CA, en-IE, en-NZ, es-UY, et-EE, fr-BE, fr-CA, fr-FR, nb-NO, nn-NO, nl-BE, nl-NL, sv-SE), with SCTIDs taken from SNOMED International's Snowstorm dialect configuration.
3. **`preferredForLanguage` designations for every refset-preferred synonym** (tagged with its bare RF2 language, plus the dialect tag when known), so editions whose refset isn't in the table — German, Spanish, Swiss, … — get explicit preference metadata rather than relying on designation ordering.

Because the commits were re-signed, GitHub won't mark this PR as merged — it will close automatically when #149 lands, but your authorship is preserved on all three commits. Thanks again, and we'd be glad to see more contributions from you!

PR #150, opened 2026-06-11
feat(observability): uptime, Prometheus /metrics, and OTLP traces across all servers

mauripunzueta:
## Summary

Adds a shared `helios-observability` crate and wires it into all four server
binaries (`hfs`, `hts`, `sof-server`, `fhirpath-server`):

- `GET /metrics` Prometheus endpoint — `http_requests_total`,
  `http_request_duration_seconds`, `uptime_seconds`, with a `service` global
  label and a templated `route` label.
- `/health` enriched with `uptime_seconds` + `started_at`.
- Per-request metrics + a tracing span. Tenant is a span attribute only, never a
  metric label (cardinality).
- Feature-gated (`otel`) OTLP trace export via `tracing-opentelemetry`
  (opentelemetry 0.32). OTLP metrics are produced out-of-process by a Collector
  scraping `/metrics` — we avoid the unmaintained `opentelemetry-prometheus`
  (protobuf RUSTSEC advisory).

## Verification
- Workspace `cargo check` (default) compiles; observability also compiles with
  `--features otel`. Clippy clean (CI flags) for touched crates. Unit tests pass.
- End-to-end smoke test against `fhirpath-server`: `/health` uptime + `/metrics`
  counter/gauge/histogram confirmed.

## Notes / known gaps in local verification
1. Full HFS server not run locally — SQLite WAL fails over the WSL filesystem
   under the Windows toolchain (environment limitation, not a code issue). The
   stateless `fhirpath-server` exercised the same wiring successfully.
2. Clippy not run with `--all-features` locally (build window). A pre-existing
   `dead_code` warning on `build_embedded_job_store` appears only under default
   features; CI's `--all-features` run won't hit it. CI is the authoritative gate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -300,6 +300,14 @@ jobs:
         # helios-serde's own quick-xml (client FHIR XML) is already on 0.41 and
         # sits behind the non-default `xml` feature. Remove these once
         # object_store and octofhir-ucum ship builds on quick-xml >= 0.41.
+        #
+        # crossbeam-epoch 0.9.18 (transitive; present on main identically).
+        # RUSTSEC-2026-0204 is an invalid-pointer dereference in the
+        # `fmt::Pointer` impl for `Atomic`/`Shared`, reachable only when
+        # debug-formatting an already-invalid epoch pointer — we never format
+        # these types. 0.9.18 is the latest release; no fixed version exists yet.
+        # Ambient advisory (published 2026-07-06, after main's last green run).
+        # Remove once crossbeam-epoch ships a patched release.
         run: >-
           cargo audit
           --ignore RUSTSEC-2026-0118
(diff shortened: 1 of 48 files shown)


smunini:
@claude and @mauripunzueta how will this work if hfs is run in a clustered environment?

mauripunzueta:
Good question @smunini. The resource numbers (`resource-counts`, `activity`, `resource-distribution`, and the per-tenant counts) read from the shared database, so every instance returns the same fleet-wide numbers. The live traffic/latency and uptime come from in-process state, so those are per-instance — I've labeled them explicitly in the responses (a `scope` field, and an `instance` on the authenticated ones) so they can't be mistaken for fleet-wide numbers. For cluster-wide traffic/latency you aggregate the per-instance `/metrics` in Prometheus/Grafana (I configured histogram buckets so p95 aggregates correctly across instances). Auth and tenant resolution are stateless, so no shared session state — any instance can serve any request.

I added a "Clustering / multi-instance" section to `docs/console-metrics.md` with the details. TL;DR: DB-backed endpoints are cluster-wide; the in-process traffic/uptime views are per-instance (use Prometheus for the fleet view).

smunini (approved):

PR #151, opened 2026-06-11
feat: per-user JSON UI settings store

mauripunzueta:
## Summary

Adds a simple, extensible **per-user settings store** for the upcoming web UI. Settings like dark mode are just the first of many — the stored value is an **opaque JSON object**, so new keys need zero schema or code changes (the frontend owns the document shape). Example settings: theme/dark mode, recent FHIR queries by resource type, default tenant, active FHIR version.

The design is **hybrid**: a server-side store is the source of truth (settings roam across devices), intended to be fronted by a client-side `localStorage` cache for instant first paint.

## Design decisions

- **Dedicated `user_settings` table/collection**, kept separate from the FHIR `resources` store so private UI prefs never leak into `CapabilityStatement`, `_history`, `_search`, or `$export`. Schema migration for SQLite (v12→v13), PostgreSQL (v12→v13), and MongoDB (v4→v5).
- **`SettingsStore` trait** (`get`/`put`/`patch`) in `helios-persistence` core + an RFC 7386 JSON merge-patch helper. Implemented for SQLite, PostgreSQL, and MongoDB with **read-modify-write** and a monotonic `version` for optimistic locking (surfaced as a weak `ETag`).
- **Keyed by user only** (`issuer|subject`), tenant-independent — a setting like "default tenant" is inherently cross-tenant. Falls back to a fixed `local|default` key when auth is disabled.
- **REST endpoints** `GET`/`PUT`/`PATCH /_user/settings`. The leading-underscore path keeps them authenticated yet **exempt from FHIR scope checks** and invisible to FHIR machinery. `PUT` replaces; `PATCH` merge-patches (great for toggling one key); both honor `If-Match`, `GET` honors `If-None-Match` (304). `GET` returns `{}` by default so the UI always gets a usable document.

## Backend coverage

The store is implemented on **every standalone primary backend** that offers the required read-modify-write + monotonic-version primitives:

| Backend | Status | Concurrency mechanism |
|---------|--------|-----------------------|
| SQLite | ✅ Implemented | Transactional read-modify-write |
| PostgreSQL | ✅ Implemented | `SELECT … FOR UPDATE` read-modify-write |
| MongoDB | ✅ Implemented | Version-conditioned update + retry (no replica set required) |
| S3 | ⏳ Tracked follow-up (#199) | Needs conditional `PutObject`; recommended role is Archive |
| Elasticsearch | ➖ N/A (correct 501) | Search-only, never a standalone primary |

On unsupported backends the endpoints return `501 Not Implemented` with a message that **names the supported backends and explicitly excludes S3/Elasticsearch**, so operators get an explained response rather than a bare 501.

### MongoDB implementation notes

MongoDB standalone deployments don't support multi-document transactions, so rather than a `SELECT … FOR UPDATE` the write path reads the current document, computes the new one, then persists with a filter pinned to the read version. A concurrent writer that bumped the version makes the update match nothing — retried for unconditional writes, surfaced as an `OptimisticLockFailure` for `If-Match` writes. A `Some(0)` "must-not-exist" precondition maps to an insert whose duplicate-key error (from a new unique `user_key` index) becomes the same lock failure. This gives identical semantics to the relational stores without requiring a replica set.

## API

| Method | Path | Behavior |
|--------|------|----------|
| GET | `/_user/settings` | Fetch document (defaults to `{}`), `ETag: W/"{version}"` |
| PUT | `/_user/settings` | Replace whole document (JSON object) |
| PATCH | `/_user/settings` | RFC 7386 merge-patch a subset of keys |

## Tests

- Persistence unit tests: merge-patch semantics + SQLite store (round-trip, version increment, patch merge/delete, optimistic lock).
- PostgreSQL integration tests (testcontainers).
- MongoDB integration tests (gated on `HFS_TEST_MONGODB_URL`): missing→None, put/get/version, patch merge+delete, patch-creates, optimistic lock, and a concurrent-patch test asserting no lost updates.
- REST `UserKey` extractor unit tests.
- End-to-end HTTP suite (8 tests): default `{}`, round-trip, single-key merge, null-delete, stale `If-Match` → 412, matching `If-Match`, non-object body → 400, `If-None-Match` → 304.

## Notes for review

- `defaultTenant` is keyed per-user globally (deliberate — it's cross-tenant). Future per-tenant settings can nest as `perTenant: {...}` without changing the store key.
- The frontend `localStorage` cache layer is out of scope here (UI team owns it).
- **S3 support** is deferred to #199 — its archival role and object-store concurrency/version story differ from a transactional primary; the follow-up scopes the conditional-write work required.


smunini (changes requested):
The design looks good, but this needs to support **all of our primary (standalone) storage backends**, not just SQLite and PostgreSQL. Right now MongoDB and S3 fall through to `501`, and both are first-class standalone primary FHIR stores per `HFS_STORAGE_BACKEND` (`sqlite`, `postgres`, `mongodb`, `s3` — see `crates/hfs/src/main.rs`).

**Blocking:**
- **MongoDB** — a full standalone primary backend (`ResourceStorage`, `VersionedStorage`, `SearchProvider`, `ConditionalStorage`, history, `BundleProvider`, bulk export/submit). A `mongodb`-only deployment would run a working FHIR server whose UI settings silently 501. It has the transaction/optimistic-locking primitives to implement the same read-modify-write + monotonic `version` used in the SQLite/Postgres stores. Please add a `SettingsStore` impl.

**Please address or explicitly track:**
- **S3** — also offered standalone (`HFS_STORAGE_BACKEND=s3`, CRUD/versioning/history). Weaker case since its recommended role is Archive and the object-store concurrency/`version` story differs, but a `mongodb`/`s3` user shouldn't hit an unexplained 501. Either implement it or file a tracked follow-up rather than leaving an implicit 501.

Elasticsearch returning 501 is correct — it's search-only and never a standalone primary, so no action needed there.

Requesting changes for the MongoDB gap.

mauripunzueta:
Thanks @smunini. Addressed in fe36ad641:

- **MongoDB** — added a `SettingsStore` impl for `MongoBackend` (blocking gap closed). Standalone mongod has no multi-document transactions, so it uses a version-conditioned update: read → compute → write with a filter pinned to the read version, retrying on an unconditional race and returning `OptimisticLockFailure` for `If-Match`. A `Some(0)` precondition maps to an insert guarded by a new unique `user_key` index. Same read-modify-write + monotonic-version semantics as the SQLite/Postgres stores, no replica set required. Wired into `start_mongodb` and `start_mongodb_elasticsearch`, with an integration suite (incl. a concurrent-patch no-lost-updates test).
- **S3** — filed as tracked follow-up #199 rather than implemented: its archival role and object-store concurrency/version story differ, and correct optimistic locking needs conditional `PutObject`. The `/_user/settings` 501 message now names the supported backends and explicitly excludes S3/Elasticsearch, so it's no longer an unexplained 501.
- **Elasticsearch** — left at 501 as you noted (search-only).

crates/hfs/src/main.rs
```diff
@@ -504,32 +512,33 @@ async fn start_mongodb(
     let backend = Arc::new(backend);
     let serve_audit_state = audit_state.clone();
 
-    // MongoDB primary; embedded SQLite sidecar for job state.
-    #[cfg(feature = "sqlite")]
-    {
-        let jobs = build_embedded_job_store(&config)?;
```
smunini, line 508: @claude why is the embedded job store deleted in this PR?  Seems to be not related to user ui settings?

crates/hfs/src/main.rs
```diff
@@ -1464,29 +1464,20 @@ async fn start_postgres(
     let backend = Arc::new(backend);
 
     let serve_audit_state = audit_state.clone();
+    // The PostgreSQL backend also hosts the per-user settings store, so it always
```
smunini, line 1486: @claude Is this true for other backend types?

crates/hfs/src/main.rs
smunini: It isn't deleted — `build_embedded_job_store(&config)?` is still called here (now a few lines down in `start_mongodb`, and in `start_mongodb_elasticsearch`), and the function itself is unchanged. What this hunk removed is only the surrounding control-flow scaffolding.

Why it got restructured, and why it *is* related to the settings work: the old MongoDB path had two mutually-exclusive branches, and the fall-through called `create_app_with_auth(Arc::try_unwrap(backend).unwrap_or_else(|_| unreachable!("... uniquely owned ...")), ...)`. That builder takes the backend **by value** and relies on `Arc::try_unwrap` succeeding — i.e. the backend `Arc` must have a refcount of exactly 1.

The new settings store needs a *second* shared handle to that same backend (`let settings_store = Some(backend.clone());`), which makes `Arc::try_unwrap(backend)` fail. So both branches had to collapse into a single path that always uses the `Arc`-taking builder `create_app_with_auth_bulk_and_settings(backend, …, export_bundle, None, settings_store)`, with the export bundle (and the embedded job store inside it) computed into a local that's `None` when bulk export is off.

Net: ownership-driven refactor forced by the settings store needing a shared `Arc`, not a functional removal — same embedded-job-store behavior as before.

crates/hfs/src/main.rs
smunini: Partly true — and this comment caught a real gap, now fixed in 57e1b09b.

Per-startup-path status of the settings store on this branch:

| Startup fn | Primary | Settings store wired? |
|---|---|---|
| `start_sqlite` | SQLite | ✅ |
| `start_postgres` | Postgres | ✅ |
| `start_mongodb` | MongoDB | ✅ |
| `start_mongodb_elasticsearch` | Mongo (+ES) | ✅ (from the `mongo` primary) |
| `start_sqlite_elasticsearch` | SQLite (+ES) | ❌ → **now ✅** |
| `start_postgres_elasticsearch` | Postgres (+ES) | ❌ → **now ✅** |
| `start_s3` / `start_s3_elasticsearch` | S3 | ❌ (deferred to #199) |

The S3 and Elasticsearch-only cases are intentional. But the two composite paths were the inconsistency you're pointing at: `start_sqlite_elasticsearch` and `start_postgres_elasticsearch` use the *exact same* settings-capable `sqlite`/`pg` backends (they even pass `sqlite.clone()`/`pg.clone()` into `build_bulk_export`), yet they fell through to the plain `create_app_with_auth` builder with no settings store — so a `HFS_STORAGE_BACKEND=sqlite`+ES or `postgres`+ES deployment returned **501 on `/_user/settings`** despite the primary supporting it, unlike the analogous `start_mongodb_elasticsearch`.

57e1b09b wires the settings store from the primary in both composite paths (mirroring the Mongo+ES treatment) and narrows the now-unused import gates. Verified compiling across `sqlite`, `sqlite+es`, `postgres+es`, `s3+es+sqlite`, and `s3`.

smunini (approved):

PR #159, opened 2026-06-14
build: drop stale crates/hts workspace exclude

darthcav:
## Problem

The root workspace manifest excludes `crates/hts` with a comment claiming the crate has no `Cargo.toml`, while simultaneously listing `crates/hts` in `default-members`. The crate *does* have a manifest, so the `exclude` wins and `helios-hts` silently drops out of the workspace:

```
$ cargo build -p helios-hts
error: package ID specification `helios-hts` did not match any packages
help: a package with a similar name exists: `helios-hfs`
```

The same failure hits `cargo test -p helios-hts`, `cargo run --bin hts`, and any other root-level invocation that names the crate.

## Why it went unnoticed

`hts.yml` only triggers on `workflow_dispatch`, `schedule`, and `workflow_call` — never on `push`/`pull_request`. So the broken workspace resolution from a fresh `main` checkout is not exercised by normal PR CI.

## Fix

Remove the contradictory `exclude` block so `helios-hts` resolves as a workspace member again, and update `Cargo.lock` for the now-included crate (`helios-hts` plus its `tar`/`ctor` dependencies).

## Verification

- `cargo metadata --locked` resolves cleanly
- `cargo build -p helios-hts` ✓
- `cargo test -p helios-hts` ✓ (all tests pass)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2071,6 +2071,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "ctor"
+version = "0.2.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "32a2785755761f3ddc1492979ce1e48d2c00d09311c39e4466429188f3dd6501"
+dependencies = [
+ "quote",
+ "syn 2.0.117",
+]
+
+[[package]]
 name = "ctr"
 version = "0.9.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


smunini:
Already resolved in `main` via #152, which included the identical `build: drop stale crates/hts workspace exclude` commit. Current main has no `exclude` in `Cargo.toml`, `helios-hts` (+ `tar`/`ctor`) in `Cargo.lock`, and `cargo metadata --locked` passes. Closing as redundant — thanks!

PR #166, opened 2026-06-15
feat(fhirpath): implement resolve() and fix resource visibility in lambdas

smunini:
## Summary

Implements the FHIR-flavored FHIRPath `resolve()` function and fixes a
pre-existing limitation it surfaced (root resources were invisible inside
`where()`/`select()` lambda bodies).

### `resolve()`
Dereferences a `Reference` (or reference/canonical string) to the resource it
points to. **In-scope only** — no external/DB I/O:

- **contained** resources — matched by `#fragment` or bare `id`
- resources already in `EvaluationContext` — matched by `Type/id`, or the
  trailing `Type/id` of an absolute URL
- **unresolved** `Type/id` references return a typed stub `{resourceType: Type}`
  so `resolve() is Observation` type filters still work; bare/fragment misses
  return `{}`
- collections of references are mapped and flattened

New module `crates/fhirpath/src/resolve_function.rs`; dispatch arm + handled-functions
entry in `evaluator.rs`. README updated.

### `%resource` / `%rootResource`
Added both environment variables (both resolve to the root resource in scope,
mirroring `%context`).

### Fix: resources now survive context cloning
`EvaluationContext::clone()` set `resources: Vec::new()` because `FhirResource`
isn't `Clone`. Lambda bodies (`where`, `select`, …) run in **child** contexts, so
`%context`, `%resource`, and `resolve()` all returned empty inside them.

Fix: `resources: Arc<Vec<FhirResource>>`. `Arc` clones by ref-count without
requiring `T: Clone`, so the root resources stay visible through cloning and the
whole parent/child chain. Field reads are unchanged thanks to `Deref`.

### Test data
- Added `crates/fhirpath/tests/data/r5/input/diagnosticreport-eric.json` (from
  [FHIR/fhir-test-cases](https://github.com/FHIR/fhir-test-cases/blob/master/r5/diagnosticreport-eric.json)).
  The R5 harness resolves `inputfile=` only from `tests/data/r5/input/`, and this
  file (which lives at top-level `r5/` upstream, not `r5/examples/`) was never staged.
- `known-test-failures.json`: removed the stale `testMultipleResolve` entry (now
  **passes**) and the `testContainedId` entry (its stale reason wrongly blamed
  resolve; it remains auto-skipped by the harness for an unrelated "Unsupported
  output type: id" reason).

## Testing
- 7 new unit tests in `resolve_function.rs`
- Full `helios-fhirpath` suite (default/R4) and `helios-sof` suite pass
- R5 conformance suite green (0 failures); `testMultipleResolve` now passes
- Workspace builds clean

### Note
`resolve()` does **not** fetch from storage — references to server-stored
resources not present in the FHIRPath context resolve to a typed stub. A
storage-backed resolver is deliberately out of scope here and is tracked as a
follow-up in #167.

### DIFF
--- a/crates/fhirpath/README.md
+++ b/crates/fhirpath/README.md
@@ -458,7 +458,7 @@ These functions extend the base FHIRPath specification with FHIR-specific capabi
     *   [extension()](https://build.fhir.org/fhirpath.html#functions): ✅ (Full support with variable URL resolution)
     *   [hasValue()](https://build.fhir.org/fhirpath.html#functions): ✅ (Tests if primitive has actual value beyond extensions)
     *   [getValue()](https://build.fhir.org/fhirpath.html#functions): ❌ Not Implemented
-    *   [resolve()](https://build.fhir.org/fhirpath.html#functions): ✅ (In-scope resolution: contained resources and resources in the evaluation context; returns a typed stub for unresolved `Type/id` references. No external/DB resolution.)
+    *   [resolve()](https://build.fhir.org/fhirpath.html#functions): ✅ (In-scope resolution: contained resources and resources in the evaluation context; returns a typed stub for unresolved `Type/id` references. No external/DB resolution — see [#167](https://github.com/HeliosSoftware/hfs/issues/167) for the storage-backed resolver follow-up.)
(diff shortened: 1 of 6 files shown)


smunini:
@darthcav - per our conversation today.

PR #174, opened 2026-06-16
feat(sof): cross-resource and remote resolve() for SQL-on-FHIR

darthcav:
## Summary

Implements the SQL-on-FHIR `resolve()` function for **cross-resource** reference resolution and adds **opt-in remote resolution** against trusted FHIR servers, then exposes both through the CLI, HTTP server, and Python bindings.

Before this change, `resolve()` in SQL-on-FHIR only saw the single resource being processed plus its `contained` children. A reference such as `Encounter.subject.resolve()` pointing at a **separate** `Patient` in the same bundle did not resolve — it fell back to a typed stub and produced null columns. This was confirmed empirically with `sof-cli` (sibling Patient → null; contained Patient → resolves) before the fix.

---

## 1. Bundle-level resolution

A `ViewDefinition` can now follow a `Reference` to **any** resource in the input bundle, regardless of type, not just the resource under evaluation and its `contained` children.

- **`EvaluationContext::set_resolution_scope`** (helios-fhirpath) swaps the resource pool that `resolve()` searches, while leaving `this` — and therefore `%context` / `%resource` / `%rootResource` and the evaluation root — unchanged. Those consult `this` first and only fall back to `resources[0]`, so widening the pool is safe.
- **`build_resolution_scope`** (helios-sof) parses the whole input bundle once (all resource types, before type filtering) into a shared `Arc<Vec<FhirResource>>` and installs it on every evaluation context created during execution: per-resource row generation, `where` clauses, and the per-iteration `forEach` / `repeat` contexts.
- Streaming/NDJSON keeps in-bundle resolution scoped to the current chunk.

**Why:** the SQL-on-FHIR v2 `resolve()` semantics require resolving references across the resources being processed; the prior per-resource scope made the common Encounter→Patient style of view impossible.

## 2. Remote resolution against trusted servers (opt-in, default-off)

References whose absolute URL points at an **explicitly allowlisted** FHIR server can be fetched over the network and folded into the resolution pool.

- **Prefetch architecture.** The FHIRPath evaluator is synchronous and runs under Rayon; rather than make `resolve()` itself perform I/O, external references are collected and fetched **at the async edge before** row generation, then injected into the pool. The parallel evaluation core stays pure and a run remains reproducible from `(bundle + fetched snapshot)`.
- **Default-deny allowlist.** A reference is fetched only if it is an absolute `http`/`https` URL matching an allowlist entry on **scheme + host + port + path-prefix** (parsed-URL comparison, never substring — so `https://evil/?u=https://trusted.org` does not match).
- **SSRF / DNS-rebinding guard.** Hostnames are resolved once and the HTTP client is pinned to the validated addresses. Loopback, link-local (including the `169.254.169.254` cloud-metadata endpoint), multicast, broadcast, CGNAT and reserved ranges are always blocked. Private RFC1918 / IPv6-ULA targets are blocked **unless** `SOF_RESOLVE_ALLOW_PRIVATE_ADDRESSES` is set, which lets an allowlisted hostname resolve to an internal load balancer / reverse proxy (e.g. Traefik) — loopback and link-local stay blocked even then.
- **Other hardening:** redirects disabled; optional per-host bearer auth; per-request timeout, response-size and total-fetch caps; bounded-depth chained resolution. Any failure (disallowed host, timeout, non-2xx, oversize, unparseable) is non-fatal and falls back to the typed-stub/empty behaviour.
- **Streaming/NDJSON:** a single resolver spans the whole stream with a bounded **LRU cross-chunk cache** (a reference recurring across chunks is fetched once) and `SOF_RESOLVE_MAX_FETCHES` applied as a **per-stream** cap.

### Configuration

`SOF_RESOLVE_*` environment variables and `sof-cli` flags; the HTTP server reads the same env config.

| Variable | Default | Meaning |
|----------|---------|---------|
| `SOF_RESOLVE_REMOTE` | `false` | Master switch. |
| `SOF_RESOLVE_ALLOWED_BASE_URLS` | *(empty)* | Comma-separated trusted base URLs. |
| `SOF_RESOLVE_ALLOW_PRIVATE_ADDRESSES` | `false` | Permit allowlisted hostnames to resolve to private/internal addresses. |
| `SOF_RESOLVE_TIMEOUT_MS` | `5000` | Per-request timeout. |
| `SOF_RESOLVE_MAX_FETCHES` | `256` | Fetch cap (per-stream when streaming). |
| `SOF_RESOLVE_MAX_DEPTH` | `1` | Chained-reference rounds. |
| `SOF_RESOLVE_MAX_RESPONSE_BYTES` | `5000000` | Response size cap. |
| `SOF_RESOLVE_CONCURRENCY` | `8` | Max concurrent fetches. |
| `SOF_RESOLVE_CACHE_MAX_ENTRIES` | `10000` | Streaming cross-chunk cache size. |
| `SOF_RESOLVE_AUTH` | *(none)* | Per-host bearer tokens (`host=token,...`). |

CLI: `--resolve-remote`, `--resolve-allowed-base-urls`, `--resolve-allow-private-addresses`.

**Why:** enables cross-server joins for deployments that explicitly trust specific FHIR endpoints, without compromising the safe default (it is entirely off unless configured) and without turning the evaluator into a network client.

## 3. Python bindings (pysof)

Adds `RemoteResolveConfig`, `run_view_definition_remote`, and `process_ndjson_to_file_remote`. The async remote functions are driven from the synchronous PyO3 entry points via a current-thread Tokio runtime with the GIL released. All additions are additive; existing entry points are unchanged. The wrapper package, type stub, and README are updated.

## 4. Dependencies — `lru`

Adds `lru` for the streaming cross-chunk cache, **pinned to `0.12`**. `aws-sdk-s3` (pulled by the optional `s3` feature in `helios-persistence` / `helios-rest`) declares `lru = "^0.12"` as a non-optional dependency; matching it keeps a **single** `lru` version compiled across all feature combinations, including `s3` builds. The latest `lru` is `0.18`, but raising the pin would compile two `lru` versions whenever `s3` is enabled, because the AWS SDK's `^0.12` (a third-party crate we don't control) is semver-incompatible with `0.18`. The upgrade to the latest `lru` is tracked in `ROADMAP.md` (Build & Dependencies) for re-evaluation once the AWS SDK bumps its own pin.

## Tests

- `crates/sof/tests/resolve_tests.rs` — in-bundle cross-resource resolution, missing refs, contained refs, full-bundle scope, absolute URLs, and `where` clauses.
- `crates/sof/tests/resolve_remote_tests.rs` — wiremock-backed: allowlist pass/deny, 404 non-fatal, fetch cap, chained depth, the private-address hostname rule, and streaming cross-chunk cache / per-stream cap.
- `remote_resolver` unit tests — allowlist matching (including substring/userinfo smuggling), the tiered SSRF IP guard for IPv4/IPv6, and config parsing.
- `crates/pysof/python-tests/test_remote_resolve.py` — symbol export, config construction, and an inactive-config end-to-end run (verified at runtime against a built extension).

All `helios-sof`, `helios-fhirpath`, `helios-rest`, and `helios-hfs` test suites pass; `cargo clippy` is clean for the changed crates.

## Documentation

- `book/src/ch06-sql-on-fhir.md` — a "Resolving References Across Resources" section and a "Remote resolution against trusted servers" subsection with the full config reference and security model.
- `book/src/appendix-b-fhirpath-functions.md` — corrected `resolve()` status.
- `crates/pysof/README.md` — remote-resolve usage.
- `ROADMAP.md` — follow-ups (logical/identifier-based references, storage-backed / cross-run cache, async FHIRPath evaluator, OAuth client-credentials, and the `lru` upgrade).

## Compatibility

All library API changes are additive. `resolve()` behaviour with remote resolution disabled (the default) is unchanged apart from the new bundle-level in-scope resolution. Streaming output for the non-remote path is unchanged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3350,6 +3350,7 @@ dependencies = [
  "helios-fhirpath",
  "helios-fhirpath-support",
  "http 1.4.0",
+ "lru",
  "mime",
  "object_store",
  "parking_lot",
@@ -3369,6 +3370,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "url",
+ "wiremock",
  "zip 2.4.2",
 ]
 
@@ -5474,6 +5476,7 @@ dependencies = [
  "pythonize",
  "serde",
  "serde_json",
+ "tokio",
 ]
 
 [[package]]
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -104,6 +104,15 @@ Longer-term ideas we are exploring. These are not yet committed and may evolve s
 - **SMART on FHIR** — Full launch framework (standalone launch, EHR launch) and fine-grained scoped access
 - **GraphQL API** — [FHIR GraphQL](https://hl7.org/fhir/graphql.html) support for resource retrieval, search, and graph traversal as an alternative to the REST API
 
+### SQL-on-FHIR & FHIRPath
+
+Follow-ups to `resolve()` in SQL-on-FHIR, which today dereferences references against the input bundle (and their `contained` resources) and can optionally fetch from explicitly allowlisted, trusted remote servers:
+
+- **Logical (identifier-based) reference resolution** — `resolve()` handles literal `Reference.reference` strings (relative `Type/id`, absolute URLs, and `#contained` fragments). References that identify their target only by a business `identifier` (no `reference` string) are not yet resolved.
+- **Storage-backed and cross-run resolution cache for `resolve()`** ([#167](https://github.com/HeliosSoftware/hfs/issues/167)) — Resolve against server-stored resources, and persist fetched resources across runs, as a layer over the current in-bundle pool and per-run in-memory prefetch cache.
+- **Async FHIRPath evaluator** — The FHIRPath engine is synchronous (executed under Rayon by the SQL-on-FHIR core), so remote `resolve()` is handled by an up-front prefetch pass rather than inline I/O. A general async evaluator would let effectful functions (inline remote resolution, terminology lookups) run during evaluation.
+- **OAuth client-credentials for remote `resolve()`** — Trusted-server authentication is per-host bearer tokens today; a client-credentials grant flow would better suit servers fronted by OAuth.
+
 ### Advanced Persistence
 
 - Cassandra as a primary store
@@ -119,6 +128,20 @@ An intelligent recommendation engine for storage configuration:
 - Analyze a FHIR query and recommend an optimal persistence configuration
 - Leverage historical benchmark data to inform recommendations
 - Web UI for interactive configuration guidance
+
+### Build & Dependencies
+
+- **Upgrade `lru` to the latest version** — `helios-sof` uses `lru` for the
+  streaming remote-`resolve()` cross-chunk cache. It is currently pinned to `0.12`
+  to match `aws-sdk-s3`, which pins `lru = "^0.12"` as a non-optional dependency
+  (pulled by the optional `s3` feature in `helios-persistence` / `helios-rest`).
+  Matching the SDK keeps a single `lru` version compiled across all feature
+  combinations. The latest `lru` is 0.18; raising `helios-sof`'s pin would compile
+  two `lru` versions whenever `s3` is enabled, because the AWS SDK's `^0.12` pin (a
+  third-party crate we don't control) is semver-incompatible with 0.18. Bump
+  `helios-sof` to the latest `lru` once the AWS SDK updates its own pin, collapsing
+  back to a single version. Tracked so the pin is re-evaluated on the next AWS SDK
+  upgrade.
 
 ---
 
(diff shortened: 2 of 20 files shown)


darthcav:
@smunini I think yesterday that yesterday we misinterpreted the link resolution. It will not even allow in the actual version the resolution of resources in the same server. I added that with a new PR, including some configurability for trusted network resources.

smunini:
Merged to `main` as commit 9d46722ee (rebased onto current main and GPG-signed by the committer, since this came from an external fork). Authorship preserved (@darthcav). Thanks for the contribution!

- All `helios-sof` + `helios-fhirpath` suites pass (incl. new `resolve_tests` and `resolve_remote_tests`).
- Security review of the remote-`resolve()` SSRF/allowlist design: solid defense-in-depth, no blockers.

smunini:
> Merged to `main` as commit [9d46722](https://github.com/HeliosSoftware/hfs/commit/9d46722eeecf3302aed59c0e4e4f847f8d43ded6) (rebased onto current main and GPG-signed by the committer, since this came from an external fork). Authorship preserved (@darthcav). Thanks for the contribution!
> 
> * All `helios-sof` + `helios-fhirpath` suites pass (incl. new `resolve_tests` and `resolve_remote_tests`).
> * Security review of the remote-`resolve()` SSRF/allowlist design: solid defense-in-depth, no blockers.

I un-wound this and merged it normally.

PR #178, opened 2026-06-18
feat(sof): sync upstream tests, implement %rowIndex, MongoDB conformance parity

smunini:
## Summary

Brings the SQL-on-FHIR implementation up to date with the relocated upstream test suite and closes the `%rowIndex` gap that showed red on https://sql-on-fhir.org/extra/impls.html, then brings the MongoDB in-DB runner to full conformance parity with SQLite/PostgreSQL.

Three threads, delivered together because they're interdependent:

### 1. Sync the upstream conformance fixtures
- The official tests moved to [`FHIR/sql-on-fhir.js`](https://github.com/FHIR/sql-on-fhir.js/tree/main/tests). Resynced all 22 fixtures **verbatim** + added the new `row_index.json` (9 cases).
- `crates/sof/tests/sql-on-fhir-v2/README.md` documents the upstream source + verbatim-sync policy.
- In-memory runners now tolerate a missing top-level `fhirVersion` (`constant_types.json` dropped it) and compare rows **order-insensitively**, matching the upstream compliance runner.

### 2. Implement `%rowIndex` across every SOF engine
- **Primary engine** (`run_view_definition`, used by sof-cli/sof-server, S3 in-process, ES): per-iteration index for `forEach`/`forEachOrNull`/`repeat`; **144/144** in-memory fixtures pass.
- **In-DB SQLite + PostgreSQL**: `json_each.key` / `WITH ORDINALITY` for `forEach`; pre-order `ord_path` + `DENSE_RANK` in the repeat CTE. Verified against real Postgres via testcontainers.
- Fixed second-precision time `highBoundary()`/`lowBoundary()` in the in-DB emitter (surfaced by the synced `fn_boundary` data).
- Removed obsolete `row_index` `KNOWN_SKIPS`; raised SQLite/PG `PASS_FLOOR` **124 → 132**.

### 3. MongoDB in-DB runner: 59 → 132 (parity)
New `crates/rest/tests/sof_conformance_mongodb.rs` (testcontainers, floor **132**). Emitter work:
- Array-guarded `[n]`/`.first()`; three-valued comparison (`null`, not `false`).
- `unionAll` via `$facet`; `where()`/`extension()` via `$filter` with a scoped `$$var`; `join()` + collection columns via `$reduce`; date/time/dateTime/decimal boundary.
- `repeat:` via a server-side-JS `$function` pre-order traversal + `$unwind`.
- Coerce serde_json `arbitrary_precision` number wrappers (which `bson::to_bson` stores as objects) to real numbers in comparison/arithmetic/boundary, preserving the digit-string for decimal-boundary precision.
- Flattened trailing-`[N]` `forEach` (`contact.telecom[0]`) → single full-path unnest, flattened before indexing.

The 12 remaining MongoDB failures are the **identical** structural gaps SQLite/PG also fail (nested/sibling `repeat`, `unionAll` nested inside another `select`).

## Verification
- MongoDB / PostgreSQL conformance: real DBs via testcontainers, **132** each
- SQLite conformance **132**; in-memory engine **144/144**; SQLite runner, sof unit tests (39), Mongo emit unit tests (9)
- Default (SQLite-only) build compiles; `cargo fmt`; clippy clean under CI flags

## Note
`repeat:` on MongoDB relies on **server-side JavaScript** (`$function`) — enabled by default, but disabled in some hardened/Atlas configs, where `repeat:` views would fail at runtime. Everything else uses native aggregation operators.


smunini:
Good catch — that README section was stale from before the parity work. Fixed in e69993fea: the MongoDB suite is now described as floor 132 with full `%rowIndex` (forEach via `$unwind`'s `includeArrayIndex`, repeat via `$function`), and the per-backend table no longer lists `unionAll`/`repeat`/collections as unsupported. All three in-DB suites now share floor 132 and the same 12 structural-gap failures.

Issue #168, opened 2026-06-15
Persistence-layer audit events for bulk export, purge, and reindex


smunini:
Wire audit logging for bulk export, purge, and reindex operations.

The audit functions already exist in `helios-audit`, but the REST endpoints for these persistence-layer operations do not yet emit `AuditEvent` records. This item completes the wiring so that bulk export, purge, and reindex operations produce IHE BALP-compliant audit events like the rest of the REST/auth/lifecycle paths already do.

**Scope**
- Emit `AuditEvent` for bulk export operations
- Emit `AuditEvent` for purge operations
- Emit `AuditEvent` for reindex operations

Tracked on the [ROADMAP](../blob/main/ROADMAP.md) under *Next → FHIR Server Capabilities*.

bomanaps:
Hello, please can I get assigned to this?

PR #179, opened 2026-06-22
feat(audit): emit events for export, purge, reindex

bomanaps:
## Summary                                                                          
  - Wires BALP-compliant `AuditEvent` emission for `$purge`, `$reindex`, and bulk     
  export — closing #168.                                                              
  - `$purge` and `$reindex` land as new REST handlers backed by an `OperationsBundle`,
   while the bulk-export worker now emits worker-complete / cancelled / failed events 
  at the actual lifecycle boundary instead of leaving them silent.                    
  - Thirteen integration tests assert action codes, outcomes, and detail fields       
  end-to-end via a new `InMemoryAuditSink` test fixture.                              
   
  ## Test plan                                                                        
  - [x] `cargo fmt --all -- --check`
  - [x] `cargo clippy --no-deps` on `helios-audit`, `helios-rest`,                    
  `helios-persistence`, `helios-hfs`                                                  
  - [x] `cargo test -p helios-audit -p helios-rest -p helios-persistence -p           
  helios-hfs` (1623 pass, 0 fail)                                                     
  - [x] Manual end-to-end run against the `hfs` binary with `HFS_AUDIT_BACKEND=file`,
  confirming 11 expected `AuditEvent` records emit across $purge, $reindex, and the   
  full bulk-export lifecycle (including worker-emitted completion)


bomanaps:
@smunini please can I get a review?

smunini:
### `$purge` / `$reindex` auth: no bypass, but `authz_middleware` misclassifies the type-scoped operations

I traced the auth path end-to-end. **No unauthenticated bypass** — but the middleware-level authorization misclassifies these new operations, which causes over-restriction for least-privilege tokens and a misleading audit trail.

**Authentication is solid.** The new paths aren't in `EXEMPT_PATHS`, so with auth enabled a missing/invalid token → 401 before the handler runs. When auth is disabled, `principal` is `None` and the ops are open — identical to ordinary `DELETE`/`PUT`, so consistent.

**Handler-level authz is correct and is the real backstop.** Each handler runs its own `SmartScopePolicy::check` with the right operation (purge → `Delete`, reindex → `Update`), and purge hard-rejects `AuditEvent` targets. This matches the inline-check pattern in `batch.rs` / `bulk_export.rs`, so there is **no under-enforcement / privilege-escalation hole** — a token lacking the scope gets 403 here regardless.

**The issue is `extract_operation_for_routing` in `crates/rest/src/middleware/auth.rs`.** It doesn't understand `$`-suffixed operation endpoints on a resource type — only `_search` is special-cased in the POST branch, so everything else POSTed to `/{rt}/...` falls through to `FhirOperation::Create`:

| Request | Middleware classifies as | Handler checks | Match? |
|---------|--------------------------|----------------|--------|
| `DELETE /Patient/123/$purge` | Delete on Patient | Delete | ✅ |
| `POST /Patient/$purge` | **Create** on Patient | Delete | ❌ |
| `POST /Patient/$reindex` | **Create** on Patient | Update | ❌ |
| `POST /$reindex` (system) | None (`$`-prefix skipped) | Update on `*` | ⚠️ not gated at middleware |

Consequences:

1. **Over-restriction for narrow scopes.** A token scoped exactly to what the handler documents as sufficient — `patient/Patient.d` for purge, `.u` for reindex — is blocked by the middleware's `Create` check before the handler runs. Only tokens carrying create/`write` scope get through, so the documented scope requirement isn't what's enforced end-to-end. (Too strict, not too loose — not a security hole, but it'll surprise least-privilege clients.)

2. **Mislabeled audit trail.** `authz_middleware` emits an AuditEvent reading `Granted/Forbidden: Create on Patient` for what is really a destructive purge or a reindex — misleading, and especially so in an audit-focused PR.

3. **Inconsistent enforcement surface.** Instance purge is gated correctly at the middleware, type purge/reindex are misgated as `Create`, and system reindex isn't gated at the middleware at all — three behaviors for one feature.

**Suggested fix** — pick one so middleware and handler agree:

- **(a)** Teach `extract_operation_for_routing` to map `.../$purge` → `Delete` and `.../$reindex` → `Update`; or
- **(b)** Return `None` for these paths (as it already does for `$export`, batch, and system `$`-ops) and rely solely on the handler check, dropping the misleading middleware audit event.

Option (b) is most consistent with the existing batch/bulk-export "defer to handler" pattern. Right now it accidentally does neither and lands on `Create`.

bomanaps:
> ### `$purge` / `$reindex` auth: no bypass, but `authz_middleware` misclassifies the type-scoped operations
> 
> I traced the auth path end-to-end. **No unauthenticated bypass** — but the middleware-level authorization misclassifies these new operations, which causes over-restriction for least-privilege tokens and a misleading audit trail.
> 
> **Authentication is solid.** The new paths aren't in `EXEMPT_PATHS`, so with auth enabled a missing/invalid token → 401 before the handler runs. When auth is disabled, `principal` is `None` and the ops are open — identical to ordinary `DELETE`/`PUT`, so consistent.
> 
> **Handler-level authz is correct and is the real backstop.** Each handler runs its own `SmartScopePolicy::check` with the right operation (purge → `Delete`, reindex → `Update`), and purge hard-rejects `AuditEvent` targets. This matches the inline-check pattern in `batch.rs` / `bulk_export.rs`, so there is **no under-enforcement / privilege-escalation hole** — a token lacking the scope gets 403 here regardless.
> 
> **The issue is `extract_operation_for_routing` in `crates/rest/src/middleware/auth.rs`.** It doesn't understand `$`-suffixed operation endpoints on a resource type — only `_search` is special-cased in the POST branch, so everything else POSTed to `/{rt}/...` falls through to `FhirOperation::Create`:
> Request 	Middleware classifies as 	Handler checks 	Match?
> `DELETE /Patient/123/$purge` 	Delete on Patient 	Delete 	✅
> `POST /Patient/$purge` 	**Create** on Patient 	Delete 	❌
> `POST /Patient/$reindex` 	**Create** on Patient 	Update 	❌
> `POST /$reindex` (system) 	None (`$`-prefix skipped) 	Update on `*` 	⚠️ not gated at middleware
> 
> Consequences:
> 
>     1. **Over-restriction for narrow scopes.** A token scoped exactly to what the handler documents as sufficient — `patient/Patient.d` for purge, `.u` for reindex — is blocked by the middleware's `Create` check before the handler runs. Only tokens carrying create/`write` scope get through, so the documented scope requirement isn't what's enforced end-to-end. (Too strict, not too loose — not a security hole, but it'll surprise least-privilege clients.)
> 
>     2. **Mislabeled audit trail.** `authz_middleware` emits an AuditEvent reading `Granted/Forbidden: Create on Patient` for what is really a destructive purge or a reindex — misleading, and especially so in an audit-focused PR.
> 
>     3. **Inconsistent enforcement surface.** Instance purge is gated correctly at the middleware, type purge/reindex are misgated as `Create`, and system reindex isn't gated at the middleware at all — three behaviors for one feature.
> 
> 
> **Suggested fix** — pick one so middleware and handler agree:
> 
>     * **(a)** Teach `extract_operation_for_routing` to map `.../$purge` → `Delete` and `.../$reindex` → `Update`; or
> 
>     * **(b)** Return `None` for these paths (as it already does for `$export`, batch, and system `$`-ops) and rely solely on the handler check, dropping the misleading middleware audit event.
> 
> 
> Option (b) is most consistent with the existing batch/bulk-export "defer to handler" pattern. Right now it accidentally does neither and lands on `Create`.


Went with (b), I made extract_operation return None for any $-op  path so it defers to the handler's SmartScopePolicy::check, and my tests show patient/Patient.d now reaches $purge with no misleading Create on <Type> event.

smunini:
I've force-pushed a rework of this branch. Your three original commits are preserved in the history, followed by a revert and a rebuild on a corrected design — so the attribution stays with you, but the implementation that lands is different. Apologies for rewriting the branch rather than asking for changes; the reshaping turned out to be structural enough that review comments weren't going to get us there. Here's what changed and why.

## 1. The in-memory audit sink came out

`InMemoryAuditSink` was a test double, and it was living in the production `helios-audit` crate as public API. Feature-gating it wouldn't have helped either: CI builds `--all-features --release`, so a gated test sink ships in the release binary anyway.

Event capture is now a `CollectorSink` defined in test code (`crates/rest/tests/persistence_operations.rs`, plus an inline copy for the persistence unit tests). `crates/audit/src/sinks/` is back to cloudwatch / database / file / null.

## 2. All backends are implemented, not just SQLite and Postgres

The branch returned 501 for MongoDB, Elasticsearch, S3, and every composite deployment. `$purge` and `$reindex` now work across all 8 storage modes.

The blocker was `ReindexableStorage`, which welded "read resources" and "write search entries" into one trait. That shape can only ever describe a backend that is both — which is why the non-SQL backends couldn't implement it. It's now split into `ReindexSource` + `ReindexTarget`, and `ReindexOperation` is non-generic (`Arc<dyn ReindexSource>` + `Vec<Arc<dyn ReindexTarget>>`), because a composite has *one* source and *N heterogeneous* targets.

That split also fixes a correctness bug that the 501s were masking. On a composite (e.g. SQLite + Elasticsearch), the branch wired both operations straight to the SQL primary:

- `$purge` deleted the row from SQLite but left the document in the Elasticsearch index — **purged data stayed searchable, with its full content**. Composite purge now hits secondaries *first*, and aborts before touching the primary if one fails, so a failure can't destroy the system of record while leaving a searchable copy behind.
- `$reindex` rebuilt SQLite's index tables while Elasticsearch — the thing actually serving search in that deployment — was never touched. It now fans out to every search index.

S3-standalone `$reindex` is the one remaining 501: it has no search index of any kind, so there is genuinely nothing to rebuild.

## 3. An authorization regression — and this one's on the review, not on you

Commit `4f1cc424` made `extract_operation` return `None` for *any* `$`-suffixed path, deferring to the handler. **That came from a review suggestion on this PR, not from your original submission** — I want to be clear about that, because it's the most serious thing on the branch.

The problem: the SQL-on-FHIR handlers (`crates/rest/src/handlers/sof/`) have no handler-level scope check. Only `batch.rs`, `bulk_export.rs`, and `bulk_submit.rs` self-check. So a blanket `$`-path exemption stripped the *only* authorization check from every SoF operation — any authenticated token could run arbitrary ViewDefinitions. Topic-based subscription operations (`$status`, `$events`, `$get-ws-binding-token`) lost their check too.

It's now a narrow three-entry allowlist (`HANDLER_AUTHORIZED_OPS`: `$purge`, `$reindex`, `$reindex-status`), with a regression test asserting `$viewdefinition-run` still requires its scope.

## Also worth knowing

- **`helios-audit` is now a required dep of `helios-persistence`**, not an optional `audit` feature. That gate is *why* this PR had to hand-roll a duplicate export-audit emitter in REST — the shared helper was literally unreachable from `helios-rest`. It also meant the coverage CI job built a different crate than the main CI job did. REST now delegates to the shared helper.
- **The audit helper signatures were wrong** and had zero callers, so fixing them was free: `record_purge_event` hardcoded `outcome("0")`, so a *failed* composite purge was indistinguishable from a successful one; `record_export_event` had no phase parameter, so a completed export and a cancelled one produced identical events.
- **`$purge` and `$reindex` require their own scopes** (`system/purge`, `system/reindex`), not ordinary Delete/Update scope. A token that may soft-delete a Patient must not thereby be able to obliterate it and its entire history. `$purge` on `AuditEvent` is refused unconditionally — a caller who can erase the audit trail can erase the evidence of their own actions.
- **`balp::detect_interaction`** now maps a `$`-suffixed final segment to `Execute`, so `POST /Patient/$purge` is no longer audited as a *Create* on Patient.
- **Known limitation, documented not fixed:** reindex job state is an in-process map, so `$reindex-status` 404s from any other node in a cluster.

## Verification

`cargo fmt --check` and `clippy --all-targets --all-features -D warnings` are clean. helios-audit (115), helios-persistence (708), helios-rest (351 + 12 new end-to-end tests), sqlite (78) all pass locally.

CI on this branch has been sitting at `action_required` — it needs a maintainer to approve the fork run, so the Docker-backed MongoDB/Elasticsearch integration tests haven't actually executed yet. Approving that run is the real check on the new backend implementations.

Thanks for the original patch — the audit helpers and the shape of the operation surface came from your work, and the rework kept both.

Issue #170, opened 2026-06-15
Cluster-aware Subscription notification delivery


smunini:
The Subscriptions engine tracks connected WebSocket clients in process memory (`crates/subscriptions/src/channels/ws_manager.rs`), so in a multi-instance deployment only the instance holding a client's connection can deliver its notifications.

Subscription resources themselves are database-backed; what's needed is a shared delivery channel (e.g., a message bus or pub/sub fan-out) so resource events on any instance reach clients connected to any other.

Until then, WebSocket subscription delivery is single-instance / sticky-session only (rest-hook channels are unaffected).

**Scope**
- Introduce a shared delivery channel (message bus / pub-sub fan-out) for subscription notifications
- Ensure resource events on any instance reach WebSocket clients connected to any other instance
- rest-hook channels are already cluster-safe and out of scope

Tracked on the [ROADMAP](../blob/main/ROADMAP.md) under *Next → FHIR Server Capabilities*.

aacruzgon:
**Scope note: the in-process gap is broader than `ws_manager.rs` alone**

While investigating this, the same in-process `DashMap` pattern shows up across `crates/subscriptions`, all constructed with empty in-memory maps in `SubscriptionEngine::new()` (`engine/mod.rs:80–88`) and held behind an `Arc` within a single process — with no shared/distributed backing and nothing persisted. Behind a load balancer this produces **four** related gaps, not one. Fixing only the WS client registry won't make WebSocket subscriptions cluster-correct, so suggesting we expand the scope accordingly.

| # | Component | Location | Multi-instance impact |
|---|---|---|---|
| 1 | WS client registry | `channels/ws_manager.rs` (`clients: DashMap<(tenant, sub_id), …>`) | The triggering resource write can land on any instance, but the WS client is connected to one specific instance. The writing instance's `ws_manager` has no entry for that client → no delivery. (The originally-reported issue.) |
| 2 | WS binding token store | `channels/ws_token.rs` (`tokens: DashMap<…>`, opaque UUID, consumed-on-first-use) | `$get-ws-binding-token` stores the token in instance A's map; the subsequent WS upgrade connect may land on instance B, whose map has no such token → `validate_and_consume` returns `None` → bind rejected. |
[...]

Issue #169, opened 2026-06-15
Database-backed SQL-on-FHIR export job state


smunini:
Replace the in-memory job controller (`InMemoryController` in `crates/rest/src/export/`) behind the async `$viewdefinition-export` / `$sqlquery-export` operations with database-backed job state, following the pattern already used by Bulk Data `$export` (whose job store shares the primary database).

**Problem**

Today, job status, tenant ownership, progress, and cancellation state live only in process memory:
- A server restart invalidates every in-flight and completed status URL, undermining the spec's 24-hour manifest validity.
- A second instance cannot see jobs submitted to the first.

⚠️ **Until this change is complete, deployments serving SQL-on-FHIR async operations must not be clustered** — run a single HFS instance for these endpoints, or pin them to one instance behind the load balancer. Synchronous `$viewdefinition-run` / `$sqlquery-run` and Bulk Data `$export` are unaffected.

**Scope**
- Move SoF export job state (status, tenant ownership, progress, cancellation) into database-backed storage
- Follow the existing Bulk Data `$export` job store pattern
- Preserve 24-hour manifest validity across restarts and across instances

Tracked on the [ROADMAP](../blob/main/ROADMAP.md) under *Next → FHIR Server Capabilities*.

PR #180, opened 2026-06-23
fix(sof-export): clean up partial results on cancel, plus a TTL reaper (#144)

### LINKED ISSUE #144: SoF export: clean up partial results when a job is cancelled
## Context

The SQL-on-FHIR spec (operations-common, as amended by [HL7/sql-on-fhir#365](https://github.com/HL7/sql-on-fhir/pull/365)) says servers SHOULD clean up any partial results when an export is cancelled via `DELETE` on the status URL.

Today, cancelling a `$viewdefinition-export` / `$sqlquery-export` job correctly transitions it to `Cancelled` (polls return 404, and as of the cancellation-race fix the background task can no longer resurrect it), but **already-written output shards are never deleted** and remain downloadable via `GET /export/{job_id}/{filename}` — `read_shard` checks tenant ownership only, not job status (`crates/rest/src/export/in_memory.rs`).

## Proposed change

- Add a `delete_job(&self, job_id: &str)` API to the `ExportSink` trait (`crates/rest/src/export/sink.rs`) and implement it for all three sinks:
  - `FilesystemSink` — `remove_dir_all` on the job directory
  - `InMemorySink` — drop matching keys
  - `S3Sink` — `ListObjectsV2` on the `{prefix}exports/{job_id}/` key prefix + delete objects
- Call it from the cancellation path (and from the background task when it finishes a job that was cancelled mid-run).
- Optionally also gate `read_shard` on job status so cancelled jobs' files 404 even before deletion completes.

## Severity

Spec SHOULD (not SHALL). Tenant isolation on the download route is already enforced, so this is a cleanup/storage-hygiene gap, not an access-control one.

mauripunzueta:
## Summary

Closes #144 and addresses the adjacent storage-hygiene gaps it surfaced in the SQL-on-FHIR async export controller (`$viewdefinition-export` / `$sqlquery-export`).

Previously, cancelling a job transitioned it to `Cancelled` but its already-written output shards were never deleted and stayed downloadable; and finished jobs were never reclaimed at all — the in-process `jobs` map and the sink output grew unbounded for the process lifetime, despite the completion manifest advertising a 24h `Expires`.

## What changed

**1. Cancellation cleanup (#144 as scoped)**
- New `delete_job(&job_id)` on the `ExportSink` trait, implemented for all three sinks:
  - `FilesystemSink` — `remove_dir_all` on the job dir (missing dir is not an error)
  - `InMemorySink` — drops entries under the `{job_id}/` key prefix
  - `S3Sink` — paginated `list_objects_v2` + `delete_object` under the job key prefix (idempotent)
- Called from the cancellation path, and again from the background task if it finishes a job that was cancelled mid-run (covers the write-after-cancel race).
- `read_shard` is gated on terminal state so a cancelled job's files 404 even while deletion is still draining.

**2. TTL reaper for finished jobs**
- A background reaper deletes a terminal job's output and drops its bookkeeping once it ages past the TTL, bounding both storage and the in-memory map. Aligned with the manifest's advertised 24h `Expires`.
- `JobStatus::Cancelled` now carries `cancelled_at`, and a `terminal_at()` helper lets all three terminal states age uniformly.
- New config: `HFS_EXPORT_OUTPUT_TTL` (default `86400`) and `HFS_EXPORT_CLEANUP_INTERVAL` (default `300`). The interval is clamped to ≥ 1s (`tokio::time::interval` panics on zero).

**3. Failed-job partial cleanup**
- A failed job's result URL returns 500 with no manifest, so its shards are unreachable but were never deleted. They're now removed on failure, and `read_shard` gates `Failed` alongside `Cancelled`.

## Deliberately *not* changed

`DELETE` on a **completed** job remains a no-op that preserves the `Completed` state (the status URL keeps redirecting to the result manifest), per `test_export_cancel_after_completion_preserves_completed_state`. Completed output is reclaimed by the TTL reaper above, not by `DELETE`.

## Testing

- New unit tests: `cancel_deletes_partial_output_and_download_404s`, `reap_expired_reclaims_terminal_jobs_only` (age threshold, running-immunity, output deletion + bookkeeping removal).
- Full `helios-rest` lib suite (282) and the `sof_export` integration suite (44) pass, including the completion-preservation test.
- `cargo fmt` clean; `cargo clippy --features s3 --all-targets -D warnings` clean.


smunini:
@claude and @mauripunzueta how does this change work in a clustered multi hfs environment?

smunini:
@aacruzgon and @mauripunzueta , this is another case where we need a cluster-capable job store.  Relates to https://github.com/HeliosSoftware/hfs/issues/169 and https://github.com/HeliosSoftware/hfs/issues/170 and https://github.com/HeliosSoftware/hfs/pull/150

We need a discussion document that describes the areas of hfs that need to be modified, via an environment variable config setting to use a unified cluster-capable job store and document it's use for those who want to run hfs in a cluster.

smunini (approved):

smunini:
https://github.com/HeliosSoftware/hfs/discussions/223

PR #188, opened 2026-07-02
feat(ui): add htmx ui foundation

angela-helios:
## Summary

Adds a minimal HTMX-based UI foundation for HFS.

This introduces a new optional `helios-ui` crate under `crates/ui`, mounted by the `hfs` binary when the `ui` feature is enabled. The UI currently exposes a small proof-of-concept page at `/ui` that displays the running HFS version, plus an HTMX-powered `/ui/version` partial refresh endpoint.

The `headless` feature path is also included so HFS can be built without compiling or mounting the UI.

## Changes

- Add new `helios-ui` crate for optional web UI code
- Add `ui` default feature to `helios-hfs`
- Add `headless` feature flag for no-UI builds
- Mount `/ui` and `/ui/version` from the HFS binary when UI is enabled
- Add minimal HTMX interaction for refreshing the displayed HFS version
- Keep FHIR routes available as fallback behind the UI routes

## Testing

- `cargo fmt --all`
- `cargo test -p helios-ui`
- `cargo check -p helios-hfs`
- `cargo check -p helios-hfs --no-default-features --features R4,sqlite,headless`


smunini (changes requested):
@angela-helios I would prefer to name the crate 'ui' on disk instead of 'web' and have the crate name be 'helios-ui' instead of 'helios-web'  I think web is too much of a generic term.

doug-helios:
NIT should we update CLAUDE.md? @angela-helios  @smunini  @mauripunzueta

crates/web/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+homepage = "https://github.com/HeliosSoftware/hfs/tree/main/crates/web"
+keywords = ["helios-software", "fhir", "ui", "htmx"]
+
+[dependencies]
+# Web framework (matches the version used by helios-hfs / helios-rest)
+axum = "0.8"
+
+# Compile-time, type-checked, auto-escaping templates (Jinja2-like).
+# Markup lives in templates/, never in Rust source.
+askama = "0.12"
```
doug-helios: I believe this is outdated as there is 0.16.0 already available

doug-helios (approved):

Issue #206, opened 2026-07-06
auth: HFS_AUTH_AUDIENCE is not enforced when the token lacks an aud claim


angela-helios:
**Found while verifying #198 end to end against the committed local Keycloak (docker/keycloak/).**

With `HFS_AUTH_AUDIENCE=hfs-api` configured, a token that carries **no `aud` claim at all** is accepted:

```
# Keycloak client_credentials token (the committed realm has no audience mapper -> no aud claim)
GET /Patient (Bearer <token without aud>) -> 200   # expected: 401 Invalid audience
```

Root cause: `crates/auth/src/provider/jwks_bearer.rs` calls `validation.set_audience(&[aud])`, but `jsonwebtoken`'s default `required_spec_claims` is `{"exp"}` — audience (and issuer) validation only runs **when the claim is present**. A token missing `aud` bypasses the check entirely, which defeats the purpose of audience restriction (cross-service token reuse is exactly the attack `aud` exists to stop, and an attacker-friendly IdP config is one with no audience mapper).

Fix sketch: when `expected_audience` is configured, add `"aud"` to `validation.required_spec_claims` (and similarly `"iss"` when `expected_issuer` is set, for defense in depth — HFS already refuses to boot without an issuer, but the required-claim flag is what makes a missing `iss` fail).

Also worth doing alongside: the committed realm (`docker/keycloak/realm.json`) should gain a hardcoded-audience protocol mapper so the local setup can exercise audience validation at all — tracked in the #198 write-up.

angela-helios:
Fix in #229. `set_audience`/`set_issuer` only validate their claim when present (jsonwebtoken’s `required_spec_claims` defaults to `{"exp"}`), so a token omitting `aud` slipped past a configured `HFS_AUTH_AUDIENCE`. The PR adds `aud` (and `iss`) to `required_spec_claims` when configured, so a missing claim now fails validation — extracted into a testable `build_validation()` helper with unit tests. Verified against the local Keycloak in the #198 pass.

Issue #205, opened 2026-07-06
auth: JTI replay cache rejects legitimate reuse of bearer access tokens (default config unusable with Keycloak)


angela-helios:
**Found while verifying #198 end to end against the committed local Keycloak (docker/keycloak/).**

With the default configuration (`HFS_AUTH_JTI_BACKEND=memory`), the second request presenting the **same access token** is rejected:

```
GET /Patient  (Bearer <token>)  -> 200
GET /Patient  (same token)      -> 401 {"issue":[{"code":"login","details":{"text":"JTI replay detected: aa0efc3b-..."}}]}
```

Keycloak (like every IdP) issues access tokens with a `jti` claim and a 300s lifetime, and OAuth2 clients reuse the token for its whole lifetime by design. `JwksBearerAuthProvider::authenticate` runs `jti_cache.check_and_store` on **every bearer validation** (`crates/auth/src/provider/jwks_bearer.rs`, step 8), so any real-world client gets one successful request per token. HFS is effectively unusable against Keycloak out of the box; the `redis` backend has the same behavior (plus persistence across restarts, verified).

**Why this is a design gap, not a broken cache:** single-use `jti` semantics belong to the *client assertion* JWT of the `private_key_jwt` flow (RFC 7523 §3), and that check is performed by the token issuer (the IdP), not the resource server. A resource server treating bearer *access tokens* as single-use breaks the OAuth2 contract.

**Suggested direction** (pick in review):
- Stop JTI-checking bearer access tokens altogether (the cache stays for any future flow where HFS itself consumes single-use assertion JWTs), or
- Make the default `HFS_AUTH_JTI_BACKEND=disabled` and document `memory`/`redis` as opt-in for deployments that deliberately issue single-use tokens.

**Workaround verified:** `HFS_AUTH_JTI_BACKEND=disabled` restores normal token reuse (200/200/200).

Full verification write-up: see the #198 deliverable.

angela-helios:
[...]

PR #207, opened 2026-07-06
docs(auth): verify open/SMART auth end to end + local Keycloak (#198)

### LINKED ISSUE #198: Auth: verify open/SMART auth end to end, configure local Keycloak, and scope out the UI login screen + IdP theming
Verify our open/standards-based authentication end to end against a local Keycloak, and scope out the browser login experience for the web UI.

## 1. Ensure open authentication works correctly

HFS does **not** act as an authorization server — `helios-auth` performs local SMART-on-FHIR / OAuth2 token validation only (JWT/JWKS), and token issuance stays with an external IdP (see `crates/auth/README.md`).

- [ ] Confirm SMART Backend Services (`private_key_jwt`) flow validates correctly end to end
- [ ] Confirm JWKS fetch/refresh, `HFS_AUTH_ISSUER` / `HFS_AUTH_AUDIENCE` / `HFS_AUTH_ALGORITHMS` validation, and SMART v2 scope enforcement behave as intended
- [ ] Confirm the `/.well-known/smart-configuration` discovery document (`crates/auth/src/discovery.rs`) advertises the right endpoints
- [ ] Confirm the JTI replay cache works for both backends (`HFS_AUTH_JTI_BACKEND`, `HFS_AUTH_REDIS_URL`)

## 2. Configure Keycloak locally

A pre-configured Keycloak 26 (`fhir` realm, test clients, SMART scopes) already exists under `docker/keycloak/`.

- [ ] Bring it up: `docker compose -f docker/keycloak/docker-compose.yml up -d` (port 8180)
- [ ] Run HFS against it with `HFS_AUTH_JWKS_URL` / `HFS_AUTH_ISSUER` / `HFS_SMART_TOKEN_ENDPOINT` pointed at the realm
- [ ] Exercise a real request with a token from `docker/keycloak/get-token.sh` (full-access + readonly clients) and confirm accept/deny behavior
- [ ] Document any gaps between the committed realm config and what the current code expects

## 3. Investigate the login screen (browser / interactive flow)

The web UI (`crates/ui`) will eventually need an interactive user login (authorization_code + PKCE), distinct from the backend-services flow above. Since HFS is not the auth server, the **login screen is owned by the IdP** — the browser redirects to Keycloak (or Okta/Auth0/Entra) to authenticate.

- [ ] Document the intended browser flow for the UI (redirect to IdP authorize endpoint, callback, session handling)
- [ ] Determine what HFS needs to add to support it (authorize/callback routes, session/cookie, discovery wiring)

## 4. Login-screen customization options

- [ ] Investigate customizing the **Keycloak** login screen (Keycloak themes / branding — logo, colors, layout) and whether we ship a Helios theme
- [ ] Summarize customization options for **other IdPs** (Okta, Auth0, Entra ID) and what is/isn't portable across them
- [ ] Recommend how far HFS should go: IdP-native theming vs. any HFS-side involvement

## Deliverable

A short write-up covering: auth verified working (2 flows), local Keycloak steps, the UI login-flow design, and a login-screen customization matrix across Keycloak + other IdPs.

angela-helios:
Deliverable for #198. Everything in the verification section was **executed against a running stack** (HFS + Keycloak 26.1 from `docker/keycloak/`), not read off the code.

## `docs/auth-verification.md` covers

1. **Auth verified end to end** — a results table (14 checks): SMART discovery served + auth-exempt, `401` with no token, SMART v2 scope enforcement (full-access writes; read-only client gets `403` on create and on an out-of-scope type), signature/tampering, algorithm allow-list, issuer, JWKS fetch, JTI (all three backends), and audience (positive + negative).
2. **Local Keycloak runbook** — bring-up, token retrieval, running HFS against the realm, accept/deny.
3. **UI login flow** — the intended browser **Authorization Code + PKCE** flow (sequence diagram) and exactly what HFS must add (authorize/callback routes, server-side state + PKCE verifier, session cookie, discovery wiring, a public realm client). Design for a follow-up, not built here.
4. **Login-screen theming matrix** — Keycloak / Okta / Auth0 / Entra: what each can customize and how portable it is, with a recommendation (do IdP-native theming; ship a Helios **Keycloak** theme; document branding knobs for the others; keep HFS out of the login-screen business).

## Defects found while verifying (filed separately)

- **#205** — the JTI replay cache rejects legitimate reuse of a bearer access token: the 2nd request with the same still-valid token is `401`ed. Single-use `jti` belongs to `private_key_jwt` client assertions (IdP-side), not resource-server access tokens. With the default `memory` backend HFS is effectively unusable against Keycloak until `HFS_AUTH_JTI_BACKEND=disabled`. `redis` behaves the same (persistence across restarts verified).
- **#206** — `HFS_AUTH_AUDIENCE` is bypassed by a token with no `aud` claim (`jsonwebtoken` only validates `aud` when present).

Neither is fixed here — this pass verifies and reports.

## Low-risk fixes included (so the runbook works as written)

- `docker/keycloak/realm.json`: hardcoded-audience mapper (`aud: hfs-api`) on both service clients, so audience validation is exercisable locally — verified matching `aud` → `200`, wrong expected `aud` → `401`. (Makes #206 observable; doesn't fix the missing-claim bypass.)
- `docker/keycloak/get-token.sh`: parse the token JSON with whichever of `jq`/`node`/`python3` actually works (a Windows `python3` stub alias resolves but fails), instead of `python3` only.

## Not covered (called out in the doc)

- `private_key_jwt` with a real asymmetric client assertion — the committed realm's clients use `client-secret`; HFS's validation role is identical either way, but the advertised capability isn't exercised until the realm gains such a client.
- Discovery advertises `private_key_jwt` + RS384/ES384 while the realm uses `client-secret` + RS256 — a config-fidelity gap (documented, not filed).

Closes #198.


angela-helios:
### Helios login theme (part 4 of #198 — login-screen customization)

Answering the customization question concretely: **yes, and here it is.** Since HFS is not the auth server, the login screen is IdP-owned — so this brands **Keycloak** (the IdP we bundle) with a CSS-only theme: Helios logo, Figtree, accent `#33b8ff`, light surface, all from the Dashboard V1.1 tokens. It extends the stock `keycloak` theme with **no template forks**, so it survives Keycloak upgrades.

Wired up: `docker/keycloak/themes/helios/` mounted into the container, selected via the realm `loginTheme`, and rendered end to end through a new public Authorization Code + PKCE client (`hfs-web`) + `demo`/`demo` test user — the scaffolding the §3 browser-login build will plug into.

Rendered and verified via headless Chrome against the running Keycloak. The password show/hide input-group needed extra styling for Keycloak 26 — a defect caught only because the page was actually rendered, not curl'd.
helios-login-screenshot

smunini (approved):

PR #216, opened 2026-07-07
feat(ui): wire the "FHIR resources over time" chart to real counts (#185)

### LINKED ISSUE #185: Expose Prometheus resource-count metrics to power the "FHIR Resources over time" UI chart
## Summary

The UI needs to render a **"FHIR Resources over time"** chart (see mockup below): a time-series/area chart of cumulative resource counts, broken down by resource type (Patient, Observation, Encounter, Condition, MedicationRequest, DiagnosticReport, Procedure, AllergyIntolerance, …), with a per-type series selector and current totals in the legend.

To serve this, we need to expose resource-count data that can be aggregated **over time and by resource type**. The natural home for this is Prometheus metrics scraped on an interval, which gives us the time dimension for free.

## Current state

There is currently **no Prometheus / metrics instrumentation anywhere in the workspace**:

- No `prometheus` / `metrics` / `opentelemetry` dependency in any `Cargo.toml` (0 occurrences in `Cargo.lock`).
- No `/metrics` endpoint. The only observability surface today is health/probe endpoints in `helios-rest` (`/health`, `/_liveness`, `/_readiness` in `crates/rest/src/handlers/health.rs`).
- Resource counting exists only as on-demand query logic, never recorded as a metric:
  - `count(...)` on the storage trait — `crates/persistence/src/core/storage.rs:368`
  - `count_resources(tenant, resource_type)` and `list_resource_types(...)` — `crates/persistence/src/search/reindex.rs:109` (used only for reindex progress)
  - FHIR `_summary=count` — `crates/rest/src/handlers/search.rs:519`
- `ROADMAP.md:236` currently marks "No Prometheus/OpenTelemetry integration … **Not planned**" — this issue would supersede that line, so we should update the roadmap as part of the work.

> Note: despite the phrasing "extend our Prometheus metrics," there are no existing metrics to extend — this is foundational work to stand up the metrics facility, with the resource-count metric as its first consumer.

## Proposed work

1. **Add a Prometheus metrics facility** to the server:
   - Add a metrics registry (e.g. `prometheus` or `metrics` + `metrics-exporter-prometheus`).
   - Expose a `GET /metrics` endpoint in `helios-rest` (text exposition format), wired into the `hfs` binary. Decide auth posture (likely auth-exempt like `/health`, or gated — see open questions).

2. **Emit a resource-count metric by type (and tenant):**
   - A gauge such as `hfs_fhir_resources_total{resource_type="Observation", tenant="..."}` reflecting the current count per resource type per tenant.
   - Reuse the existing `count_resources` / `list_resource_types` persistence logic to populate it, refreshed on a scrape/interval basis (avoid expensive per-scrape full counts on large stores — consider a cached/periodically-refreshed value).

3. **Serve the chart's time series:**
   - With Prometheus scraping the gauge over time, the "resources over time" chart is a range query against Prometheus.
   - Confirm whether the UI queries Prometheus directly or whether HFS should expose a small aggregation/query endpoint for the UI to consume (see open questions).

## Chart reference (UI mockup)

Cumulative area chart, x-axis = date, y-axis = resource count (0–50k in the mock), one series per selected resource type, multi-select of resource types, legend shows current totals (e.g. Patient 1,204 · Observation 38,910 · Encounter 9,3xx).

## Open questions

- Should `/metrics` be auth-exempt (like `/health`) or protected? Multi-tenant metrics may leak tenant existence/volume.
- Per-tenant label cardinality — acceptable, or aggregate at the exposition layer?
- Does the UI scrape Prometheus directly (needs a Prometheus deployment + query proxy), or should HFS provide a purpose-built time-series endpoint for the chart? This affects whether we also need historical storage inside HFS vs. relying on Prometheus retention.

mauripunzueta:
Closes #185.

## Context

The observability work merged in #150 stood up the Prometheus `/metrics` facility and the authenticated console `resource-counts` endpoint — but the dashboard's **"FHIR resources over time"** chart still rendered a hardcoded SVG with sample data (the code comments called this out as follow-up work). This PR wires that chart to real per-type cumulative counts, closing the remaining gap in #185.

## Design decisions (resolving #185's open questions)

- **Server-side rendered, default tenant.** The `/ui` dashboard is an operator view mounted outside the auth layer, so the chart renders the server's **default tenant** counts server-side (no browser token needed, no client JS).
- **No tenant leak.** Per-tenant counts are still *never* exported to the public `/metrics` endpoint. This reuses the same storage count path as the authenticated console JSON endpoint — a JSON/render path, not a Prometheus label — consistent with the posture established in #150.
- **UI stays storage-agnostic.** Rather than making `helios-ui` depend on the persistence layer, the server registers a provider with a light process-global registry in `helios-observability` (mirroring the existing `uptime`/`metrics` globals); the UI reads a plain snapshot.

## Changes

- **`helios-observability`** — new `dashboard` module: a process-global `DashboardProvider` registry + plain `DashboardSnapshot` data types (no persistence/FHIR types, keeps the crate light).
- **`helios-rest`** — `StorageDashboardProvider` (default tenant, 8 dashboard types, 30-day window) implementing that trait, registered in `build_app`. Extracted a shared `resource_count_series` helper so the console `resource-counts` handler and the dashboard provider share **one** cumulative-bucketing implementation.
- **`helios-ui`** — depends on `helios-observability`; renders the real chart server-side (computed SVG polyline, y-axis ticks, x-axis dates) plus a legend of per-type totals that doubles as a `?type=` series selector (progressive enhancement, works without JS). Falls back to placeholder data when no provider is registered. Drops the now-unused `chart-unit-patients` i18n key across en/es/de.
- **`ROADMAP.md`** — flips "No performance metrics … Not planned" to implemented.

## Verification

Opening this PR primarily to run CI: the authoring environment has no C linker, so `cargo check`/`test`/`clippy` could not be run locally. `cargo tree` confirms the dependency graph resolves. New unit tests cover the dashboard projection, chart geometry, formatting helpers, query parsing, and the empty/unknown-type fallbacks; the console `resource-counts` JSON shape is unchanged (existing integration tests still apply).


crates/observability/src/dashboard.rs
```diff
@@ -0,0 +1,133 @@
...
+//! surface.
+
+use std::sync::{Arc, RwLock};
+
+use async_trait::async_trait;
+
+/// One daily point of a single resource type's cumulative growth curve.
+#[derive(Clone, Debug)]
+pub struct DashboardPoint {
+    /// `YYYY-MM-DD` (UTC) label for the bucket.
```
smunini: Is day the most specific precision that we can support? I'd like the chart to have better precision - to mins or seconds.

smunini: Good question — day was never a storage limit, so yes. Done in 71b7194: the chart now supports minute resolution, but getting there needed more than changing the `GROUP BY`.

**Precision wasn't the constraint.** Every backend already stores `last_updated` at millisecond precision (RFC3339 text in SQLite, `TIMESTAMPTZ` in Postgres, BSON `DateTime` in Mongo). Day bucketing was just a choice in the query.

**But finer buckets alone would have made the chart worse.** Two reasons:

1. *The series wasn't really a time series.* It bucketed the live `resources` rows by the `last_updated` of each resource's **current version**, so every resource sat in the bucket of its most recent *edit*. Update a Patient created three weeks ago and it silently leaves that day's bucket and joins today's — the past changes shape under you. At day resolution over 30 days that drift is easy to miss; at minute resolution it becomes the dominant feature, and you'd be looking at edit recency, not growth.

2. *Point cardinality explodes.* Span and bucket were coupled. Keep the 30-day window and switch to minutes and you get 43,200 points **per resource type** — ~345K across the 8 charted types, serialized into an SVG polyline on every page load. Seconds would be 2.6M per series.

**What I changed:**

- **`ResourceStorage::count_deltas_by_bucket`** (persistence) — aggregates the immutable `resource_history` log into fixed-width, epoch-aligned buckets: creation `+1`, delete `-1`, plain update `0`. Write events never move once written, so the curve is stable under later edits and stays honest as buckets get finer. Implemented on SQLite, Postgres and MongoDB; the composite store delegates to its primary.
- The curve still **ends exactly on the live total**: deltas are summed forward from a `count() - net-in-window` baseline, so anything history can't reconstruct (a resurrecting `PUT` after a delete, or rows predating the log) is absorbed into the baseline rather than skewing the endpoint.
- **`DashboardWindow`** (observability) — I deliberately did *not* expose bucket width as a free knob, since that invites the blow-up above. Instead each window pairs a span with the bucket that samples it: **1h/1min, 24h/30min, 30d/1day**, each landing in the 30–60 point range.
- **UI** — a window selector beside the chart, as plain links (works without JavaScript). Intraday windows label the x-axis with UTC clock times; 30d keeps calendar dates.

Verified end-to-end against a live sqlite-backed server (writes land in the correct minute bucket on the 1h view), with Postgres and Mongo covered by integration tests against real engines — worth it, since Postgres rejected the query until the bucket-width param was cast `::bigint` (it infers `numeric` from `EXTRACT(EPOCH ...)`), which no unit test would have caught.

Two things worth knowing:

- The console `resource-counts` JSON keeps its calendar-day `date`/`count`/`cumulative` shape, but `count` is now a **net** delta and so can be negative where it previously couldn't.
- `count_by_day` is still in place (the console endpoint uses it via the shared helper's day-width path). It's arguably redundant now — happy to retire it in a follow-up, but I didn't want to widen this PR.

If you want to go finer still — second buckets — the mechanism now supports it; it's just a new `DashboardWindow` variant with a span short enough to keep the point count sane (e.g. 5min/1sec = 300 points).

smunini (approved):

PR #220, opened 2026-07-08
feat(tenants): first-class tenant maintenance (registry, admin API, UI)

### LINKED ISSUE #213: Tenant maintenance API: list, add, and delete tenants
## Summary

We need to ensure HFS has sufficient **API capability to support tenant maintenance**. Today, tenants are not first-class managed entities — a "tenant" is simply an identifier string that is resolved per request (via `X-Tenant-ID` header, URL prefix, or JWT claim) and used to scope data. **Any well-formed tenant ID is implicitly accepted the first time it is used.** There is no tenant registry, no tenant table, and no CRUD API.

This issue tracks adding first-class tenant-maintenance APIs.

## Required capabilities

- [ ] **List tenants** — return the set of known/provisioned tenants (with metadata: id, display name, status, created date, resource counts where feasible).
- [ ] **Add a tenant** — provision a new tenant explicitly (register it, and where applicable run the schema/database provisioning path).
- [ ] **Delete a tenant** — deprovision/remove a tenant and (per config) tear down its data.

## Current state (findings)

- **No REST endpoints** exist for list/add/delete tenants. Handlers live in `crates/rest/src/handlers/` and routes in `crates/rest/src/routing/fhir_routes.rs`; there is no `tenant`/`admin`/`tenants` handler or route.
- **No tenant registry** — tenants are never enumerated or stored. Config only provides `HFS_DEFAULT_TENANT` plus routing-mode vars (`HFS_TENANT_ROUTING_MODE`, `HFS_TENANT_STRICT_VALIDATION`, `HFS_JWT_TENANT_CLAIM`) in `crates/rest/src/config.rs`.
- **Reusable building blocks already present:**
  - `TenantId`, `TenantContext`, `TenantPermissions` in `crates/persistence/src/tenant/`.
  - Tenant resolver/middleware/extractor in `crates/rest/src/tenant/` and `crates/rest/src/middleware/`.
  - **Unused** SQL provisioning helpers in `crates/persistence/src/strategy/{schema_per_tenant.rs,database_per_tenant.rs}`: `create_schema_sql`, `drop_schema_sql`, `list_tenant_schemas_sql`, `provision_tenant_sql`, `deprovision_tenant_sql`, `list_tenant_databases_sql` (plus `auto_create_schema` / `drop_on_delete` flags). None are currently called anywhere — good starting point for wiring add/delete.

## Design considerations

- **Tenant registry**: decide where the canonical list of tenants lives (new persistence table vs. per-strategy enumeration like `list_tenant_schemas_sql`).
- **AuthN/AuthZ**: these are administrative operations — must be gated by an appropriate admin scope/permission (see `helios-auth`), not open to normal FHIR clients.
- **Delete safety**: deleting a tenant is destructive. Needs confirmation semantics and a config-gated data teardown (`drop_on_delete`), plus audit logging (`helios-audit`).
- **Routing modes**: behavior should be consistent across `header_only` / `url_path` / `both` routing modes.
- **API shape**: decide between a FHIR-style operation/resource vs. an admin REST surface (e.g. `/admin/tenants`).

## Acceptance criteria

- An authenticated admin can list, create, and delete tenants via API.
- Creating a tenant provisions storage as configured; deleting respects the teardown policy.
- Operations are permission-gated and audit-logged.
- Tests cover list/add/delete across supported routing modes and storage backends.

A UI mockup for the tenant maintenance page will be linked in a follow-up comment.

angela-helios:
Makes tenants **first-class managed entities** — list / add / delete across persistence, the REST API, and the web UI. Until now a "tenant" was only an implicit identifier string resolved per request; there was no registry, no metadata, and no lifecycle API. Implements #213.

Built to the **reviewed scope** (the mockup feedback): the page keeps search, resources, created, and delete; it drops the status / provisioning / storage-backend cards and the routing column (all server-global or non-existent, not per-tenant), and distinguishes the **id** (used in the API) from an optional human-friendly **display name**.

## Persistence (`helios-persistence`)
- `ResourceStorage` gains a tenant-registry surface, **default-off** so other backends are unaffected: `supports_tenant_registry`, `list_tenants`, `get_tenant`, `register_tenant`, `deregister_tenant`, `purge_tenant_data`, plus the `TenantRecord` type (`id`, optional `display_name`, `created_at`).
- SQLite: schema **v14** adds a `tenants` table; PostgreSQL (schema v14), MongoDB (schema v6, unique id index), and S3 (`tenants/<id>.json`, conditional-put register) implement the same surface, with unit/integration tests per backend.

## REST API (`helios-rest`)
- `GET`/`POST`/`DELETE /admin/tenants`. **List** merges the registry with live per-tenant counts (`count_by_tenant`), so tenants that only have data still appear (`registered:false`, `created_at:null`). **Add** validates the id and `409`s on duplicates. **Delete** is deregister-only unless `?purge=true` also tears down data (destructive, opt-in).
- Mounted in the console **admin** tier — system-context scope + authentication — and **audit-logged**. Returns `501` when the backend has no registry.
- 8 integration tests (list/add/delete, data-only discovery, purge, 404, invalid ids, conflict).

## Web UI (`helios-ui`)
- Server-rendered, htmx-driven `/ui/tenants` in the Dashboard V1.1 style: two stat cards, search, an add slide-over, and a per-row delete with confirm. Every mutation returns the refreshed table fragment (works with or without the swap). New nav item; i18n keys added across en/es/de (parity test green).
- `mount()` now takes the storage handle; `hfs` shares the SQLite backend Arc via the new `create_app_with_auth_arc`.

## Verified end to end
Against a running `hfs` (default SQLite): create tenants via the API and the UI form, data-only tenants discovered with counts, search filtering, delete/deregister, and the JSON API reflecting UI mutations. Screenshot below.

## Scope notes / follow-ups
- Registry + purge now cover **every backend mode**: SQLite, PostgreSQL, MongoDB, and S3 primaries; `*-elasticsearch` composites delegate the registry to the primary and fan the purge out so offloaded search documents are cleared too. S3 returns `501` only in bucket-per-tenant mode without a default system bucket (there is no cross-tenant place to keep the records).
- Every `hfs` serve path threads its storage into the web UI (composite modes hand the composite, keeping a UI purge search-aware).
- Auto-registering a tenant on first write (so `created_at` is captured for organically-created tenants) stays out of scope (tracked in #240) — the hot write path is untouched.

Closes #213.
Screenshot 2026-07-08 105134
Screenshot 2026-07-08 105147


smunini (changes requested):
Only sqlite backend is currently supported.  Ensure that this will work with all backends.

angela-helios:
@smunini done in 712ff550 — the registry now works on **all backends**, matching the issue's acceptance criteria:

- **PostgreSQL**: schema v14 adds the `tenants` table (`created_at` stored as RFC 3339 text so records read back byte-identical across backends); purge runs in one transaction over `search_index`, `resource_fts`, `resource_history`, and `resources`.
- **MongoDB**: schema v6 adds a `tenants` collection with a unique `id` index, so a double-register race surfaces as a duplicate-key error.
- **S3**: one JSON record per tenant at `[prefix/]tenants/<id>.json`, registered with `If-None-Match: *` for uniqueness; purge sweeps the tenant's `resources/` and `history/` prefixes. The only remaining `501` is bucket-per-tenant mode without a default system bucket — there is no cross-tenant location to keep the records in that topology.
- **Elasticsearch** (search secondary, never a primary): implements the tenant data purge via `delete_by_query` with an exact `tenant_id` term — the index wildcard alone would also sweep prefix-colliding tenants (`acme` vs `acme_corp`), and there's a regression test for exactly that.
- **Composite** (`*-elasticsearch` modes): registry reads/writes delegate to the primary; purge fans out to secondaries so offloaded search documents are cleared too, and a failed secondary purge is surfaced instead of swallowed.
- **hfs binary**: every serve path now threads its storage into the web UI (previously SQLite only); composite modes hand the composite so a UI purge stays search-aware.

Test coverage: registry CRUD + purge integration tests for Postgres (testcontainers) and MongoDB (replica set), S3 mock tests including the bucket-per-tenant edge cases, composite delegation/fan-out unit tests, and the ES exact-term purge test. Postgres and Mongo verified green locally against real servers; the ES one runs in CI (the ES container won't start on my machine — known Docker Desktop cgroup issue that hits the pre-existing ES tests too).

One thing intentionally left out of this PR: auto-registering organically-created tenants on their first write (so `created_at` gets captured without an explicit `POST /admin/tenants`). That touches the hot write path and needs a seen-tenants cache, so it's tracked separately in #240.

smunini (approved):

PR #221, opened 2026-07-08
docs(mcp): design + requirements ADR for the FHIR MCP server

angela-helios:
Phase-1 deliverable for the FHIR MCP Server issue: **research the landscape and write the design/requirements plan** before building the crate (per the issue's "Design spike + ADR" phase and the ask to study what the leading FHIR-MCP servers do first).

`docs/mcp-server-design.md` covers:

- **Landscape research** (July 2026) — a comparison table of the leading FHIR-MCP servers: Aidbox/Health Samurai (alpha, streamable HTTP+SSE, CRUDS+validate, AccessPolicy auth), xSoVx/fhir-mcp (the "security-hardened" reference — PHI engine, AuditEvent, SMART+break-glass, token efficiency), The Momentum, Medplum, flexpa — and **what we adopt** from each.
- **Key decisions (ADR):**
  - **Use the official Rust SDK `rmcp`** (v0.12, edition 2024, Apache-2.0, `#[tool]` macro, both transports, OAuth) rather than hand-rolling JSON-RPC — with a spike action item to confirm ergonomics against Claude.
  - **stdio first, streamable HTTP next** (skip deprecated HTTP+SSE).
  - **Share a version-agnostic service layer beneath the REST handlers** so MCP and REST don't drift — phased (interim `mcp::service` module for the MVP, full extraction as a follow-up) so it doesn't block.
  - Auth (stdio = local/config principal; HTTP = `helios-auth` bearer + SMART scopes → tool availability), safety (reads on / writes off behind `HFS_MCP_ALLOW_WRITE` + scope, every call `helios-audit`-logged, delete needs `confirm`), and **result-shaping for token budgets** (summary projections, paging, resource URIs instead of inlining huge bundles).
- **Full tool/resource/prompt surface**, `HFS_MCP_*` config, the `crates/mcp/` crate/binary shape, a 6-phase build order, a testing strategy, and **open questions to settle at the spike**.

This is a document for review — no code yet. Once the SDK choice and the shared-service boundary are ratified, the read-only stdio MVP (phase 2) is the next PR.

Relates to the FHIR MCP Server issue.

### DIFF
--- a/docs/mcp-server-design.md
+++ b/docs/mcp-server-design.md
@@ -1,0 +1,368 @@
+# HL7 FHIR MCP Server — design & requirements (ADR)
+
+**Status:** Design spike / requirements — for review before building
+**Scope:** the umbrella design for a new `helios-mcp` crate + `mcp` binary that
+exposes HFS's FHIR capabilities as Model Context Protocol tools, resources, and
+prompts.
+**Owner:** Angela
+**Date:** 2026-07-08
+
+This is the "phase 1" deliverable the issue calls for: study what's out there,
+understand what the leading FHIR-MCP servers do, and write a plan — the design
+decisions and requirements — *before* writing the crate. It ends with a
+recommended build order and the open questions to settle first.
+
+---
+
+## 1. Goal
+
+Make the Helios FHIR Server directly usable by LLM agents (Claude Desktop/Code,
+IDE assistants, other MCP clients) without those agents hand-rolling FHIR REST
+calls, FHIRPath, or SQL-on-FHIR. The MCP server is a **thin bridge** from MCP's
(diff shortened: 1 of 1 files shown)


smunini (changes requested):
This is a good start, and we should move this from a PR to a Discussion Document.  Without naming names of other solutions out there - we should come up with our own opinion on what would make a really great MCP server for the FHIR community. Let's turn this PR into a Discussion document and invite the community to comment on it instead of a docs file here.

angela-helios:
Moved to a Discussion per review: **#231** (Ideas). Reworked it into our own opinionated take on what a great FHIR MCP server should be — dropped the named comparison of other solutions and the docs-file framing, and added open questions to invite community comment. Closing this PR in favor of the discussion; no docs file will land until the design is ratified there.

Issue #227, opened 2026-07-08
HTS Terminology Server perf regression on main since v0.2.0 (EX02/EX08 expansion collapse + ~30% uniform drop)


mauripunzueta:
## Summary

The HTS Terminology Server has a **performance regression on `main`** vs the v0.2.0 release (2026-06-17). Two distinct symptoms at 30 VUs:

1. **Two ValueSet-expansion tests collapse** to near-zero throughput / timeouts (high confidence this is a real regression).
2. **Broad ~65–80% throughput** on essentially every other test, on **both** SQLite and Postgres (medium confidence — needs bisect to separate real overhead from benchmark-definition drift).

This is **not** caused by PR #219 — see "Isolation" below. Found while running the HTS no-regression check for #219.

## Evidence (HTS Terminology Benchmark, vu50)

| Test | v0.2.0 (run 27693178932) | current `main` (run 28980123786) |
|---|---|---|
| **SQLite EX08** | 24,493 rps, 0% err | **68 rps** |
| **PG EX02** | 21,840 rps, 0% err | **6 rps, ~13% err, p95 = 30s (statement_timeout)** |
| SQLite LK05 | 36,470 | 25,574 (70%) |
| SQLite VC01 | 30,103 | 22,065 (73%) |
| PG LK05 | 37,595 | 23,109 (61%) |
| PG VC01 | 30,584 | 19,453 (64%) |

The EX08 (SQLite) / EX02 (PG) collapse from ~20K rps to single/double digits with 30s statement-timeouts is the clearest signal — that is not measurement noise.

## Isolation — this is a `main` regression, not from any single open PR

Confirmed with a branch-vs-main control, both using **identical** current benchmark definitions and comparable runner load:
- **`main`** (run 28980123786) and the **#219 branch** (run 28977396638) are **statistically identical** — SQLite 95–101% of each other, PG 102–107% — and **both** reproduce the EX02/EX08 collapse. So the regression predates/excludes #219 (whose diff touches only FHIR search/transaction/sqlite paths, no HTS terminology code).
- The gap is therefore something merged to `main` between v0.2.0 (2026-06-17) and now.

## Suspect commits (since v0.2.0)

[...]

PR #233, opened 2026-07-09
fix(hts): don't default to hierarchical expansion for paged is-a filters (#227)

mauripunzueta:
Fixes the ValueSet-`$expand` half of #227 — the EX02/EX08 collapse.

## Root cause

`6c1876c01` ("nest is-a/descendent-of filter expansions by default") made any compose whose include carries an `is-a` / `descendent-of` filter default to tree mode, so the IG `search/search-filter-yes` fixture would nest. The default also fired for **paged** expansions, and tree mode is not a shape a paged caller can use:

- every closure-backed `LIMIT`/`OFFSET` fast path is gated on `req.hierarchical != Some(true)` (`postgres/value_set.rs:357,783`, `sqlite/value_set.rs:591`), so nesting skips all of them;
- `pg_filter_is_a` then fetches the **full** descendant set — the code comment already says "~140k rows for a SNOMED root" (`postgres/value_set.rs:3088-3097`);
- `build_hierarchical_expansion` loads the **entire** `concept_hierarchy` edge table for the system (`SELECT parent_code, child_code FROM concept_hierarchy WHERE system_id = $1`, `postgres/value_set.rs:3370`, `sqlite/value_set.rs:5923`);
- the tree branch returns with `offset: None` and no truncation — `count`/`offset` are dropped (`postgres/value_set.rs:1004-1013`, `sqlite/value_set.rs:1144-1154`).

So `count=10` over a SNOMED is-a root read ~10^5 concepts plus ~10^6 edges **per request** and answered with the whole subtree. The benchmark's EX02 (`descendent-of 64572001`, count=10) and EX08 (`is-a 404684003` + property + text filter, count=10) are exactly that shape:

| Test | v0.2.0 | main | cause |
|---|---|---|---|
| SQLite EX08 | 24,493 rps | 68 rps | full subtree materialized per request |
| PG EX02 | 21,840 rps | 6 rps, p95 = 30s | same, hits `statement_timeout` |

The same block also put an **extra `ValueSet` lookup on every URL-resolved `$expand`** that omitted `hierarchical`/`excludeNested`, which is EX01/EX03/EX04's hot path.

## Fix

Suppress the *default* when the request pages (`count`/`offset`). Explicit `hierarchical=true` / `excludeNested=false` is untouched.

This is what the IG requires, not just what's fast. Scanning `HL7/fhir-tx-ecosystem-ig`'s `test-cases.json`:

- **Every** expand fixture pairing a subsumption filter with `count`/`offset` expects a **flat** list: `simple-cases/simple-expand-isa-c2`, `-o2`, `-o2c2`, and `snomed/expand-pc-none`.
- **No** fixture that expects nesting passes `count`/`offset` — they all pass `excludeNested`, or nothing.
- A text `filter` must **keep** nesting: `search/search-filter-yes` sends `filter=data` and expects `data-exchange1/2/3` nested under `data-exchange`. So `filter` is deliberately *not* part of the guard.

## Tests

This path had no Rust coverage at all — `search-filter-yes` was only exercised by the external IG conformance workflow. Adds three cases to `crates/hts/tests/value_set_ops.rs`:

- `expand_is_a_filter_nests_by_default` — unpaged is-a filter nests.
- `expand_is_a_filter_stays_flat_when_paged` — `count=2` stays flat and honours `count` (the #227 guard; also pins `simple-expand-isa-c2`).
- `expand_is_a_filter_with_text_filter_still_nests` — mirrors `search-filter-yes` locally.

## Validation

- HTS IG conformance + HTS benchmark dispatched on this branch; results posted below.
- Does **not** address the other half of #227 (the uniform ~65–80% throughput drop on both backends, suspected per-request observability middleware). That stays open on the issue.

### DIFF
--- a/crates/hts/src/operations/expand.rs
+++ b/crates/hts/src/operations/expand.rs
@@ -1620,7 +1620,18 @@ async fn process_expand_inner<B: TerminologyBackend>(
         // pure-full-system case this applies to URL-resolved ValueSets too
         // (the IG `search/search-filter-yes` fixture is URL-resolved). Inspect
         // the inline compose when present, else fetch the referenced VS.
-        if hierarchical.is_none() {
+        //
+        // Paging suppresses this default. Tree mode returns the whole subtree —
+        // the backends drop `count`/`offset` and report no `offset` — so nesting a
+        // paged request would silently ignore what the caller asked for, and
+        // answering `count=10` over a SNOMED is-a root would first read that
+        // root's ~10^5 descendants. The IG agrees: every fixture pairing a
+        // subsumption filter with count/offset (`simple-cases/simple-expand-isa-
+        // {c2,o2,o2c2}`, `snomed/expand-pc-none`) expects a flat list, and no
+        // fixture expecting nesting pages. A text `filter` does NOT suppress it —
+        // `search/search-filter-yes` filters and still expects a tree. An explicit
+        // `hierarchical=true` / `excludeNested=false` nests regardless.
+        if hierarchical.is_none() && count.is_none() && offset.is_none() {
             let ctx = TenantContext::system();
             let resolved_vs: Option<Value> = if value_set.is_some() {
                 value_set.clone()
--- a/crates/hts/tests/value_set_ops.rs
+++ b/crates/hts/tests/value_set_ops.rs
@@ -168,6 +168,140 @@ async fn expand_is_a_codesystem_flat_with_exclude_nested() {
     );
 }
 
+/// An include carrying an `is-a` filter designates a hierarchy subtree, so an
+/// unpaged expansion of it nests by default (IG `search/search-filter-yes`).
+#[cfg(feature = "sqlite")]
+#[tokio::test]
+async fn expand_is_a_filter_nests_by_default() {
+    let app = TestApp::new();
+    app.import_bundle_ok(ISA_HIERARCHY_BUNDLE).await;
+
+    let (status, body) = app
+        .post_fhir("/ValueSet/$expand", isa_filter_expand(&[]))
+        .await;
+    assert_eq!(status, StatusCode::OK, "{body}");
+
+    let top = body["expansion"]["contains"].as_array().unwrap();
+    let top_codes: Vec<&str> = top.iter().filter_map(|c| c["code"].as_str()).collect();
+    assert_eq!(
+        top_codes,
+        vec!["parent"],
+        "an is-a filter should nest its matches under the filter root, got {body}"
+    );
+
+    let children: Vec<&str> = top[0]["contains"]
+        .as_array()
+        .expect("parent should nest its children")
+        .iter()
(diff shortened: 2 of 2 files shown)


mauripunzueta:
## Validation results

All green, and the expansion collapse is fully recovered.

| Check | Result |
|---|---|
| HFS CI (`29021783522`) | ✅ — all 3 new tests executed and passed |
| HTS IG Conformance, both backends (`29021802731`) | ✅ |
| HTS Benchmark preflight, both backends (`29021813350`) | ✅ EX02, ✅ EX08 |
| HTS Benchmark full, both backends (`29023969774`) | ✅ 20/20 preflight, results below |

### Throughput vs. the main-control run (`28980123786`), vu50

**SQLite** — controls are flat (LK02 0.99×, CM01 1.00×, LK05 0.98×), so the noise floor is ±2%:

| Test | main | this PR | | p95 main → PR |
|---|---:|---:|---|---|
| **EX08** | 68 rps | **19,027 rps** | **280×** | 3,200 ms → 5 ms |
| **EX02** | 12,181 rps | **16,852 rps** | 1.38× | 8 ms → 5 ms |
| EX05 | *failed preflight* | **17,172 rps** | — | — |

**Postgres** — controls drifted up ~5–9% (slightly less contended run), so discount accordingly:

| Test | main | this PR | | p95 main → PR | err main → PR |
|---|---:|---:|---|---|---|
| **EX02** | 6 rps | **16,514 rps** | **2,750×** | 30,031 ms → 6 ms | 12.9% → 0.0% |
| **EX01** | 440 rps | **5,539 rps** | 12.6× | 466 ms → 15 ms | 0% → 0% |

### Notes on what the data does and doesn't show

- **SQLite EX05 failed preflight on main** and was excluded from the benchmark entirely. It passes here. Main wasn't just slow on that test — it was silently skipping it.
- **Postgres preflight on main only recorded 11 of 20 tests** (no baseline exists for CM01/CM02/EX03–EX08/SS01), so I'm deliberately *not* claiming a PG EX08 number. All 20 pass preflight on this branch.
- **PG EX01's 12.6× is consistent with dropping the per-request `ValueSet` lookup** this block added to every URL-resolved `$expand`. Note SQLite EX01 is flat (0.99×) — so that lookup's cost was Postgres-specific (round-trip + pool pressure), not universal.
- **The uniform ~30% drop is unchanged**, as expected: SQLite LK05 is 25,185 here vs 36,470 at v0.2.0 (69%), VC01 21,720 vs 30,103 (72%). This PR doesn't touch that, which is what makes the EX02/EX08 recovery cleanly attributable to the expansion fix. That half of #227 stays open — current evidence points at `middleware::track` (global `Mutex` in `reqlog::record` on every request, an always-enabled INFO span, ~9 label allocations) rather than at the `hts::probe` logs named in the issue, which predate the v0.2.0 baseline by five weeks and cannot be the regression.

smunini (approved):

Issue #199, opened 2026-07-06
Support per-user UI settings (SettingsStore) on the S3 backend


mauripunzueta:
## Summary

Add a per-user `SettingsStore` implementation to the **S3** storage backend so a `HFS_STORAGE_BACKEND=s3` deployment can serve the `/_user/settings` endpoints instead of returning `501 Not Implemented`.

## Context

PR #151 introduced an opaque, per-user JSON UI settings store (`SettingsStore` trait in `helios-persistence`, `GET`/`PUT`/`PATCH /_user/settings` in `helios-rest`). It is implemented for **SQLite**, **PostgreSQL**, and now **MongoDB** — every standalone transactional primary backend.

Per review feedback on #151 (@smunini), S3 is also offered as a standalone primary (`HFS_STORAGE_BACKEND=s3`, CRUD/versioning/history), so an S3-only operator currently hits an (explained) `501` on the settings endpoints. This issue tracks closing that gap. The `501` message now names the supported backends and explicitly excludes S3/Elasticsearch, so the behaviour is documented rather than silent in the meantime.

## Why deferred from #151

- S3's recommended role is **Archive**, not a hot transactional primary, so the settings use case (frequent small read-modify-write updates from a UI) is a weaker fit.
- The `SettingsStore` contract relies on a **read-modify-write with a monotonic `version`** for optimistic locking (surfaced as a weak `ETag`). On SQLite/Postgres this is a `SELECT … FOR UPDATE` transaction; on MongoDB a version-conditioned update + retry. S3 has no transactions — correctness requires **conditional writes** (`PutObject` with `If-Match` / `If-None-Match`), whose availability through the backend's object-store client/SDK needs to be verified before implementing.

## Work

- [ ] Confirm the S3 backend's object-store client exposes conditional `PutObject` (`If-Match` on the object's ETag, `If-None-Match: *` for create-if-absent).
- [ ] Store one object per user key (kept out of the FHIR resource key space, mirroring how the relational stores keep `user_settings` separate from `resources`).
- [ ] Implement `get_settings` / `put_settings` / `patch_settings` with optimistic concurrency mapped onto conditional writes; translate a failed precondition to `ConcurrencyError::OptimisticLockFailure`.
- [ ] Wire the store into the `start_s3` / `start_s3_elasticsearch` paths in `crates/hfs/src/main.rs` (via `create_app_with_auth_bulk_and_settings`).
[...]

PR #241, opened 2026-07-10
feat(ui): saved FHIR queries per resource type, sorted by last access

### LINKED ISSUE #234: User settings: maintain a list of saved FHIR queries by resource type, sortable by last access
## Summary

Expand our user settings capabilities so a user can maintain a list of saved FHIR queries, organized by resource type, sortable by last-access datetime.

## Where we are today

User settings already exist, but as an **opaque JSON document** — the server stores and versions it, and imposes no schema:

- Core model: `StoredUserSettings` + `SettingsStore` trait — `crates/persistence/src/core/user_settings.rs`
  (`user_key`, `document: serde_json::Value`, monotonic `version` surfaced as weak ETag `W/"{version}"`, `updated_at`)
- REST: `GET` / `PUT` / `PATCH` on `/_user/settings` — `crates/rest/src/handlers/user_settings.rs`, routed at `crates/rest/src/routing/fhir_routes.rs:236`
  - `PUT` replaces the whole document, `PATCH` is RFC 7386 JSON Merge Patch, both with `If-Match` optimistic locking
- Backends: SQLite, PostgreSQL, MongoDB. Elasticsearch and S3 deliberately return `501` (see the rationale comment at `crates/rest/src/handlers/user_settings.rs:118-139`)
- Identity: `UserKey` = `"{issuer}|{subject}"`, or `"local|default"` when auth is disabled — `crates/rest/src/extractors/user.rs`

A `recentQueries` key (shaped roughly `{"Patient": ["name=smith"]}`) appears in doc comments and tests, but it is **only an example** — there is no typed model, no last-access tracking, and no server-side behavior behind it. The only live consumer of user settings today is the UI theme toggle (`crates/ui/src/lib.rs`, issue #197).

## What we want

Users should be able to keep a durable, per-user list of FHIR queries, grouped by resource type, and see them ordered by when they were last used.

Proposed document shape under a `savedQueries` key:

```json
{
  "savedQueries": {
    "Patient": [
      {
        "id": "01J...",
        "name": "Smiths in Boston",
        "query": "name=smith&address-city=Boston",
        "createdAt": "2026-07-01T12:00:00Z",
        "lastAccessedAt": "2026-07-09T09:14:22Z",
        "accessCount": 12
      }
    ],
    "Observation": []
  }
}
```

Requirements:

1. **Grouped by resource type.** Keys are FHIR resource type names; values are lists of saved queries.
2. **Sortable by last access.** Each entry carries a `lastAccessedAt` timestamp that is bumped whenever the query is run. Sorting is by descending `lastAccessedAt` by default; entries never accessed sort by `createdAt`.
3. **CRUD.** Add, rename, update, and delete individual saved queries.
4. **Survives the existing concurrency model.** ETag / `If-Match` semantics must keep working.

## Design questions to settle first

These are the decisions that determine how much work this is — worth agreeing on before writing code.

- **Opaque blob vs. typed model.** Do we keep the document schema-less and let the client own `savedQueries` entirely (zero server changes, fastest path), or do we introduce a server-side type and validate it? The former means no server work at all; the latter buys us validation, a stable contract, and a place to hang server-side behavior.
- **Merge-patch and arrays.** RFC 7386 replaces arrays wholesale — a client that wants to bump `lastAccessedAt` on one query has to `PATCH` the entire array for that resource type. With `If-Match`, two tabs touching the same resource type will collide and one gets a `412`. Options: model each list as an object keyed by query id (so merge-patch can touch a single entry), add a dedicated sub-resource endpoint (e.g. `POST /_user/settings/saved-queries/{resourceType}/{id}/$touch`), or accept the read-retry loop on the client.
- **Where does `lastAccessedAt` get written?** Client-side on query execution (simple, but trusts the client clock and costs an extra round trip per query), or server-side when the query is actually run (accurate, but couples the search path to the settings store).
- **Bounds.** Cap the number of saved queries per resource type and/or per user, and cap the document size. Today nothing bounds the document, and it lives in one row/doc — an unbounded list is a real footgun on every backend.
- **Validation.** Do we validate that the resource type is a real FHIR resource type for the active version, and that the query string parses as a valid search expression?
- **Backends.** Anything here that changes the `501` story for Elasticsearch and S3? (Probably not — this rides on the existing document.)

## Suggested scope

Assuming we start with the opaque-document approach:

- [ ] Agree on the `savedQueries` shape and the concurrency approach (object-keyed-by-id is likely the right call, given merge-patch array semantics)
- [ ] Document the shape as a convention in `crates/persistence/src/core/user_settings.rs`
- [ ] Decide and implement the `lastAccessedAt` write path
- [ ] Add caps (per-resource-type count, total document size) with a clear error
- [ ] Tests: round-trip, single-entry patch without clobbering siblings, sort-by-last-access, cap enforcement, concurrent-write `412`
- [ ] UI: saved-query list per resource type, sorted by last access, with add/rename/delete

Related: #197 (theme toggle, the first real consumer of user settings), #199 (S3 settings store).

angela-helios:
Implements user-maintained saved FHIR queries on the **opaque-document approach** from the issue's suggested scope: the server stays schema-less, the shape is a documented convention, and the client owns the behavior. Closes #234.

## Design decisions (from the issue's open questions)

- **Object keyed by query id, not arrays** — as the issue anticipated, RFC 7386 replaces arrays wholesale, so `savedQueries.<ResourceType>.<id>` lets a merge patch create, rename, touch, or delete a single entry without clobbering siblings. Convention documented in `helios-persistence`'s `user_settings` module docs.
- **`lastAccessedAt` is written client-side** via merge-patch when the query is run — zero coupling of the search path to the settings store, no new endpoint; per-entry patches make the write race-free in practice, and the remaining If-Match races are absorbed (below).
- **Bounds**: the REST handlers now enforce a **256 KiB** whole-document cap (`413`) and **100 saved queries per resource type** (`422`) with clear messages. The bound applies to the *post-merge* document: PATCH merges against the version it read, pins the write to that version, and — when the caller sent no `If-Match` of its own — absorbs a benign concurrent-writer race with a bounded retry.
- **Validation** of resource types / query syntax is deliberately left out (opaque document, client-owned), matching the suggested scope.
- **Backends**: rides entirely on the existing `/_user/settings` document — the S3/Elasticsearch `501` story is unchanged, and the UI shows an explanatory message on `501`.

## REST (`helios-rest`)
- `PUT`/`PATCH /_user/settings` validate structure and caps; everything else (ETag/If-Match, merge-patch semantics) is unchanged.
- Tests: single-entry touch preserving siblings, delete-one-entry via `null`, cap enforcement on PUT **and** on the post-merge PATCH result, array-shape rejection, oversized-document `413`, and stale-`If-Match` `412` on PATCH.

## UI (`helios-ui`)
- New **`/ui/queries`** page: server-rendered shell, list hydrated client-side by `saved-queries.js` from `/_user/settings` (same per-user document and fetch pattern as the theme toggle).
- Unlike `theme.js` (last-write-wins scalar), the script runs the full conditional-request cycle: captures the `ETag`, sends `If-Match` on every patch, and re-reads + retries on `412` — two tabs touching the same type converge instead of erroring.
- Grouped by resource type; sorted by `lastAccessedAt` descending, never-run entries fall back to `createdAt`. Add / rename / delete, plus **Run**, which opens the FHIR search in a new tab and bumps `lastAccessedAt`/`accessCount`.
- Sidebar gains a Saved Queries entry (new bookmark icon) and an `active_page` mechanism for `aria-current`; strings added across en/es/de (parity test green). DOM is built with `textContent` (no user-content innerHTML).

## Verified end to end
Against a running `hfs` (SQLite): page + asset serve, save via the exact PATCH the script sends, touch preserving siblings, stale `If-Match` → `412`, per-type array shape → `422`, delete via `null` patch, and the Run URL returning the search bundle. `helios-rest` user-settings tests 16/16, `helios-ui` 21/21 (incl. i18n parity), persistence user-settings unit tests green.

## Out of scope (per the issue)
- Server-side `lastAccessedAt` stamping on the search path.
- FHIR-type/query-string validation of saved entries.
- Changing the S3/Elasticsearch settings-store `501` (#199 tracks S3).


angela-helios:
Design note from the Search-page mockup review: the Search Builder mock shows a **Recent** chip, so in the target design saved/recent queries surface inside the Search page itself. This PR's `savedQueries` document (keyed-by-id entries with `lastAccessedAt`) is exactly the data that chip would read — when the Search page lands, its builder can consume the same `/_user/settings` document, and the standalone `/ui/queries` page can stay, move, or fold into it as design prefers. Also flagging for design: this PR adds a Saved Queries sidebar entry that isn't in the current Figma nav — placement is Brett's call.

angela-helios:
Pushed `cb79e19d`: the page now follows the hi-fi console's **Search Builder** instead of the three-field form — a featured, editable `GET /Patient?...` URL is the primary control (the resource type comes from the path), **Run** opens the search and records it, and naming it saves it below.

New piece: a **Recent** dropdown backed by a `recentSearches` key in the per-user settings document, so run history roams like everything else here. Deliberately an array (vs. `savedQueries`' id-keyed objects): it's a small bounded cache (10, deduped, newest first) rewritten wholesale on each run, so merge-patch's replace-the-array semantics are exactly right. Convention documented next to `savedQueries` in `helios-persistence`'s `user_settings` module; no server-side changes needed — the PATCH validation only constrains `savedQueries` shape and total document size.

Verified against a running hfs: `recentSearches` round-trips the REST validation, and the page hydrates the dropdown (same-day time vs. date stamps), loads a recent back into the builder, and keeps the saved list as before.

smunini (dismissed):

smunini (approved):

PR #246, opened 2026-07-10
docs(ui): record the History & Versions diff-rendering decision

### LINKED ISSUE #236: Research: htmx-friendly diff rendering for the History & Versions view
## Summary

Pick a diff rendering approach for the **History & Versions** screen (§5 of `docs/ui-requirements.md`), so a user can select two versions of a resource from `_history` and see what changed between them.

This issue is a **research/decision** task. The outcome should be a decision recorded here (and ideally an ADR or a short section in `docs/ui-requirements.md`), not necessarily an implementation.

## Constraints that drive the choice

These come from the existing `helios-ui` crate (`crates/ui/Cargo.toml`, `crates/ui/templates/layouts/base.html`) and are the reason most off-the-shelf diff widgets are a poor fit:

1. **No runtime CDN.** Assets are vendored and embedded at compile time via `rust-embed` + `axum-embed` so the server works in air-gapped deployments.
2. **Markup lives in `templates/`, never in Rust source** — Askama, compile-time checked, auto-escaping.
3. **Progressive enhancement.** htmx swaps fragments, but core views work with JavaScript off (cf. the language switcher, which is deliberately real links).
4. **Content is FHIR JSON**, not source code. The interesting question is "what field changed", not "what line changed".

Constraint 3 is the sharp one: a diff rendered by a client-side JS library is a blank pane with JS disabled.

### Layer 1 — semantic diff: `json-patch` (already a dependency)

`json_patch::diff(&left, &right) -> Patch` produces an RFC 6902 patch. I verified this against the locked version:

- `json-patch` **3.0.1** is already in `Cargo.lock`, pulled in by both `helios-rest` and `helios-persistence`.
- The `diff` function is behind the `diff` feature, which is **on by default** (`default = ["diff"]`), so it is already compiled in today.
- MIT/Apache-2.0.

**Zero new dependencies.** And it is pleasingly symmetric: `crates/rest/src/handlers/patch.rs` already *applies* RFC 6902 patches for the FHIR `PATCH` interaction. Generating one for the history view is the exact inverse operation, using the same crate and the same vocabulary the server already speaks.

This gives a change list a clinician or integrator can actually read:

```
replace  /status                     draft → active
replace  /name/0/family              Smith → Smythe
add      /telecom/1                  { "system": "email", ... }
remove   /identifier/2
```

### Layer 2 — textual diff: [`similar`](https://github.com/mitsuhiko/similar)

For the "show me the raw JSON diff" view, diff the pretty-printed JSON of both versions.

- Apache-2.0, **no dependencies by default**, ~42M recent downloads, last released 2026-05-23 — actively maintained (Armin Ronacher).
- Myers (default), Patience, and LCS algorithms.
- `inline` feature gives `iter_inline_changes` for **word-level intra-line highlighting** — the thing that makes a diff readable when one field in a long line changed.
- Has a `udiff` module for unified-diff output.

Then render both layers into Askama templates and let htmx swap the fragment when the user changes the version selectors. No JavaScript required for the diff to display.

**Known gap, stated plainly:** `similar` has **no HTML output helper**. We iterate `Change`/`InlineChange` ourselves and emit rows from an Askama template. That is a feature, not a problem, given constraint 2 — but it does mean the CSS is ours to write. Cribbing the visual language of [diff2html](https://diff2html.xyz/) (side-by-side, paired lines, +/- gutters) in `app.css` is the cheap path.

## Alternatives considered, and why not

| Option | Why not |
|---|---|
| [**diff2html**](https://diff2html.xyz/) (~168k weekly downloads) | Consumes a *unified diff string*, which we'd have to generate server-side anyway — so we'd take on `similar` **and** a vendored JS bundle **and** highlight.js. It also breaks the JS-off baseline. Its CSS is worth stealing; its runtime is not. |
| [**jsondiffpatch**](https://github.com/benjamine/jsondiffpatch) (~260k weekly downloads) | Genuinely good, and JSON-aware with a real HTML formatter + SVG array-move arrows. But it duplicates in JS what `json-patch` already does in Rust, needs a vendored ESM bundle, and yields nothing with JS off. Worth revisiting **only** if we later want client-side interactivity like collapsible unchanged subtrees. |
| **Monaco / CodeMirror merge view** | Heavyweight editor runtimes for a read-only view. Contradicts the air-gapped, no-CDN, progressive-enhancement stance outright. |
| `imara-diff` | Faster (used by gitoxide), but lower-level and offers no inline/word-level API. Diffing two FHIR resources is not a performance problem. Wrong trade. |

## Open design questions for whoever picks this up

1. **Array churn.** RFC 6902 paths are index-based, so inserting an element at `name[0]` can render as a cascade of `replace` ops rather than one `add`. FHIR arrays (`name`, `telecom`, `identifier`) are exactly where this bites. Do we post-process the patch to detect moves, or accept the noise in v1?
2. **Noise floor.** `meta.versionId` and `meta.lastUpdated` change on *every* version. Filter them from the semantic view by default, behind a "show metadata changes" toggle.
3. **Which two versions?** Adjacent-by-default (`n` vs `n-1`), with two `<select>`s to pick arbitrary pairs. Plain form submission keeps it working without JS; htmx upgrades it to a fragment swap.
4. **Where does the diff get computed?** A `helios-ui` handler that fetches both versions via `vread`, or a new REST endpoint? Note there is no standard FHIR `$diff` operation, so anything server-side and FHIR-facing would be a Helios extension. Keeping it inside the UI crate avoids inventing non-standard API surface.
5. **Deleted versions.** `crates/rest/src/handlers/history.rs` supports the R6 Trial Use delete-history operations. What does a diff against a deleted version render as?

## Suggested acceptance criteria

- [ ] Decision recorded (this issue or an ADR) on the semantic and textual diff approach.
- [ ] Spike: `json_patch::diff` over two versions of a real FHIR resource, with the array-churn behavior in Q1 characterized by an actual example.
- [ ] Confirm `similar` + `inline` renders acceptably through Askama without a JS runtime.
- [ ] Note the `app.css` work implied by having no HTML helper.

## References

- [`similar`](https://github.com/mitsuhiko/similar) · [docs.rs](https://docs.rs/similar)
- [`json-patch`](https://crates.io/crates/json-patch) — already vendored
- [`imara-diff`](https://lib.rs/crates/imara-diff)
- [diff2html](https://diff2html.xyz/) · [jsondiffpatch](https://github.com/benjamine/jsondiffpatch)
- [RFC 6902 — JSON Patch](https://datatracker.ietf.org/doc/html/rfc6902)
- `docs/ui-requirements.md` §5 (branch `docs/ui-requirements`)

angela-helios:
Closes #236. The research issue asked for "a decision recorded here (and ideally an ADR or a short section in `docs/ui-requirements.md`)" — this is that record, as a standing doc in `docs/` (the ui-requirements doc lives on an unmerged branch, so the decision gets its own file the same way `docs/search-conformance-plan.md` does).

Contents: the adopted two-layer server-side approach (`json_patch::diff` semantic view — zero new deps, the inverse of the existing PATCH interaction — plus `similar` with word-level inline highlighting rendered through Askama, no JS runtime), the executed spike output characterizing RFC 6902 array churn on a front-inserted `name[]` element, and the recorded decisions on metadata-noise filtering, version selection, where the diff is computed, and deleted-version rendering. Alternatives table and full discussion stay in the issue.

Implementation is tracked by the History & Versions screen work, not this issue.

### DIFF
--- a/docs/history-diff-rendering.md
+++ b/docs/history-diff-rendering.md
@@ -1,0 +1,70 @@
+# History & Versions: diff rendering (decision record)
+
+**Status:** Decided — see [#236](https://github.com/HeliosSoftware/hfs/issues/236) for the
+full research, alternatives table, and discussion.
+**Scope:** how the History & Versions screen (`docs/ui-requirements.md` §5)
+renders the difference between two versions of a resource from `_history`.
+
+## Decision
+
+Render the diff **server-side, in two layers**, into Askama templates; htmx
+swaps the fragment when the version selectors change, and the base view works
+with JavaScript off (plain form submission). No client-side diff library.
+
+1. **Semantic layer — `json_patch::diff`.** Already a default-feature
+   dependency of `helios-rest` and `helios-persistence` (3.0.1 in
+   `Cargo.lock`), so this adds **zero dependencies**, and it is the exact
+   inverse of the RFC 6902 patches `crates/rest/src/handlers/patch.rs`
+   already applies for the FHIR `PATCH` interaction. Rendered as a
+   field-level change list (`replace /name/0/family Smith → Smythe`).
+2. **Textual layer — [`similar`](https://docs.rs/similar)** (Apache-2.0, no
+   default deps, actively maintained) with the `inline` feature for
+   word-level intra-line highlighting over the pretty-printed JSON of both
+   versions. `similar` has no HTML formatter; we iterate
+   `Change`/`InlineChange` and emit rows from an Askama template — which is
+   what the crate's no-inline-markup rule wants anyway. The diff-table CSS is
+   ours to write in `app.css`; crib diff2html's visual language
+   (side-by-side, paired lines, +/- gutters). The dependency is added when
+   the screen is built, not before.
+
+diff2html, jsondiffpatch, Monaco/CodeMirror merge views, and `imara-diff`
+were evaluated and rejected (CDN/vendored-JS weight, JS-off blank pane, or
+no word-level API) — the comparison table lives in the issue.
+
+## Spike result: RFC 6902 array churn, characterized
+
+`json_patch::diff` over two versions of a Patient where v4 renames a family,
+**inserts a new name at the front of `name[]`**, and appends a telecom:
+
+```
+replace /meta/lastUpdated  → 2026-07-09T12:01:00Z
+replace /meta/versionId    → 4
+replace /name/0/family     → Smythe
+replace /name/0/given/0    → Johanna
+add     /name/0/use        → official
+replace /name/1/given/0    → Jon
+remove  /name/1/use
+add     /name/2            → {family: Smith, given: [J], use: nickname}
+add     /telecom/1         → {system: email, value: j@x.org}
+```
+
+The front-insert renders as a six-op cascade across `name/0..2` instead of a
+single `add /name/0`; the clean append (`telecom/1`) is one op.
+
+## Design decisions
+
+- **Array churn:** accepted for v1. Move detection is a presentation-layer
+  post-process that can be added later without changing the architecture;
+  the textual layer already shows an insert legibly.
+- **Noise floor:** `meta.versionId` / `meta.lastUpdated` change on every
+  version — filtered from the semantic view by default, behind a "show
+  metadata changes" toggle. Verified in the spike: stripping `meta` before
+  `diff` removes exactly those two ops.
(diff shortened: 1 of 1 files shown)


smunini:
Can you implement a prototype and we can evaluate from there.

smunini (changes requested):
Let's do a prototype

smunini (approved):

Issue #238, opened 2026-07-09
SearchParameter viewer & editor UI: Resource Filter rail, faceted browse, CRUD with registry-aware validation


smunini:
## Summary

A viewer and editor for `SearchParameter` resources: browse the 1,375 parameters the server actually resolves searches against, filter down to the ones you care about, and create / update / delete tenant-scoped ones.

**Interactive mockup:** [`docs/mockups/search-parameter-editor.html`](https://github.com/HeliosSoftware/hfs/blob/docs/237-238-mockups/docs/mockups/search-parameter-editor.html) — self-contained, open locally in a browser; PNG preview alongside (PR #247)

Loaded with the **real R4 registry** from `data/search-parameters-r4.json` — 1,375 parameters across 135 base resource types — plus two clearly-marked synthetic tenant parameters to exercise the write paths. Nothing persists.

> **Depends on #235** (*Store SearchParameters in storage; make the in-memory registry a TTL cache over it*). Until storage is the source of truth, the write half of this screen has nowhere to write. The read half works today.

## Reusing the Resource Filter

The left rail is the **Resource Filter** from the Resources screen, reused unchanged in behavior: a search box over resource types, an **All types** row, a **Recently used** group, and a live count next to every type. Here it filters the parameter list by `base`.

This is the right primitive to borrow, because the registry is *already* keyed this way — `params_by_type: HashMap<String, HashMap<String, Arc<SearchParameterDefinition>>>` is literally `base → code → param`. The rail is a direct view of the outer map. Selecting `Patient` narrows 1,375 parameters to the ~40 that resource actually supports.

Two facet rows sit above the table (**type** and **source**), scoped to the current base type and showing live counts:

| Parameter type | Count |
|---|---:|
| token | 536 |
| reference | 472 |
| string | 133 |
| date | 109 |
| composite | 46 |
| uri | 45 |
| quantity | 27 |
| number | 6 |
| special | 1 |

## CRUD model

`SearchParameterSource` (`crates/fhir/src/search/registry.rs:58`) already has the three cases this UI needs — `Embedded`, `Stored`, `Config` — so the editor keys off it rather than inventing a concept:

- **Embedded** (spec, compiled in from the data file) — **read-only**. The action offered is **Create override**: clone it into a `Stored` parameter with the same `code` and a fresh canonical URL.
- **Stored** (POSTed / tenant-scoped) — full create, update, delete. Delete is an inline confirm, never a browser dialog.

[...]

Issue #237, opened 2026-07-09
Compartment Editor UI: CRUD for CompartmentDefinitions + a compartment route tester


smunini:
## Summary

A UI for managing `CompartmentDefinition` resources — create, update, delete — plus a **tester** that answers the question you actually have when a compartment route misbehaves: *is this resource type in this compartment, via which search parameter, and what search does the server run?*

**Interactive mockup:** [`docs/mockups/compartment-editor.html`](https://github.com/HeliosSoftware/hfs/blob/docs/237-238-mockups/docs/mockups/compartment-editor.html) — self-contained, open locally in a browser; PNG preview alongside (PR #247)

The mockup is clickable. It is loaded with the **real R4 CompartmentDefinition set** from `crates/fhir-gen/resources/R4/`, not placeholder data, so the counts and parameter lists below are the ones you'll actually see. Nothing persists.

## Why now

We already ship compartment search end-to-end — `crates/rest/src/handlers/compartment.rs` serves `GET [base]/[compartment]/[id]/[type]` and the `/*` all-types fan-out, backed by `helios_fhir::get_compartment_params()` (codegen'd from the JSON files above), implemented across SQLite, PostgreSQL, Elasticsearch, and MongoDB.

What's missing is any way to *see* it. Today, answering "why does `GET /Patient/123/Observation` return nothing?" means reading a 145-entry JSON file. The `helios-ui` crate already ships a `compartments.svg` icon and a nav slot, so this screen is anticipated — it just has nothing behind it.

## What the mockup shows

**Three panes:** a rail listing the definitions, a tabbed editor, and inline results.

### Definition tab — metadata CRUD
Canonical URL, code, status, version, publisher, description, plus the two switches that actually change server behavior:
- **`search`** — flip it off and *no* compartment route resolves for that compartment. The tester says so explicitly rather than silently returning nothing.
- **`experimental`**

Editing anything marks the definition dirty (dot in the rail, **Save** / **Revert** enabled). **Revert** restores from an untouched base copy, so it's honest rather than decorative.

### Members tab — membership CRUD
The 145 resource types per compartment, filtered by **Members · All types · Excluded**, each row showing its parameters as editable chips. Add a param (type + Enter), remove one, or drop a type from the compartment entirely.

Membership is state, so it's encoded as a chip, not just an absence: `member` (green) / `excluded` (red). Real R4 member counts —

| Compartment | Members | of 145 types |
|---|---:|---:|
| Patient | 66 | 145 |
[...]

PR #252, opened 2026-07-10
feat(ui): SearchParameter viewer (#238) + Compartment viewer & route tester (#237) — read halves

angela-helios:
Read/serve halves of #238 (SearchParameter viewer) and #237 (Compartment viewer + route tester) in `helios-ui`, **re-architected** so primary storage is the single source of truth for conformance resources and the UI reads them over HTTP.

Stacked on #244 (`feat/235-storage-backed-search-params`), now merged; `main` is merged in.

## What changed vs. the original read-only-viewer approach

The viewers no longer read spec data from disk or vendor `CompartmentDefinition` JSON in the UI crate. Instead:

1. **The server seeds the spec resources into primary storage at startup** — SearchParameters (via #244) and CompartmentDefinitions (new), into **every provisioned tenant** (auto-provisioning `default`), idempotently (seed-if-absent).
2. **The plain FHIR routes serve them** — `GET /SearchParameter` and `GET /CompartmentDefinition` read primary storage; no new handlers, no registry/catalog intermediary.
3. **The UI fetches both over HTTP** — a server-side self-call to its own FHIR API (`crates/ui/src/conformance.rs`), degrading to a warning banner on failure.

## Server side

- `data/compartment-definitions-{r4,r4b,r5,r6}.json` bundles built from the spec `CompartmentDefinition`s (degenerate `questionnaire` excluded; R6 adds `group`); `helios_fhir::compartment::CompartmentDefinitionLoader`. A parity test asserts the bundles match the codegen `get_compartment_params` table for every version.
- `seed_spec_compartment_definitions` + combined `seed_tenant_conformance` (`crates/persistence/src/search/seeder.rs`), mirroring #244's seeder: idempotent `create` (`AlreadyExists` ⇒ existing), `count` fast-path.
- `crates/hfs/src/main.rs` seeds both sets across every provisioned tenant on boot; admin (`POST /admin/tenants`) and UI tenant-create seed newly-provisioned tenants.
- **Provisioned-only tenant enforcement** (`crates/rest/src/extractors/tenant.rs`), gated by `HFS_TENANT_REQUIRE_PROVISIONED` (**default off** for backward compatibility; enforced only on registry backends). `default` is auto-provisioned so single-tenant/unauthenticated deployments are unaffected.

## UI

- `conformance.rs`: `ConformanceSource` trait, `HttpConformanceSource` (self-calls `{loopback}/{type}?_count=…`), and a `StaticConformanceSource` for hermetic tests. `mount()` gains `self_base_url`, `outbound_auth`, and `fhir_version`.
- Deletes the vendored `crates/ui/data/compartments/` tree; drops the SearchParameter disk read. View models, templates, and i18n keys are unchanged.

## Tests

- Compartment parity (bundles == codegen table, all versions); compartment seeding (idempotent, discoverable, per-tenant, partial-set); provisioned-only enforcement; `GET /CompartmentDefinition` endpoint; UI pages via injected `StaticConformanceSource`. #244's search-param seeding tests still pass. Verified end-to-end on a live server (seeded 1377 SP + 5 CD; both endpoints serve from storage; both UI pages render real data).

## Per-tenant SearchParameter registry

Follow-on to the storage-backed model: the in-memory SearchParameter registry was a single process-global structure per backend that merged **every tenant's** stored parameters, so a `SearchParameter` POSTed under one tenant resolved searches for all tenants. It is now **per-tenant** — `acme1` and `acme2` operate on different parameter sets and the FHIR search API / CapabilityStatement behave per tenant id.

**Model — shared base + per-tenant overlay.** A new `TenantSearchRegistries` (`crates/persistence/src/search/tenant_registries.rs`) holds one shared **base** registry (embedded + spec + custom, identical for all tenants) and lazily builds each tenant's registry as `base.clone()` + that tenant's **active stored** params (which shadow standard params via the existing #242 precedence). The `SearchParameterRegistry` struct and its shadowing logic are unchanged — there are just N of them.

- `SearchProvider::search_param_registry(&self, tenant) -> Arc<…>` (was `&self -> &Arc`) across all backends (SQLite/Postgres/Mongo/ES/S3) + composite.
- Per-tenant extractor at every index/write site; write-hooks invalidate/reload only the writing tenant; stored-param queries filtered by `tenant_id`; the TTL refresh drops the per-tenant caches; **`/metadata` is now per-tenant**. ES shares the container (its loader points at the primary's DB); reindex builds the extractor per target tenant.
- Draft `SearchParameter` creates skip invalidation (only active params overlay), so bulk seeding doesn't trigger an O(n²) rebuild storm.
- New `HFS_SEED_CONFORMANCE` (default on) gates the conformance seeding (startup + on-provision).

Tested end-to-end through the router (`crates/rest/tests/per_tenant_search_params.rs`): a custom param under `acme1` resolves for `acme1`, is rejected for `acme2`, and appears in `acme1`'s `/metadata` only.

## Follow-ups

- **Service token (auth-on):** the UI self-call currently uses the existing static-bearer `OutboundAuthProvider` (`HFS_OUTBOUND_BEARER_TOKEN`), so an auth-enabled deployment must provision a valid token or the conformance pages degrade. The follow-up is to mint a short-lived, auto-refreshed `system/*.rs` token via the planned `JwtAssertionOutboundAuthProvider` (SMART Backend Services `client_credentials` + `private_key_jwt`) from `HFS_UI_*` credentials. Documented at the wiring point (`TODO(service-token)` in `crates/hfs/src/main.rs`) and in `crates/auth/src/outbound.rs`.
- **Single-version storage:** seeding uses `default_fhir_version`, so the UI version pill reflects what the server runs, not every built version. The SearchParameter viewer's source/shadow chips lose their backing once data is plain stored resources.
- Postgres/Mongo reload the full stored-param set (all tenants) on any active `SearchParameter` write; fine given how rare those writes are, but could be narrowed to the writing tenant.
- Enabling `HFS_TENANT_REQUIRE_PROVISIONED` by default (would require updating integration tests that use ad-hoc tenant ids).

https://claude.ai/code/session_01Bb6qw4W9z6zWqKEsCmC2ub


angela-helios:
`90b44da7` grows the queries page's builder to the rest of the hi-fi Search screen — it lives on this branch (not #241) because the parameter catalog is fed by #238's registry snapshot:

- **Visual rows ⇄ URL**: Conditions (param / modifier / value), Includes, and Result controls stay in two-way sync with the featured GET URL. Colon modifiers rejoin the key, comparator prefixes (`ge`, `le`, …) rejoin the value.
- **Registry-fed suggestions**: `GET /ui/queries/params?type=X` returns a server-rendered `<datalist>` fragment (options labeled with the param type) from the same snapshot as `/ui/search-parameters`, swapped as the URL's type changes. Hypermedia, not a UI JSON API.
- **In-page results**: Run fetches the search from the FHIR API itself and renders `Bundle.total`, a table honoring `_elements` (compact HumanName/CodeableConcept/Reference/Quantity formatting), OperationOutcome diagnostics on errors, and paging over `Bundle.link`. Explicit runs record to `recentSearches`; paging doesn't. `?url=` deep-links into a loaded builder + executed search.

Verified against a running hfs with seeded Patients (screenshots below reflect `/ui/queries?url=/Patient?name=smith&_elements=name,gender,birthDate`). The hi-fi's **AI search** panel is the remaining piece of that screen — needs an NL→FHIR-URL backend, so left as a follow-up.

PR #259, opened 2026-07-14
feat(persistence): per-user SettingsStore on the S3 backend (#199)

### LINKED ISSUE #199: Support per-user UI settings (SettingsStore) on the S3 backend

mauripunzueta:
Closes #199.

## Why

`HFS_STORAGE_BACKEND=s3` returned `501` from `GET/PUT/PATCH /_user/settings`. S3 was the last standalone primary backend without a `SettingsStore` — SQLite, PostgreSQL and MongoDB already have one. Elasticsearch remains out of scope (search-only, never a standalone primary), so this completes the set.

## How

S3 has no transactions, so the trait's read-modify-write + monotonic `version` is mapped onto a **compare-and-swap over conditional `PutObject`**: `If-None-Match: *` to create, `If-Match: {etag}` to update.

- A settings write is **never** an unconditional put — that is what a silent lost update would look like.
- An **unconditional** writer that loses the race re-reads and retries (bounded, with jittered backoff — the AWS SDK never retries a 412).
- A **conditional** (`If-Match`) writer that loses the race gets `ConcurrencyError::OptimisticLockFailure` and does *not* retry.
- The client-facing `version` lives in the object body; the S3 ETag is a separate, internal CAS token. The two are never conflated, and a client's `If-Match: W/"6"` is never forwarded to S3.

The existing FHIR CRUD path already hard-depends on conditional `PutObject` (`storage.rs:454,536,652,716`), so this adds **no new requirement** on the object store.

## Object naming (the security-critical bit)

The object is named `sha256_hex(user_key)`, never the key itself.

`user_key` is `"{issuer}|{subject}"` built from unvalidated JWT claims: it can contain `/` or `..`, be empty, or be arbitrarily long. The keyspace's existing `sanitize()` is **lossy** (both `/` and space map to `_`), so using it here would let two distinct principals collide on one object — a cross-user settings read/write. Hashing is injective, path-safe (cannot escape the settings prefix into the FHIR `resources/` keyspace), fixed-length (S3 caps keys at 1024 bytes), and keeps issuer/subject out of S3 access logs and bucket listings.

The user key is stored *in the body* and asserted on every read, so any future regression in the key derivation fails loudly rather than silently serving one user another user's settings.

## Tenancy

Settings are user-global (the trait takes no `TenantContext`), so `settings_location()` resolves a bucket without one and builds the key from the global prefix with **no tenant segment**. In `BucketPerTenant` mode it uses `default_system_bucket` and errors actionably when unset. The `_system.user-settings` segment contains a `.`, which tenant IDs cannot (`[A-Za-z0-9_-]{1,64}`) — so no tenant prefix can ever reach the settings namespace, and no bucket scan enumerates it.

## Drive-by fixes

- **`ConditionalRequestConflict` (HTTP 409)** — returned when two conditional writes race — was falling through to `Internal` and surfacing as a **500 on the existing resource paths too**. Now mapped to a failed precondition. A *bare* 409 is deliberately **not** mapped: it is shared by transient conflicts like `OperationAborted`, and the resource paths turn a precondition failure into `AlreadyExists`/`VersionConflict`, so misreading one would state a falsehood about stored state.
- **`tokio`'s `time` feature** is now explicit in `helios-persistence`. The MongoDB store's backoff already needed it and only compiled via feature unification through the mongo driver.
- Every startup path now uses `create_app_with_auth_bulk_and_settings`, which incidentally fixes a **mongodb-only build** that failed to resolve the `cfg(any(sqlite, postgres))`-gated `SettingsStore` import.
- Removed an `Arc::try_unwrap(...).unwrap_or_else(|| unreachable!())` panic path in `start_s3_elasticsearch`.

## Tests

**Mock-backed unit tests** (`s3/tests.rs`) — round-trip, version increments, merge-patch incl. null-delete, optimistic locking, *update-race retry preserves the concurrent writer's update*, *create-race retry*, conditional write does **not** retry (asserted via put count), bounded give-up, corrupt body, cross-user tripwire, key injectivity for keys a lossy sanitiser would collide, tenant independence, and global-prefix handling.

The mock records the preconditions of every `PutObject`, and one test asserts **no settings write is ever unconditional** — a regression guard against silent lost updates.

**MinIO integration tests** (`minio_s3_tests.rs`, opt-in via `RUN_MINIO_S3_TESTS=1`) prove a *real* S3-compatible store actually enforces the preconditions — which is the whole risk of the design. Includes an 8-way create race (exactly one winner) and a 4-way concurrent merge-patch that **fails if any writer's update is lost**.

## Note for the reviewer

This was developed on a machine with no C linker, so it has not been compiled or run locally — `cargo fmt --check` is clean, and correctness was reviewed by hand, but **CI is the first real compile/test gate**. Flagging that honestly rather than implying it was verified.

Left as **draft** pending a green CI run.


mauripunzueta:
Reopening to re-trigger CI: GitHub delivered no workflow run for a2ccf291 (0 runs, 0 check-runs on the head SHA), so the PR shows no checks. No code change — this commit is docs only.

smunini (approved):

PR #260, opened 2026-07-14
perf(postgres): fix search collapse at volume (#224)

### LINKED ISSUE #224: Postgres FHIR search is slow at full data volume (med ~7s, p99 ~51s, one composite query 500s)
## Summary

At full Synthea data volume (~1.3M `search_index` rows), the **Postgres** backend serves FHIR search **slowly under concurrent load**: median request ~7s, p99 ~51s, and at least one query type consistently times out with a 500. This surfaced while wiring HFS into the HealthSamurai FHIR Server Performance Benchmark (#190). CRUD and import are healthy; **search latency at volume is the sole remaining problem** and is a genuine query-plan/concurrency issue, not a stale-statistics artifact.

This is split out of PR #219 (which lands the validated bug fixes + config + truthful reporting) so it can be diagnosed as a focused perf investigation.

## Evidence

Postgres, `k6/search.js`, 30 VUs / 2m, against a fully-imported DB (~1.3M rows). Two runs on branch `ci/fhir-benchmark-reporting-fixes`:

| Metric | Run 28908173123 (pre-ANALYZE) | Run 28911652793 (post-ANALYZE) |
|---|---|---|
| checks pass | — | 99.77% (449/450) |
| median | ~8s | **7.18s** |
| p90 | — | 16.85s |
| p95 | — | 21.83s |
| p99 | ~45s | **51.04s** |
| max | 62s | **65.51s** |
| throughput | (collapsed from ~11.5k) | 453 reqs / 3.0 rps |
| hard failures | 3 timeouts | **1** (500 @ 39.5s) |

Import in the same run is healthy for comparison: **0.1% failed (2/1700), median 99.7ms**. So the DB is well-populated and writes are fine — reads at volume are the issue.

The single failing query is **`GET /Observation?combo-code-value-quantity=...`** — a composite search that hits the 30s `statement_timeout` and returns 500 (`internal error in postgres: Failed to execute search: db error`). Every other query type passes its check but shares the high aggregate latency.

## What we ruled out: stale planner statistics

Hypothesis was that a freshly bulk-loaded Postgres has no table stats, so the planner seq-scans `search_index` despite the indexes existing. We added an `ANALYZE` step after the import suite (commit f57f4957b) and validated it in run **28911652793**. Result: **no measurable effect** — latency (med 7.18s / p99 51s / max 65s) is statistically identical to the pre-ANALYZE run. That commit has been **reverted** (a39fc5ca4); the slowness is not a cold-stats artifact.

## Likely root cause (to confirm)

A genuine query-plan / index-coverage / concurrency problem at volume — candidates to investigate:
- Missing/insufficiently **covering indexes** on `search_index` for the hot search shapes (especially composite/quantity queries like `combo-code-value-quantity`, `code-value-quantity`, `component-code-value-quantity`).
- Query structure forcing large intermediate result sets or repeated index probes per matching row under 30-VU concurrency.
- Interaction with the modestly-tuned ephemeral container (`shared_buffers=2GB`, `effective_cache_size=6GB`, scaled down from the benchmark's published 10G/25G) — buffer-cache pressure under concurrent scans.

## Suggested next steps

1. Add a **per-query-tagged / `EXPLAIN (ANALYZE, BUFFERS)`** debug pass in the benchmark workflow (or reproduce locally) to attribute latency per search shape and capture the actual plans — the k6 summary only gives aggregate `http_req_duration`, so we can't yet attribute the 7s median to specific query types.
2. Focus first on the composite/quantity searches (the one confirmed timeout is `combo-code-value-quantity`).
3. Evaluate covering indexes and/or query-builder changes for the hot shapes (same class of work as the HTS PG perf push — closure/covering-index/paginated-SQL).
4. Re-run the benchmark to confirm latency collapses and the 500 disappears.

## Reproduction

Run the **FHIR Server Benchmark** workflow (`.github/workflows/fhir-benchmark.yml`), `backend=postgres`, `tests=import,search`. Import populates ~1.3M rows; search runs immediately after at 30 VUs. Artifacts (`fhir-benchmark-postgres-<run_id>`) contain `search.json` / `search.log` with the numbers above.

## Caveats

Absolute latency on the self-hosted runner is **relative-only** — the ephemeral Postgres is tuned down from the benchmark host (8vCPU/30G DB) to fit the runner. Numbers directly comparable to the public report need matched hardware. But a 7s median at 30 VUs is a real problem regardless of hardware and warrants a query/index fix.

Refs #190. Follow-up from PR #219.

mauripunzueta:
Fixes #224 (partially — see "What is still broken").

At ~10M `search_index` rows the Postgres backend served FHIR search at a **7.18s median / 51s p99**, and `Observation?combo-code-value-quantity=` reliably timed out with a 500.

## Result (measured in CI, postgres, 30 VUs / 2m)

| metric | baseline (#224) | this PR | |
|---|---|---|---|
| **median** | 7,180 ms | **48 ms** | **149x** |
| p90 | 16,850 ms | 5,909 ms | 2.9x |
| p95 | 21,830 ms | 12,761 ms | 1.7x |
| p99 | 51,040 ms | 30,011 ms | capped |
| max | 65,510 ms | 30,077 ms | capped |
| **throughput** | 3.0 rps | **14.2 rps** | **4.7x** |
| hard failures | 1 | 32 (1.5%) | see below |

The failure count rising is **the truth surfacing, not a regression**. Every failure is a search cut at exactly 30.0s by `statement_timeout` — which previously was armed on 1 pooled connection in 10, so runaway queries mostly ran free to 65s. Now every connection enforces it.

Per-shape (the attribution #224 said we lacked — this PR adds it to the workflow):

| shape | before | after |
|---|---|---|
| `Encounter?status` p95 | 26,959 ms | **54 ms** |
| `Observation?component-value-quantity` median | 4,756 ms | **236 ms** |
| `Observation?combo-value-quantity` p95 | 23,481 ms | **1,445 ms** |
| `Encounter?class` median | 1,362 ms | **50 ms** |
| `Patient?address` median | 249 ms | **52 ms** |
| `Organization?name` / `Patient?name` median | full table scans | 29 / 40 ms |
| all 7 `reference` shapes | — | 11–75 ms |

## The fixes

**Statistics — a 506x win on its own.** `search_index` is one wide table, so `value_token_code` mixes LOINC codes, Encounter statuses and ActCodes in one column. Every token search binds `resource_type` **and** `param_name` **and** the code, and the three are near-perfectly correlated. Postgres multiplied the marginals independently: `Encounter?status=finished` matches **all 65,659** Encounters but was estimated at **1,832** — a 36x under-estimate — so the planner materialised every id, heap-fetched all 65k rows to sort them, and returned 21 (5,261 ms). A 3-column MCV makes it walk the ordered index and stop after ~21 rows: **10.4 ms, 4 buffers**.

This is also **why the ANALYZE tried earlier in #224 did nothing**: no amount of per-column statistics can express a cross-column correlation. (Second reason: after a bulk load `relallvisible` is 0 until a **VACUUM** runs, so index-only scans silently degrade to heap fetches. ANALYZE never sets the visibility map. The workflow now VACUUMs.)

**Composite** had no value predicate in its `WHERE` at all — it aggregated the parameter's whole slice on every request. Now prefiltered. The `GROUP BY resource_id, composite_group` + per-value `HAVING` are untouched: they are what force all components into the *same* group.

**OR'd sublinks never become semi-joins.** Token/reference OR-lists emitted `id IN (…) OR id IN (…)`; Postgres only pulls a sublink up into a semi-join from the top-level `AND`-list, so under an `OR` it stays a re-executable `SubPlan`.

**`ILIKE` over a `COALESCE` is unindexable.** Replaced with a `LIKE` against `COALESCE(value_string_folded, lower(value_string))` — semantically identical (`fold_text` already lowercases; `lower()` on the raw fallback reproduces what `ILIKE` did) but indexable. The fallback is load-bearing: `value_string_folded` is populated on write and **never backfilled**.

**The pool was pinned at 10 and unreachable** — `parse_connection_string` never consulted `HFS_PG_MAX_CONNECTIONS`, and `HFS_DATABASE_URL` takes that path. By Little's Law ~2/3 of every request's latency was queueing for a connection that did not exist.

**`statement_timeout` was set on one connection of ten** (a session GUC, `SET` on one borrowed client). This resolves the paradox in the issue — a "30s timeout" producing a 500 at 39.5s and a max of 65.5s.

**The benchmark built a debug binary.** Now `--release`.

## Rejected on real-data measurement (documented in-code)

Every one of these was ranked *highly* by a local replica and is **wrong** against the real 656,737-Observation dataset:

| candidate | real cost |
|---|---|
| correlated `EXISTS` composite | 8,920 ms on a match; 20,095 ms on a zero-match |
| drop the prefilter + covering index | 16,465 ms |
| self-join on `composite_group` | flattened → same ordered-scan trap |
| `EXISTS`-for-`IN` | no effect; Postgres already flattens `IN` |
| dropping `idx_search_resource` as redundant | it takes ~12M scans in a 30s run |

Only **222 of 656,737** Observations match the composite, so any plan that walks `resources` in sort order and probes per row must traverse ~190k rows to collect 21 hits. The shipped grouped subquery **cannot be flattened** — which forces Postgres to build the small candidate set and probe by primary key. Its unflattenability is a feature.

## What is still broken

**Composite is still ~12s at 30 VUs, and no query or index shape fixes it.** The Bitmap Heap Scan fetches ~113k rows in ~112k buffer **reads** — nearly one random page per row, because a parameter's rows are scattered across a ~10M-row table. That is I/O. The real fix is a **storage change**: write one row per composite group with every component's value column populated, so a single index answers `code = X AND value > Y within one group` directly. Needs a writer change, a migration and a backfill — worth its own issue.

**A sparse-match tail.** With accurate statistics Postgres now picks the ordered-index scan more often, which is right when many rows match and wrong when almost none do (it walks the whole type). `Observation?code` p95 regressed 5.7s → 30s and `Encounter?date` p95 2.6s → 17.8s, while `status`/`class`/quantity improved far more. Net p95 is 1.7x better and throughput 23% better, but this is a real trade and the next thing to attack.

**Import** median moved 99.7 ms → 530 ms. Not yet attributed (build mode, four new indexes, and pool 10→32 all changed at once). Needs isolating before merge.

## Safety

Schema **v14 is additive only** — no index dropped, no query semantics changed; ~6s on 1.45M rows. `initialize_schema` now takes an advisory lock, because several instances routinely share one database (`cluster-smoke` runs two) and `schema_version` is read-then-written without a transaction.

Three new tests, all green on the **old** code too:
- **composite components must share a `composite_group`** — a two-group blood-pressure panel. The pre-existing single-group test is structurally blind to the cross-group false positive that a careless composite rewrite ships.
- composite OR-list values stay isolated across values.
- string search still matches rows whose `value_string_folded` is NULL.

CI green.


mauripunzueta:
## First real benchmark run — big win, but the headline bug is **not** fixed

Run [29342370727](https://github.com/HeliosSoftware/hfs/actions/runs/29342370727), postgres, 30 VUs / 2m, release build.

| metric | baseline (#224) | this run | |
|---|---|---|---|
| **median** | 7,180 ms | **48.5 ms** | ✅ 148x |
| p90 | 16,850 ms | 7,630 ms | ✅ |
| p95 | 21,830 ms | 18,659 ms | ~flat |
| p99 | 51,040 ms | 30,006 ms | ✅ capped |
| max | 65,510 ms | 30,073 ms | ✅ capped |
| throughput | 3.0 rps | **11.5 rps** | ✅ 3.8x |
| hard failures | 1 | 27 (1.56%) | ⚠️ see below |

**The failure increase is the truth surfacing, not a regression.** All 27 are searches at exactly 30,002–30,009 ms — the `statement_timeout` firing. It was previously armed on 1 pooled connection in 10, so runaway queries mostly ran free to 65 s. Now every connection enforces it, and p99/max are pinned at exactly 30.0 s. That is the fix working.

### Per-shape attribution (the thing #224 said we couldn't see)

**Fixed:**

| shape | median |
|---|---|
| `Organization?name` | 19 ms |
| `Patient?name` | 26 ms |
| `Observation?category` | 28 ms |
| all 7 `reference` shapes | 4–69 ms |
| `Patient?birthdate` / `Observation?date` | 52–70 ms |

**Still broken:**

| shape | median | p95 |
|---|---|---|
| `Observation?code-value-quantity` | **13,190 ms** | 30,015 |
| `Observation?combo-code-value-quantity` | **12,271 ms** | 30,019 |
| `Observation?component-code-value-quantity` | 3,088 ms | 5,531 |
| `Encounter?status` | 8 ms | **26,959 ms** |
| `Encounter?class` | 1,362 ms | 24,062 ms |
| `Observation?combo-value-quantity` | 169 ms | 23,481 ms |

Two distinct problems remain:

1. **The composite timeout — the original bug — is still there.** The prefilter measured ~70 ms against a synthetic 1.45M-row replica but runs at 12–13 s here. The replica's data distribution is evidently not representative (real Synthea Observations carry many more composite groups per resource), so my local A/B flattered it. **I should not have trusted the replica over real data on this shape.**

2. **A bimodal tail on `status` / `class` / `combo-value-quantity`** — median 8 ms, p95 27 s. That is the zero-match cliff documented in the PR body: `status=missing-status` gives the ordered index scan no way to know the set is empty, so it walks every Encounter. I measured that as a bounded ~2x on the replica and accepted it; at real volume it is not bounded, and I under-weighted the risk.

### Note on the `EXISTS` rejection

I rejected the correlated-`EXISTS` composite because it cost 440 ms on a zero-match versus a 70 ms baseline. Against a **13,000 ms** baseline that trade is obviously worth taking. The rejection was correct given the numbers I had and wrong given the numbers that matter.

### Next

Run [29349631586](https://github.com/HeliosSoftware/hfs/actions/runs/29349631586) captures `EXPLAIN (ANALYZE, BUFFERS)` for these shapes **against the real imported dataset**, and A/Bs the `EXISTS` candidate in the same pass — including its zero-match case, and whether disabling the ordered scan removes the cliff (which would make the tail a planner-choice problem rather than missing index coverage).

### Also open

Import median moved 99.7 ms → 530 ms. Not yet attributed — the build mode, the four new indexes, and the pool bump (10 → 32, which triples concurrent writers on a write-heavy suite) all changed at once. Needs isolating before this merges.

mauripunzueta:
## Real-data plans: two of my own hypotheses were wrong

Run [29349631586](https://github.com/HeliosSoftware/hfs/actions/runs/29349631586) + [29354381099](https://github.com/HeliosSoftware/hfs/actions/runs/29354381099) captured `EXPLAIN (ANALYZE, BUFFERS)` against the real imported dataset.

### First: the dataset is 8x bigger than the issue says

The issue states "~1.3M `search_index` rows". In reality `code` **alone** is 1.3M:

| param | rows | resources |
|---|---:|---:|
| `combo-code-value-quantity` | 2,463,191 | 656,737 |
| `code-value-quantity` | 1,805,918 | 656,737 |
| `code` | 1,315,695 | 656,737 |

656,737 Observations. The real index is ~10M rows. Every local estimate I made was against a replica ~14x too small on the parameter that matters, and it ranked the candidate fixes **backwards**.

### Wrong hypothesis 1: "the zero-match cliff"

There is no zero-match cliff. `Encounter?status=missing-status` runs in **0.8 ms**.

The slow token case is the **opposite** — `status=finished` matches *all* 65,659 Encounters:

| | time |
|---|---|
| `status=finished`, 2-col stats (shipped) | **5,261 ms** |
| `status=finished`, **3-col MCV** | **10.4 ms** (4 buffers) |
| `status=missing-status`, 3-col (control) | 0.05 ms |
| `category=laboratory`, 3-col (control) | 11.8 ms |

**506x, from a statistics change alone.** Postgres estimated the token scan at `rows=1832` against an actual 65,659 — a **36x under-estimate** — so it materialised 65k ids, heap-fetched all of them to sort, and returned 21. The v14 statistics covered `(param_name, value_token_code)`, but the query also binds `resource_type` and all three are near-perfectly correlated; the marginals were still being multiplied independently. Fixed with a 3-column MCV — pushed in 406de2ae.

### Wrong hypothesis 2: "switch composite to correlated EXISTS"

I said in the previous comment that the `EXISTS` trade was "obviously worth taking". It is not — it is catastrophic:

| composite variant | time |
|---|---|
| shipped (OR-prefilter + IN) | **1,158 ms** |
| correlated `EXISTS` (match) | 8,920 ms |
| correlated `EXISTS` (zero-match) | 20,095 ms |
| no prefilter + covering index | 16,465 ms |

Only **222 of 656,737** Observations match this composite (0.03%). Any plan that walks `resources` in sort order and probes per row must traverse ~190k rows to collect 21 hits. The shipped grouped subquery **cannot be flattened or parameterized**, which forces Postgres to build the small candidate set first and probe by primary key — the only viable shape for a sparse composite. **Its unflattenability is a feature.** Removing the prefilter also makes it 14x worse, so that is load-bearing too.

Good thing this was measured rather than shipped on the reasoning in my last comment.

### What is still broken

Composite remains ~12 s at 30 VUs, and no query or index shape fixes it. The Bitmap Heap Scan fetches ~113k rows in **~112k buffer READS** — nearly one random page per row, because a parameter's rows are scattered across a ~10M-row table. That is I/O, not CPU or plan choice.

The real fix is a **storage change**: write one row per composite group with every component's value column populated, so a single index answers `code = X AND value > Y within one group` directly. That needs a writer change, a migration and a backfill — too big for this PR, and worth its own issue.

### Method note

The composite numbers cannot be A/B'd on a laptop-sized replica: the cost is random heap I/O on a 10M-row table, and any replica that fits in cache has no such cost. Mine ranked the covering-index variant *last* when in CI it is the only one that touches the I/O at all. Read `Buffers: shared read=`, not `Execution Time`.

mauripunzueta:
Marking ready. The two remaining problems are spun off so they don't block this 149x median / 4.7x throughput win:

- **#279** — Postgres composite search is I/O-bound at volume; needs a denormalized one-row-per-group storage layout (writer + migration + backfill). This is the composite ~12s / 500 timeout that this PR could not fix by query or index shape.
- **#281** — accurate statistics now favor the ordered-index scan, which regressed the sparse/zero-match `Observation?code` and `Encounter?date` p95. Two-sided trade; net still positive.

Still open on this PR before merge: the **import median regression** (99.7ms → 530ms), which needs isolating — build mode, four new indexes, and pool 10→32 all changed at once. Happy to attribute that here or in a follow-up, whichever you prefer.

smunini (approved):

PR #261, opened 2026-07-14
fix(hts): resolve versionless canonicals to the owning package, not a copy (#200)

### LINKED ISSUE #200: HTS: $validate-code default-resolves audit-event-type to incomplete 4.0.1 fragment instead of complete THO 1.0.0
## Summary

`hts.heliossoftware.com` holds two versions of the CodeSystem `http://terminology.hl7.org/CodeSystem/audit-event-type` for the same canonical URL:

- **`4.0.1`** (the copy bundled with FHIR R4 core) — defines only `rest`
- **`1.0.0`** (the complete standalone THO CodeSystem) — defines `rest`, `hl7-v2`, `hl7-v3`, `document`, `object`

A bare `$validate-code` call (canonical URL + code, no `version`) **default-resolves to the incomplete `4.0.1` fragment**, so valid THO codes such as `object` are reported as unknown.

## Reproduction

```
GET https://hts.heliossoftware.com/CodeSystem/$validate-code?url=http://terminology.hl7.org/CodeSystem/audit-event-type&code=object
=> result: false — "Unknown code 'object' ... version '4.0.1'"

GET .../$validate-code?url=...&code=object&version=1.0.0
=> result: true
```

For comparison, `tx.fhir.org/r5` default-resolves the same bare call to `1.0.0` and returns `true`; `tx.fhir.org/r4` behaves like HTS (resolves to `4.0.1`, returns `false`).

## Impact

Any client validating bare (system, code) pairs against HTS gets false negatives for the four codes that exist only in THO `1.0.0`. This is the same "one canonical URL → multiple versions → resolver picks the wrong row" shape we've hit before. Surfaced by #183 / #184 (audit-events CI validator), which is currently worked around client-side by pinning `version=1.0.0` in `validate_report.py`.

## Suggested direction

When multiple versions of a CodeSystem share a canonical URL and no `version` is supplied, prefer the complete/most-recent standalone THO definition over the stripped FHIR-core fragment (or don't load the fragment when the complete version is present). Removing the client-side version pin in `validate_report.py` would then be the acceptance check.

Also worth noting: HTS appears to only honor the `version` query parameter for `CodeSystem/$validate-code` — `systemVersion` / `system-version` were ignored in testing.

Follow-up to #183.

mauripunzueta:
Closes #200.

## The bug is not a version-sorting bug

A bare `$validate-code` on `http://terminology.hl7.org/CodeSystem/audit-event-type` rejects valid THO codes such as `object`, resolving to the truncated `4.0.1` row instead of the complete THO `1.0.0` one.

The two rows are near-twins — same FHIR id, same name, same status, **both `content: complete`**, both non-empty:

| | version | publisher | content | concepts |
|---|---|---|---|---|
| `hl7.fhir.r4.core` copy | `4.0.1` | NEMA/DICOM | complete | 1 (`rest`) |
| `hl7.terminology` original | `1.0.0` | HL7 | complete | 5 (`rest`, `hl7-v2`, `hl7-v3`, `document`, `object`) |

So **a semver fix would not help** (`4.0.1` > `1.0.0` under any ordering — the `4.0.1` is the FHIR *release* version, not the code system's own), and **a content-based rule would not help** (both are `complete`). Nothing intrinsic to the rows separates them. Only provenance does.

## It is much wider than `audit-event-type`

`hl7.fhir.r4.core` declares canonical `http://hl7.org/fhir` yet re-ships **746 CodeSystems and 603 ValueSets** owned by `terminology.hl7.org`. Measured against the vendored packages:

- **208 CodeSystems** currently resolve to the stale core copy
- in **32** of them the winning row has fewer concepts → **1,297 concepts are unvalidatable** via a bare call

Confirmed live on `hts.heliossoftware.com` — `claimcareteamrole` + `rehab` fails exactly like `audit-event-type` + `object`.

## The rule

Every FHIR NPM package declares a `canonical` base in `package.json`. **A package publishing a URL under its own base owns that resource; publishing a URL outside its base is re-publishing someone else's.** Recorded at import as `authority_rank`, used as one tier of same-URL precedence:

- `0` `OWNER` — **positive evidence only**: the package declares a `canonical` base that covers the URL, or a native importer (SNOMED RF2, LOINC, RxNorm, …) loaded the publisher's own distribution.
- `1` `UNKNOWN` — the default for **everything else**: a resource arriving with no package context (REST write, `$import-bundle`), or a package that does not declare a base covering the URL.

No hardcoded URL list, no hand-maintained package-priority table; it self-maintains as packages are added, and covers ValueSets for free. The rule only ever **promotes on positive evidence, never demotes on its absence** — this is deliberate and load-bearing. A resource arriving outside any package stays `unknown`: it neither claims ownership nor is branded a copy, so it still wins its own URLs outright (nothing else ships them) but does **not** outrank an owning package's copy of a *shared* URL. This asymmetry is why `us.nlm.vsac`, which declares a fhir.org registry base yet ships 14,850 `cts.nlm.nih.gov` ValueSets, keeps those URLs instead of being mislabelled a foreign copy — and why replaying `hl7.fhir.r4.core`'s truncated copy through `$import-bundle` cannot stamp itself authoritative and permanently reinstate this bug.

Order: `content` → `has-concepts` → **`authority_rank`** → `version` → `id`.

The new tier sits **below** the first two on purpose: a populated copy must still beat an empty stub from the owning package, and those tiers are the only thing keeping SNOMED `version=current` resolving to the real edition (the stub's version string is literally `"current"`, and `'c' > '2'`, so it text-sorts *above* `"20260501"`).

Copies are **ranked, not discarded** — `version=4.0.1` still resolves, and `resource_json` stays byte-faithful to what the package shipped.

## Upgrading an existing database (the part that would otherwise no-op)

`authority_rank` cannot be backfilled — nothing already stored on a row says which package supplied it. So:

- the column is **nullable**; `NULL` means "never claimed" and readers coalesce it to `0`, i.e. **an un-re-imported database behaves exactly as before**
- adding the column is the signal that the DB predates provenance, so the migration **clears the `.tgz` rows from the `bootstrap_imports` ledger** and the next startup re-imports those packages

Without that, the ledger (which skips files whose size and mtime are unchanged) would leave every row at its default and this fix would be a **silent no-op on the live server**. Only `.tgz` entries are cleared — the multi-GB SNOMED/LOINC/RxNorm archives are not re-imported.

## Folded in, because it is the same defect

- **Unified same-URL resolution.** It had no home (the trait has no resolve method), so it was reimplemented in backend SQL at **34 sites in three mutually inconsistent variants**. All now share `cs_precedence_order_by` / `vs_precedence_order_by`, embedded verbatim by both backends so they cannot drift. This fixes a **latent Postgres-only bug**: `resolve_system_id_pg` lacked the content and has-concepts tiers while its doc comment claimed it matched SQLite — so on Postgres an empty `content=not-present` stub could beat a real import and `$expand` returned nothing.
- **The "unknown code" diagnostic** re-queried by URL with its own ordering, so a miss could name a version the code was never validated against. It now reports the row actually resolved.
- **`CodeSystem/$validate-code` now honours `systemVersion` / `system-version` / `force-system-version` / `check-system-version`** — already honoured on `$expand` and `ValueSet/$validate-code`, silently ignored here (the second defect in #200). Caching is unaffected: the canonical pins already disable the handler cache.

Both backends are covered, per the multi-DB requirement.

## Tests

Regression tests drive the **real `.tgz` importer** (a plain `$import-bundle` carries no provenance, so it could not exercise the mechanism) on **SQLite and Postgres**:

- bare `$validate-code` resolves to the owning package's complete row
- the outcome is **independent of import order** (bootstrap walks the directory alphabetically, which is what let the copy win)
- an explicit `version=4.0.1` still reaches the copy
- a miss reports the resolved version, not the highest version string
- `authority_rank_for` unit coverage, including the fail-open case (package declares no canonical)

## Verification caveat

This environment has **no C linker**, so I could not run `cargo test` locally. Every changed file parses (rustfmt), and I executed the real schema and the real SQL against a live SQLite covering: both import orders, the legacy-DB upgrade path (unchanged before re-import, fixed after), the tx-ecosystem `vnn` fixtures (tie on rank → highest version still wins), the SNOMED `version=current` canary, and empty-owner-vs-populated-copy. CI is the first real compile and test run — reviewing its result is the next step.

## Acceptance

`GET /CodeSystem/$validate-code?url=http://terminology.hl7.org/CodeSystem/audit-event-type&code=object` → `result: true`, so the `TX_SYSTEM_VERSION_OVERRIDES` pin in `validate_report.py` (on the unmerged #183 branch) is no longer needed.


mauripunzueta:
✅ **Update (resolved):** The inherited `Test Rust` failure (`value_set_ops::expand_is_a_filter_with_text_filter_still_nests` → `no such table: concepts_search_fts`) is fixed on `main`. The missing `concepts_search_fts` FTS5 table was a pre-existing break on `main` (tracked as #274 / duplicate #272), **not** from this PR.

It was fixed by **#273**, merged into `main` on 2026-07-15. `Test Rust` and all other checks are now green on this PR with no change here. (#280 was a parallel fix for the same table and is now redundant.)

All 14 tests added by this PR pass — the provenance, upsert-semantics, migration, and version-pin tests are all green on both SQLite and Postgres.

<sub>_Edited to reflect that #273 (not #280) landed the fix and CI is now green._</sub>

smunini (approved):

smunini (approved):

PR #263, opened 2026-07-14
test(persistence): assert every backend fails safely when its store is unreachable

### LINKED ISSUE #212: Extend backend error-handling (unreachable-server) test pattern to PostgreSQL / SQLite / S3
## Summary

Extend the reusable **backend error-handling test pattern** introduced for MongoDB in **#211** to the other standalone primary storage backends (**PostgreSQL**, **SQLite**, and optionally **S3**), so each backend has coverage asserting that its operations fail *gracefully* — surfacing a `StorageError::Backend` — when the store is unreachable or misconfigured, rather than panicking, hanging, or returning a misleading result.

## Motivation

No backend currently has a connection-failure test. The `.map_err(...)` error arms that wrap driver/connection failures into `StorageError::Backend` are therefore entirely unexercised (this is the same class of uncovered lines seen in the settings-store coverage discussion). A small, uniform pattern per backend closes that gap and documents the expected failure contract.

## Template (already implemented for MongoDB in #211)

`crates/persistence/tests/mongodb_tests.rs`:

- `unreachable_backend()` — builds a `MongoBackend` pointed at a dead address (`127.0.0.1:1`), lazily constructed so it never connects.
- `assert_backend_error(result)` — asserts `Err(StorageError::Backend(_))`.
- A test driving `read` / `count` / `create` concurrently and asserting each fails with a backend error.

## Scope / per-backend notes

- [ ] **PostgreSQL** — cleanest analog to Mongo: point the backend at an unreachable host/port with a short connect timeout and assert operations return `StorageError::Backend`. High value.
- [ ] **SQLite** — semantics differ (embedded, file-based; there is no "unreachable server"). The analogous failure is an unopenable DB path (permission denied / missing directory) or a disk I/O error, and it may surface at construction rather than at operation time. Needs adaptation; confirm what failure is worth asserting.
- [ ] **S3** *(optional)* — point at an unreachable endpoint / invalid bucket and assert a backend error. Object-store semantics differ; its recommended role is Archive, so lower priority.
- **Elasticsearch** — out of scope (search-only, never a standalone primary), consistent with how it's treated elsewhere.

## Notes

- Follow-up to the MongoDB test work in **#209** (testcontainer harness) and **#211** (the error-handling pattern). 
- **Contingent on that approach being accepted** — if the reviewers don't want this style of test, close this issue; no work is lost.

mauripunzueta:
Closes #212.

## Summary

Extends the MongoDB error-handling pattern from #211 to **every** storage backend — and fixes the two production bugs the new tests exposed.

The issue scoped this to PostgreSQL / SQLite / S3 and declared Elasticsearch out of scope ("search-only, never a standalone primary"). That turned out to be backwards: **Elasticsearch is the one backend that actually violated the contract**, and it violated it on the search path, in a shipped topology.

## The contract

One new test target, `crates/persistence/tests/backend_error_handling.rs`, states it once:

> An operation against an unreachable or misconfigured store MUST return `Err(StorageError::Backend(_))`. It MUST NOT return `Ok(None)`, `Ok(0)`, or any other success, and it MUST NOT hang.

The `Ok(None)` clause is the point. A `read` that reports "not found" when it never reached the database is indistinguishable from one where the resource is genuinely absent — and "this patient has no recorded allergies" vs "I could not reach the database" are different clinical statements. `count` returning `Ok(0)` is the same lie wearing different clothes.

9 tests, all five backends, **no server and no Docker** — they always run. Each is wrapped in a deadline, so a backend that *hangs* fails loudly rather than stalling CI.

`ResourceStorage` is object-safe, so a single `&dyn ResourceStorage` driver exercises every backend even though their constructors differ wildly:

| Backend | How it's broken | Failure surfaces at |
|---|---|---|
| SQLite | unmigrated store; unopenable path | operation; construction |
| PostgreSQL | refused connection; **blackholed server** | construction (`new()` verifies connectivity) |
| MongoDB | dead address | operation |
| Elasticsearch | dead node, incl. a dedicated `search` test | operation |
| S3 | dead endpoint, lazy and eager | operation; construction |

The MongoDB test **moves here** from `mongodb_tests.rs`. It needs no server, so it did not belong in a suite whose tests all skip without one, and its `mongodb_integration_*` name implied a Docker dependency it never had.

## Bug 1 — Elasticsearch silently reported an unreachable cluster as an empty store

Four catch-all arms:

- `storage.rs` `read`: `Err(_) => return Ok(None)` — **any** transport error became "resource not found"
- `search_impl.rs` `search`: a connection failure returned `SearchAttempt::EmptyIndex` → an empty result set
- `count` / `search_count`: `_ => Ok(0)`
- `create_or_update`: `_ => ("1", true)` — a failed existence check meant "new resource, version 1", silently clobbering history

This was **live**. `composite/storage.rs` *prefers* the Search backend for search, and four of the eight supported `HFS_STORAGE_BACKEND` modes are `*-elasticsearch`. So in a `postgres-elasticsearch` deployment with ES down, `GET /Patient?identifier=X` returned **HTTP 200 with an empty Bundle** — and `update_health(id, result.is_ok())` recorded the dead cluster as *healthy*.

Transport failures now flow through the **existing** retry machinery (so a rolling restart or a not-yet-ready cluster is still absorbed) and surface as `BackendError::Unavailable` once retries are exhausted. The genuinely-empty cases — `index_not_found_exception`, and doc-not-found → 404 — still return empty, so healthy behaviour is unchanged and no existing ES test changes.

⚠️ **Behaviour change worth a reviewer's attention:** a search against a down cluster now errors instead of returning an empty 200. The old code did this deliberately ("so searches don't 500 against a backend that isn't ready yet"). Trading a wrong answer for a loud one is the right call for clinical data, but it is a real change and I'd rather surface it than smuggle it in.

## Bug 2 — PostgreSQL `connect_timeout_secs` was dead config

Declared, defaulted, serde-serialized… and **never read**. `create_pool` never touched it.

Wiring it alone does **not** fix the hang: tokio-postgres applies `connect_timeout` only to `TcpStream::connect`, leaving DNS, the TLS handshake and the Postgres startup/auth exchange awaited **unbounded**. A peer that completes the TCP handshake and then stalls — a hung server, or an LB/security group accepting into a blackhole — hung the caller forever.

So this wires `cfg.connect_timeout` **and** deadpool's `create_timeout`, which bounds the whole of connection establishment. `wait`/`recycle` are deliberately left unbounded: they govern behaviour under load (queueing for a free slot), not reachability, and capping them would change how a saturated pool sheds traffic.

`postgres_backend::blackholed_server_is_bounded_and_surfaces_backend_error` is the regression test — it binds a listener and never accepts, so the kernel completes the handshake and the client waits. **It hangs on `main`.**

## Not in this PR (real findings, filed separately)

- Postgres `SET statement_timeout` is applied to exactly **one** pooled connection, then returned to the pool — every other connection gets the server default, so `statement_timeout_ms` is effectively unenforced. Fixing it via a deadpool `post_create` hook would newly enforce a 30s cap on queries that have never been capped, which could break long-running bulk-export work. Deserves its own PR and its own load validation.
- `PostgresConfig::schema_name` is a third dead field.
- `crates/rest/src/error.rs` maps every `BackendError` except `UnsupportedCapability` to a **500**, and `RestError` has no `ServiceUnavailable` variant — so "the database is down" is indistinguishable from "the server has a bug". No 503, no `Retry-After`, and the readiness probe never consults storage.
- S3 collapses `NoSuchBucket` into the same `NotFound` that `get_object` swallows into `Ok(None)`, so a typo'd or permission-revoked bucket reads as an *empty store*.

## Testing

`cargo test -p helios-persistence --all-features --test backend_error_handling` — needs no Docker.

Note: I was unable to compile locally (no C linker in my environment), so CI is the first real check on this. Every API used was verified against the vendored crate sources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,7 +307,8 @@ compressed when the client sends `Accept-Encoding`.
 | `HFS_MONGODB_URL` / `HFS_MONGODB_URI` | *(none)* | MongoDB connection string |
 | `HFS_MONGODB_DATABASE` | `helios` | Database name |
 | `HFS_MONGODB_MAX_CONNECTIONS` | `10` | Connection pool size |
-| `HFS_MONGODB_CONNECT_TIMEOUT_MS` | `5000` | Connection timeout (ms) |
+| `HFS_MONGODB_CONNECT_TIMEOUT_MS` | `5000` | TCP handshake timeout (ms) |
+| `HFS_MONGODB_SERVER_SELECTION_TIMEOUT_MS` | `15000` | How long an operation waits for a usable server before failing (ms). This, not the connect timeout, bounds how quickly an unreachable MongoDB surfaces an error. |
 
 **S3**
 
--- a/crates/hfs/src/main.rs
+++ b/crates/hfs/src/main.rs
@@ -152,12 +152,19 @@ where
     let connect_timeout_ms = env("HFS_MONGODB_CONNECT_TIMEOUT_MS")
         .and_then(|value| value.parse::<u64>().ok())
         .unwrap_or(5000);
+    // Bounds how long an operation waits for a usable server. `connect_timeout_ms`
+    // only bounds a TCP handshake, so this is what actually decides how quickly an
+    // unreachable MongoDB surfaces an error.
+    let server_selection_timeout_ms = env("HFS_MONGODB_SERVER_SELECTION_TIMEOUT_MS")
+        .and_then(|value| value.parse::<u64>().ok())
+        .unwrap_or(15_000);
 
     MongoBackendConfig {
         connection_string,
         database_name,
         max_connections,
         connect_timeout_ms,
+        server_selection_timeout_ms,
         fhir_version: config.default_fhir_version,
         data_dir: config.data_dir.clone(),
         search_offloaded,
@@ -368,12 +375,17 @@ async fn create_audit_mongodb_storage(
             .ok()
             .and_then(|v| v.parse::<u64>().ok())
             .unwrap_or(5000);
+        let server_selection_timeout_ms = std::env::var("HFS_MONGODB_SERVER_SELECTION_TIMEOUT_MS")
+            .ok()
+            .and_then(|v| v.parse::<u64>().ok())
+            .unwrap_or(15_000);
 
         let config = MongoBackendConfig {
             connection_string,
             database_name,
             max_connections,
             connect_timeout_ms,
+            server_selection_timeout_ms,
             fhir_version: server_config.default_fhir_version,
             data_dir: server_config.data_dir.clone(),
             search_offloaded: false,
--- a/crates/persistence/README.md
+++ b/crates/persistence/README.md
@@ -604,7 +604,10 @@ MongoDB runtime configuration also supports:
 - `HFS_MONGODB_URL` or `HFS_MONGODB_URI` as preferred connection-string inputs
 - `HFS_MONGODB_DATABASE` to select the database name (default: `helios`)
 - `HFS_MONGODB_MAX_CONNECTIONS` to control the driver pool size (default: `10`)
-- `HFS_MONGODB_CONNECT_TIMEOUT_MS` to control the connection timeout (default: `5000`)
+- `HFS_MONGODB_CONNECT_TIMEOUT_MS` to control the TCP handshake timeout (default: `5000`)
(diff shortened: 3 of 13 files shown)


mauripunzueta:
Superseded by a fresh PR from the same branch — GitHub Actions stopped creating check-suites for this PR (no Actions check-suite was generated for the last two pushes, while other PRs in the repo continued to get runs). Reopening as a new PR to get CI to attach. Same branch, same commits.

PR #232, opened 2026-07-09
feat(validator): FHIR resource validation — schema engine, $validate, write-path enforcement

dougc95:
## Summary

Adds FHIR resource validation to HFS: a new **`helios-fhir-validator`** crate (a runtime FHIR Schema engine), the **`$validate`** operation, and optional **write-path enforcement** driven by `HFS_VALIDATION_*` config — across **R4 / R4B / R5 / R6**.

70 files changed, ~8.9k insertions.

## Relationship to the Validation strategy discussion

This work is a response to the architecture strategy document in **[discussions/215 — "Validation"](https://github.com/HeliosSoftware/hfs/discussions/215)**. It delivers most of the *functional scope* that discussion commits to, but takes a **deliberately different implementation path**, so it's offered for review/discussion rather than as a literal implementation of that doc.

**Where this PR agrees with #215:**
- **FHIR Schema** as the internal validation IR.
- An **SD → FHIRSchema converter** validated against the external FHIR Schema conformance suite (the doc's highest-risk Stage 1, proven in isolation).
- **Invariants delegated to `helios-fhirpath`** (not re-implemented) — exactly as the doc argues for.
- **`OperationOutcome`** as the public contract; **per-version** R4/R4B/R5/R6 dispatch.
- The **`$validate` operation** + **write-path validation config** surface.

**Where this PR diverges from #215 (worth a maintainer decision):**

| #215 proposes | This PR does |
|---|---|
| Compile SD→FHIRSchema→**`&'static` Rust tables** via extending `helios-fhir-gen` + a new `#[derive(FhirValidate)]` macro in `helios-fhir-macro` | A standalone **`helios-fhir-validator`** crate with a **runtime** engine that loads **embedded compressed schema packs** and interprets them (mirrors the `fhir-schema` reference impl) |
| Crate named `helios-validation` | Crate named `helios-fhir-validator` |
| Local-first bundled terminology tables + **in-process `helios-hts`** `$validate-code` fallback | Strength-aware binding checks + an **optional remote** `$validate-code` `TerminologyProvider` (no bundled local tables, no HTS wiring yet) |
| Checked-in reviewable `.fhirschema.json` artifacts per version | Schema packs shipped as compressed `*.json.gz` blobs (regenerated by the `generate-schema-packs` bin) |

## Aspect coverage (against #215's aspects table)

| Aspect | Status |
|---|---|
| Structure / Cardinality / Value domains (primitives) | ✅ implemented |
[...]

PR #265, opened 2026-07-14
docs(ui): research & design for a schema-driven FHIR resource editor (#264)

### LINKED ISSUE #264: Research & design: schema-driven FHIR Resource Editor prototype (FHIR Schema + extension editing)
## Goal

**Research and design** (no implementation yet) a **FHIR Resource Editor** prototype for the HFS web UI: a schema-aware editor that lets a user open a FHIR resource, understand its structure, add/remove nodes of information, and edit extensions — with the schema itself driving what is offerable and what is valid.

The deliverable of this issue is a **design document + prototype plan**, not production code. A throwaway spike is fine (and encouraged) to de-risk the hard parts, but the output we want is a written design we can review before committing engineering time.

## Why now

Two pieces have landed (or are landing) that make this feasible for the first time:

1. **FHIR Schema capability — PR #232 (`feat/fhir-validator`, still draft).** It introduces `helios-fhir-validator`, a runtime FHIR Schema engine with an in-memory IR (`FhirSchema`) built from StructureDefinitions and shipped as embedded per-version schema packs (R4/R4B/R5/R6). **This issue should target that engine as its schema source — not a hand-rolled metadata layer, and not the typed `helios-fhir` structs.**
2. **`crates/ui`** — the Askama + htmx, server-rendered UI crate mounted at `/ui` by the `hfs` binary, with embedded assets and no runtime CDN dependency.

## What the FHIR Schema layer already gives us (from PR #232)

`helios_fhir_validator::schema::FhirSchema` carries essentially the metadata an editor needs to render a form and offer the right "add node" choices:

| Field | Editor use |
|---|---|
| `elements` (ordered `IndexMap`) | The set of child nodes that can be added, in spec order |
| `type_`, `kind`, `base` | Which widget to render; how to resolve nested complex types |
| `min` / `max`, `array`, `required` | Cardinality: required markers, "add another" affordance, when to stop offering |
| `choices` / `choice_of` | `value[x]` — offer a type picker, then bind the concrete element |
| `binding` (`value_set`, `strength`) | Coded fields: dropdown/autocomplete backed by `$expand` against HTS |
| `fixed` / `pattern` | Prefill and lock values |
| `constraints` (FHIRPath + `human`) | Inline invariant messages |
| `refers` | Reference pickers scoped to allowed target types |
| `slicing` / `slices` (incl. `Match`) | Profile-driven slices, and the basis for extension slices |
| `extensions` | Named extension slices declared by a profile |
| `must_support`, `modifier`, `summary` | Emphasis / ordering / progressive disclosure in the form |
| `regex` (primitives) | Client- and server-side field-level checks |

So the research is mostly about **how to project this IR into an editable UI**, not about acquiring the schema.

## Research questions to answer

**1. Schema access surface.** The validator's IR is internal to a Rust crate. Does the editor consume it in-process (UI crate depends on `helios-fhir-validator`, renders forms server-side) or over an HTTP surface (a `/ui`-internal or public endpoint that projects `FhirSchema` into an editor-friendly JSON)? Consider that PR #232 keeps schemas as compressed packs and resolves cooperatively rather than by snapshot flattening — walking it per-node for a form render needs a look at cost and at the resolver API (`SchemaResolver`, `SchemaRegistry`, `CompositeResolver`).

**2. Editing model.** Tree editor over raw JSON, or a generated form? Likely both: a structured form view for the common path plus a raw-JSON/source view, kept in sync. Decide what the canonical in-flight representation is (raw `serde_json::Value` is what the validator walks, which argues for it).

**3. Adding nodes.** The core requirement. Given a cursor position in the resource, compute the set of addable children from the schema (respecting cardinality already consumed, `excluded`, choice types, slices), and render an "add element" affordance. Design how this is surfaced (typeahead over element names? a filtered tree of the type?) and how deeply nested complex types are lazily expanded.

**4. Extension editing (explicit requirement).** Cover both cases:
   - **Known/profiled extensions** — declared via the `extensions` sugar / extension slices on a profile; we know the URL, the value type, and the binding, so we can render them like any other field.
   - **Ad-hoc extensions** — the user supplies an extension URL. Design URL resolution: from stored `StructureDefinition`s in the tenant, from the schema packs, or unresolvable (fall back to a manual `value[x]` type picker). Also cover nested/complex extensions (`extension.extension`) and `modifierExtension`.
   - Note the limitation flagged in PR #232's own backlog: core extension definitions aren't in the vendored spec bundles today, so profile `extensions` sugar referencing core extension URLs currently reports `unknown-schema`. The design should say how the editor behaves in that gap.

**5. Validation feedback loop.** How and when the editor validates: live per-field (regex, cardinality, required) versus a full pass through `$validate` (added in PR #232, returns `OperationOutcome`). Map `OperationOutcome.issue.expression` / location back onto editor nodes so errors land on the right field. Note that FHIRPath constraints and terminology bindings are *deferred* (async) in the engine — the sync structural pass is cheap, the effects pass is not; that shapes the interaction design.

**6. Terminology integration.** For `binding` fields, wire code pickers to `$expand` / `$validate-code` (HTS). Decide behavior when terminology is unavailable or the value set is huge.

**7. Profiles.** Editing against `meta.profile` (and tenant/stored profiles) versus base resource type: does the user pick a profile up front, and does the form then narrow (required slices, must-support emphasis, fixed values)?

**8. Technology fit.** The UI crate today is server-rendered Askama + htmx with embedded assets and no runtime CDN. A resource editor is the most interaction-heavy thing we'd have built. Assess honestly whether htmx-driven fragments carry it, whether it needs a scoped island of client-side JS, and what that costs us against the crate's existing no-CDN / air-gapped stance. Recommend one, with the tradeoff written down.

**9. Prior art.** Survey what exists — Health Samurai's FHIR Schema-driven form work, Simplifier/Forge, the HAPI resource editor, LHC-Forms/SDC Questionnaire-based rendering, `fhir-schema`-based editors — and say what we should and shouldn't copy.

## Scope for the prototype (what "enough capability" means)

The prototype should be able to demonstrate, end to end, on at least one FHIR version:

- Load an existing resource from HFS by type + id (and start a new empty one).
- Render a schema-driven editable tree/form of it, ordered and typed by the schema.
- **Add a new node** at any valid point, with the set of offerable nodes derived from `FhirSchema` (including a `value[x]` choice-type pick and a repeating-element "add another").
- Remove a node.
- **Add, edit, and remove extensions**, including at least one profiled extension and one ad-hoc URL extension.
- Show validation results inline, anchored to the offending node, via the PR #232 `$validate` path.
- Save back to HFS (`PUT`/`POST`), surfacing an `OperationOutcome` on rejection.

Explicitly **out of scope** for the prototype: full profile authoring, Questionnaire/SDC rendering, bulk/multi-resource editing, offline conflict resolution.

## Deliverables

1. A design doc (in `docs/`, e.g. `docs/resource-editor-design.md`) answering the research questions above, with a recommended architecture and the rejected alternatives noted.
2. A component/data-flow sketch: where schema projection happens, what crosses the wire, what the editor's in-flight state is.
3. A staged implementation plan with sizing, and a call on the technology question (#8).
4. Any spike code kept on a branch, clearly labeled as a spike.

## Dependencies / coordination

- **Blocked-ish on PR #232** — it's a draft and its architecture is still under maintainer discussion (see [discussion #215](https://github.com/HeliosSoftware/hfs/discussions/215); PR #232 deliberately diverges from it with a runtime engine instead of the proposed `helios-fhir-gen` + `#[derive(FhirValidate)]` codegen path). The editor design should be robust to that decision going either way — it needs *a* FHIR Schema IR, and should not hard-couple to the runtime-pack specifics if the codegen path wins.
- Related UI work: #237 (Compartment Editor UI), #238 (SearchParameter viewer & editor), #234 (saved queries), #236 (history diff rendering). Look for shared editor/form primitives rather than a one-off.

## Not now

Per the request: **record and design only — do not build the production editor yet.**

angela-helios:
Closes #264.

Design proposal and a spike. **No production code.**

## The finding that should shape the decision

A validator and an editor want different things from a schema.

A **validator** asks *"here is a document — is it legal?"* That can be answered by code generated ahead of time.

An **editor** asks *"here is a cursor — what may I put here?"* There is no document to check. It is a query **against the schema itself**, at runtime, at every node the user touches.

Discussion #215 bets that we *"will not walk StructureDefinition snapshots at runtime — we will compile them, ahead of time, into Rust."* That bet is sound for validation and it has a consequence nobody had written down:

> **A compiled validator cannot drive an editor.** `#[derive(FhirValidate)]` gives you `validate()`. It cannot tell you *"the 22 elements addable at this cursor, in spec order, minus the ones whose cardinality is already spent, plus the 50 `value[x]` arms because the cursor is inside an extension."*

This is not an argument against codegen — it is a **requirement on it**: whichever way #215 lands, the FHIR Schema IR must survive as **runtime-queryable data**. Today's packs satisfy that. A codegen path satisfies it only if it *also* emits the IR. If it doesn't, this editor cannot be built — and neither can #237 or #238, which are leaning the same way.

**That is the one thing I most want a decision on.**

## Is the IR actually enough? Yes — and I measured it rather than assuming

Everything numeric in the doc came from running against the **real R4 pack** on `feat/fhir-validator`. The spike (`docs/spikes/resource-editor-addable.js`) implements the "what can I add here?" projection and runs it on a Patient shaped like Steve's Donald Duck example:

| Cursor | Addable nodes |
|---|---|
| root of a `Patient` | **22**, in spec order, with bindings attached |
| inside `Patient.name[0]` | **8** — *including `extension`*, the exact case Steve called out |
| inside a new, empty extension | `url`, `extension`, and `value[x]` → **a pick from 50 types** |

And the two behaviours that make it an editor rather than a form fall out for free: **spent cardinality is excluded** (`gender` is set and max=1 → not offered again; `name` repeats → offered as "add another"), and **extensions nest**, because `Extension.extension` is recursive in the IR — Steve's *"extensions of extensions, which is crazy"* needs no special case.

**The schema layer is sufficient. The editor is a projection problem, not an acquisition problem.**

## Two prerequisites on #232 — both small, neither optional

1. **There are zero concrete extension definitions in the packs.** The generator reads three bundles and `extension-definitions.json` is not among them, so `resolve("…/patient-birthPlace")` returns `None`. The editor can edit extensions *blind* today (the base `Extension` type gives url + 50 `value[x]` + recursion), but it cannot **offer** a known one. This is already on #232's backlog — it's why 11 pack profiles report `unknown-schema`.

2. **The IR has no words in it.** No `short`, no `definition`, no `comment`; `mustSupport`/`modifier`/`summary` exist as struct fields but the converter never populates them (0 occurrences across the whole R4 pack). **An editor is mostly words** — field labels, help text, the US Core red-S emphasis. None of it is there, and no amount of UI work conjures it.

Both are changes to the *generator*, and both help validation as much as they help the editor.

## The prior art reframes how ambitious we should be

I went looking for the tool we're about to build. **It does not exist.**

- **Health Samurai authored FHIR Schema** — the IR we're building on — and built their own forms product on **SDC**, not on their schema.
- **Firely**, the most mature vendor in FHIR, ships a tree editor for *definitions* and a **plain text editor** for instances.
- **HAPI** has no schema-driven resource editor at all; its answer is `$questionnaire`.
- **Medplum** is the one real attempt, and its form **skips any `extension` element with no slices** — so on a resource with no `meta.profile`, *you cannot add an extension at all*. That's most resources.

So the state of the art is a **fork**: either semantic help *without* freedom (Medplum, SDC), or total freedom *without* help (raw JSON in Simplifier/Aidbox/HAPI). **The union is unbuilt in open source** — and the spike shows our IR already carries the freedom half. That union is the differentiating feature, and §13.4 asks for an explicit decision on whether to aim for it rather than drift into it.

## Recommendations, briefly

- **In-process schema access.** The resolver is a one-method sync trait — a hashmap lookup and an `Arc` clone — and FHIRPath is an *optional* feature, so `helios-ui` can take the crate without pulling the FHIRPath engine. An HTTP schema surface would add a hop and a format to version in order to save a call that costs less than the logging around it.
- **`serde_json::Value` as the canonical in-flight state.** This is not a detail: it makes the editor **immune by construction** to the primitive-extension data-loss bug (`birthDate` + `_birthDate`, with positional null-padding in arrays) that *every* tool in the survey has — because what we don't render, we also don't destroy.
- **Whole-document round-trip over htmx**, with a ~200-line vanilla-JS island. Resources are small (a Patient is 4.8 KB of schema, a few KB of data) and `validate_sync` is ~300 µs and pure, so the server can afford to re-render and re-validate on every structural mutation. No SPA, no build step, no CDN — the crate's stance survives. I've written down where this breaks (large Bundles) and staged a task to **find the ceiling by measuring it**, not by guessing.

## Also in the doc

The risk register of what breaks everyone else (slicing round-trip; `modifierExtension` being a safety issue no editor surfaces; validating a half-finished resource — nobody distinguishes *"not yet entered"* from *"wrong"*, and we cheaply can), a staged plan with the dependencies honest, and an appendix listing exactly what was measured versus assumed.

Reviewers: the four decisions are in §13.


angela-helios:
## We stopped citing Simplifier and went and used it

Steve's note was *"go shopping… play with this thing."* So Angela created an account, made an R4 project, and we drove the tool with a Patient built to hit the hard case on purpose: `patient-birthPlace` at the root, and `humanname-own-name` **hanging off `name`** — the "extension on a name" that was flagged as the crazy one.

**The `Update` menu offers exactly three things**, and Firely labels the only editing one themselves:

> `Upload: upload a new version`
> `FHIR Read: fetch a new version from a FHIR server`
> **`Edit: use online text editor`**

That is the entire instance-editing surface, and it opens a **JSON text editor with syntax highlighting and line numbers**. To add an extension, you type the JSON block by hand. This is no longer a conclusion drawn from documentation — it is the tool describing itself.

Four more things we only learned by using it, now in §9 of the doc:

- **Their viewer renders a tree** — so they can clearly do it — **they just don't let you edit there.** And the tree shows extensions **raw**: `url: …/patient-birthPlace`, `value → city: Duckburg`. It never resolves the definition, never says "Birth Place", never says the value is an `Address`. *Even their viewer has no semantic help for extensions* — the exact gap this design targets.
- Changing `"gender": "male"` to `"gender": "masculino"` — a **required**-binding violation — **saved silently.** "Your file was saved."
- **Validation is a separate tool at a separate URL** that you go and run; issues come back as a list in a side panel, with the offending line unmarked in the code pane.
- Instances are tagged **`type: Example of Patient`**. In Simplifier's model, a resource is an *illustration for a profile*, not a first-class object.

### Credit where it's due — two things we should take

1. Their validator anchors an issue as **`At Patient.gender, line 18, position 24`** — a FHIRPath *and* a source position. That is a **better anchor than any of the prior art manages** (Medplum has to *guess* the field by fuzzy-matching FHIRPath strings), and it is the same anchoring `validate_sync` hands us for free. It confirms the shape is right.
2. Their paid editor's **keystroke-level validation feedback**. We can beat it: theirs is a server round-trip, ours is a ~300 µs pure function call.

### And one to avoid

The error rendered as `Exception: CodedValidationException: Value 'masculino' is not a correct code for valueset 'AdministrativeGender'`. Showing a **.NET exception class name** to a clinical user is a message written for the programmer. Our `OperationOutcome` mapping should be held to a higher bar than that.

### Limits of this evidence, stated plainly

We exercised the **free tier**. Simplifier advertises a paid *"full online editor"* we did not use — Firely's own launch post describes it as *"XML and JSON resources (with code highlighting and live validation) and FHIR Shorthand files"*, i.e. a better **code** editor, not a schema-driven form. The conclusion holds, but that part rests on their description rather than our hands, and the doc says so.

PR #269, opened 2026-07-14
feat(cluster): make HFS cluster-capable behind a load balancer

### LINKED ISSUE #169: Database-backed SQL-on-FHIR export job state

### LINKED ISSUE #170: Cluster-aware Subscription notification delivery

aacruzgon:
# Cluster-capable state — Phases 0–4

Long-running draft PR for the cluster-capable-state effort (discussion #223): every push gets the full suite; phases get marked ready / peeled out when merge-worthy. **All four phases are complete and their gates are met** — with `HFS_CLUSTER=true` (or the individual selectors), N `hfs` instances behind a load balancer now share async-job state: SQL-on-FHIR `$export` and `$reindex` jobs can be submitted on one instance and polled, cancelled, executed, and downloaded from any other, a hard-killed worker's jobs are recovered by the survivors, JWKS refreshes are cluster-wide single-flight — one IdP fetch per boot herd or key rotation instead of N — and, as of Phase 3, **topic-based Subscriptions are fully cluster-capable**: a Subscription created on any instance fires for matching writes served by every instance, a WebSocket client bound to instance B receives notifications for writes served by instance A, binding tokens mint/redeem across instances, `eventNumber` is cluster-wide monotonic, and push-channel deliveries run on a durable leased outbox that survives restarts. As of Phase 4, the last two process-local seams from the original inventory are closed too: **HTS terminology caches self-invalidate cluster-wide** on an opt-in switch, N cold-starting HTS instances no longer redundantly race the same terminology import, and **composite secondary-backend (e.g. Elasticsearch) sync is a crash-durable leased outbox** instead of an in-memory buffer.

Design background and status: discussion [#223](https://github.com/HeliosSoftware/hfs/discussions/223). Operator guide: book chapter "Running HFS in a Cluster" ([`book/src/ch15-cluster-deployment.md`](book/src/ch15-cluster-deployment.md)).

# Why

Discussion #223 inventoried the process-local state that breaks when `hfs` runs as multiple instances. The worst offenders shipped here:

- **A1 (#169)**: SoF `$viewdefinition-export` / `$sqlquery-export` job state lived in an in-process DashMap — poll/cancel/download landing on a different instance than the kickoff 404'd.
- **A2**: `$reindex` job state lived in an in-process map with the same failure mode (the handler docs carried the caveat explicitly).
- **F5**: Postgres `update`/`delete` computed `version_id + 1` from a prior SELECT and wrote it back unguarded — two racing writers both assigned N+1 (lost update + a 500 on the history PK). A live bug even single-instance.
- **D3** (found by this PR's own smoke harness): concurrent cold-start schema init raced `CREATE TABLE IF NOT EXISTS` at the catalog level.
- **C2**: every instance holds a per-process `JwksCache`, so N instances all hammer the IdP's JWKS endpoint on boot and on every key rotation. Functionally correct (public keys) — a thundering-herd cost, not a correctness bug, which is why C2 is the one warn-level concern rather than a boot refusal.
- **B1–B5 (#170)**: subscription reaction was per-instance, fire-and-forget, in-memory. Nothing loaded Subscription/SubscriptionTopic from the database at boot, so a Subscription created on instance A produced **zero** notifications for a matching write served by B and all state vanished on restart; WebSocket binding tokens minted on one instance couldn't be redeemed on another (every cluster WS bind failed without sticky routing); `eventNumber` duplicated across instances (breaking subscriber gap detection); a dead endpoint's failures scattered across nodes never reached the `off` threshold; and the delivery retry loop lived on a task's stack — a redeploy dropped every pending retry with no record.
- **C3**: HTS's handler- and backend-layer terminology response caches (`$expand`/`$validate-code`/`$lookup`/`$translate`/`$subsumes`) only cleared on the instance that received the write — a CodeSystem/ValueSet/ConceptMap import on instance A left B serving stale answers indefinitely. Rated **HIGH — silent wrong clinical answers**, the worst-blast-radius Phase 4 finding.
- **D1**: HTS's bootstrap terminology import (SNOMED/LOINC/RxNorm/ICD-10-CM) had a per-file ledger to dedupe unchanged files but no lock — N cold-starting instances could all see "not yet imported" and race the same heavy import concurrently.
- **E1**: `CompositeStorage`'s secondary-backend (e.g. Elasticsearch) sync queue was a crash-lossy in-process `mpsc` channel — an exhausted retry or a crash/redeploy silently dropped the propagation, leaving the primary and the secondary search index diverged with no record. Rated **HIGH — silent search divergence**.
- Plus configurations that silently degrade when clustered (SQLite primary, node-local bulk output, file audit sink) — now refused at boot.

# Changes

## Phase 0 — framing & guardrails

- **`HFS_CLUSTER` master switch + fail-fast validator** (`crates/hfs/src/cluster.rs`): under `HFS_CLUSTER=true`, boot refuses SQLite primaries (F1), `local-fs` bulk output (F2), `file` audit (F4), and per-instance job stores/controllers/sinks for SoF (A1) — one named `Configuration error:` line per violation. Pure function over a `ClusterConfigView`, so the whole refusal table is a unit-test table.
- **`HFS_JOB_STORE_BACKEND`** selector (`memory`/`database`; unset resolves to `database` under `HFS_CLUSTER`).
- **T2 cluster harness** (`crates/persistence/tests/common/cluster_harness.rs`): two *freshly constructed* backend handles over one shared container (never a cloned `Arc`), barrier-synchronized racing, and the DoD assertion rows — calibrated green against the already-cluster-safe bulk-export job store before any feature used it.
- **D3 fix**: schema init serialized cluster-wide with a Postgres advisory lock.
- Operator chapter: [`book/src/ch15-cluster-deployment.md`](book/src/ch15-cluster-deployment.md) with the per-subsystem 1-vs-N matrix, updated as each phase ships.

## F5 — transactional version-guarded writes (Postgres)

`update`/`delete` are now a CAS (`WHERE … AND version_id = $expected`) with the version bump and history INSERT in one transaction; zero rows re-selects to disambiguate `NotFound` vs `VersionConflict`, matching the sqlite/mongo trait contract. `create_or_update` absorbs transient CAS losses with a bounded retry, so unconditional PUT stays last-writer-wins at the API surface. Mongo was already a version-guarded CAS — test-only there.

## Unified cluster job store

New `ClusterJobStore` trait (`crates/persistence/src/core/cluster_job_store.rs`) over one `cluster_jobs` table (migration **v14→v15**), discriminated by `JobKind { SofExport, Reindex }` with opaque JSON payload/progress/result. Claiming clones the bulk-export lease discipline — one transaction around `FOR UPDATE SKIP LOCKED` with a fencing-token bump — and every worker-side mutation is fenced by `worker_id` + token (+ `status='running'` for state changes), so a zombie worker gets `LeaseLost` instead of resurrecting a cancelled or reclaimed job. The bulk-export traits themselves are deliberately untouched (resolved decision F3). Server seam mirrors `sof_runner()`: `ResourceStorage::cluster_job_store()`, default `None`; Postgres returns a pool-sharing `PgClusterJobStore`. A `testing::InMemoryClusterJobStore` serves as the T1 reference model and unit-test fake.

## SoF export on the job store — closes #169

- `ExportJobController` → `#[async_trait]`; the execution engine (`run_views_job`/`run_sqlquery_job`) is shared between controllers via a `JobProgress` hook.
- New `DatabaseExportJobController`: `submit` durably enqueues before the 202; polls map `cluster_jobs` rows to `JobStatus`; download URLs re-resolve per poll (S3 re-signing contract); cancel lands cross-instance and reclaims partial output. Per-instance claim/lease workers (background heartbeat, progress snapshots, cancellation observed between work units, orphaned-shard cleanup on fenced-out terminal writes) plus a kind-scoped reaper honoring `HFS_EXPORT_OUTPUT_TTL_SECS`.
- `HFS_EXPORT_CONTROLLER` (previously parsed-but-unread) is finally consumed; unset follows the job-store mode.

## Reindex on the job store (A2)

`ReindexOperation.with_cluster_store(...)`: `start` enqueues, any instance answers `$reindex-status`/cancel, per-instance workers claim and run the untouched `run_reindex` via a bridge task that mirrors progress into store snapshots and feeds cross-instance cancellation (and lease loss) back into the local cancel channel. `get_progress`/`cancel`/`list_jobs` are now tenant-checked in **both** modes (the in-memory map previously served any tenant's job to anyone who knew the UUID).

## JWKS refresh coordination (C2, Phase 2)

- **`ClusterRefreshCache`** (`crates/persistence/src/core/cluster_refresh_cache.rs`): a generic string-keyed single-flight document refresh over a new `cluster_refresh_cache` table (migration **v15→v16**; deliberately server-global — the documents are public IdP key material, no tenant data). Freshness is **watermark-based**: callers pass the `fetched_at` they already hold, and a stored document is reused only when strictly newer (and within a staleness cap and its own `max_age`) — this is what makes a key-rotation race deterministic and never reuses "my own" document. `PgClusterRefreshCache` takes a per-key `pg_advisory_xact_lock` in a transaction and runs the caller's fetch closure under it (bounded by the 10s HTTP timeout; rotation/boot events only); commit/rollback releases the lock, and ages use the DB clock so instance skew is irrelevant. Exposed via `ResourceStorage::cluster_refresh_cache()`, default `None`, mirroring `cluster_job_store()`.
- **`JwksCoordination`** (`crates/auth/src/jwks/coordination.rs`): the trait `JwksCache` consults. `JwksFetcher` is split into `fetch_raw` (preserves the raw JWKS body + `Cache-Control` — what the shared store holds) and `parse_document` (run locally by every instance); `JwksCache` is now `Clone` and its background refresh loop calls `refresh()` through a clone instead of duplicating the fetch/swap inline — one refresh seam. Errors split `Fetch` (IdP failed under the lock — propagate) from `Unavailable` (store down — **fall back to a direct IdP fetch**; auth availability outranks the dedupe optimization). Reused documents install with an age-shortened local TTL.
- **`StoreJwksCoordination`** (`crates/rest/src/jwks_coordination.rs`) bridges the two traits — auth stays DB-free, persistence stays auth-free; rest depends on both.
- **`RedisJwksCoordination`** (`crates/auth/src/jwks/redis_coordination.rs`, behind the restored auth `redis` feature): the pre-#205 coordinator resurrected as an explicit opt-in second implementation (session decision — keep a Redis scaffold for future cluster work). `SET NX EX` lock with a holder-token Lua compare-and-del, stored JSON document with TTL, lock losers poll for the winner's document, dead holders bypassed after the lock TTL; every Redis error maps to `Unavailable`, so a down Redis never takes auth down.
- **Boot wiring** (`crates/hfs`): `HFS_AUTH_JWKS_COORDINATION` = `local` | `database` | `redis`; unset resolves to `database` under `HFS_CLUSTER`. **Warn-only by design**: explicit `local` under cluster logs a warning, never refuses (contrast with the refusal table); backends without a refresh store (SQLite/Mongo/S3 primaries) warn and stay per-instance; invalid values fail fast naming the variable. The database mode's initial fetch is deferred to just after primary-backend construction (`finish_auth_boot`, called in all eight start paths) — boot still fails on a dead IdP, in every mode.

## Subscriptions cluster delivery — closes #170 (Phase 3)

A single-notifier alternative was also considered ([comment on #170](https://github.com/HeliosSoftware/hfs/issues/170#issuecomment-4836739605)): designate one instance as the sole subscription evaluator/deliverer, route WS traffic to it via an LB rule, and collapse the fan-out/token/counter work into an outbox the notifier polls. Not taken — it trades a single point of failure and a throughput ceiling on one node for less infrastructure, and it requires LB routing rules keyed on instance role, which conflicts with this effort's goal of N interchangeable instances behind a plain load balancer (no pinned/promoted node anywhere else in this PR). The fan-out design below keeps every instance a first-class delivery target instead, at the cost of the shared counter/token/fan-out machinery the alternative would have avoided.

### Persistence substrate (migration v16→v17)

Four new `persistence::core` seams, each with a `pub mod testing` in-memory reference model, Postgres implementations, and default-`None` `ResourceStorage` accessors mirroring `cluster_job_store()`:

- **`subscription_state`**: per-subscription counters (`eventNumber`, consecutive failures) as single-statement upsert-increments — rows are created lazily by the first increment, so a Subscription resource *update* can never reset shared counters — plus cluster-visible runtime status and a `subscription_notification_events` table holding built websocket bundles keyed `(tenant, sub, eventNumber)`. Also the boot-hydration read path (`SubscriptionHydrationSource`): cross-tenant and tenant-registry-independent (the registry is opt-in metadata), with R4 backport `Basic` rows pre-filtered in SQL by the fhir-types `SubscriptionTopic` coding so hydration never hauls unrelated `Basic` resources into memory.
- **`subscription_delivery`**: a durable delivery outbox under the cluster-jobs lease + fencing discipline plus a persisted retry schedule (`attempts`, `next_attempt_at`); the claim itself bumps the attempt counter, so retry bookkeeping survives any crash. Deliberately a dedicated table, **not** a new `cluster_jobs` kind — deliveries are high-volume scheduled rows that would pollute the Phase 1 claim index and need none of the job store's progress/cancel surface.
- **`ws_binding_tokens`**: single-use WebSocket binding tokens redeemed with `DELETE … RETURNING` (redeem-once by construction), expiry judged on the database clock.
- **`event_fanout`**: tiny envelopes (`ws-event`/`lifecycle`/`state`/`resync`) broadcast to every instance. `PgNotifyFanout` publishes via pooled `pg_notify` and listens on a **dedicated non-pooled** connection with a capped-backoff reconnect loop that synthesizes a local `Resync` after reconnects; memoized per backend, stopped on drop, with a `ready()` gate so consumers (and tests) can await the LISTEN session. Best-effort by contract — nothing correctness-critical rides NOTIFY.

All schedule/lease SQL runs on the **database clock** (a client-clock comparison of DB-stamped `next_attempt_at` made "due" skew-dependent and flaked once against a container clock — same lesson as the C2 refresh cache). `CompositeStorage` now forwards **all** cluster seams to its primary — including the pre-existing `cluster_job_store`/`cluster_refresh_cache` accessors it had silently dropped since Phases 1–2 (latent gap: `postgres-elasticsearch` modes reported no cluster stores).

### Engine: hydration, shared counters, shared tokens (B2/B3/B4)

`SubscriptionEngine` takes optional `ClusterHandles`; without them every path is byte-for-byte pre-Phase-3 (the full pre-existing suite runs unedited). With them: `eventNumber` mints through the shared store (store failure falls back to local with an error — delivering with a possibly-duplicate number beats dropping the notification); failure streaks and status transitions record in `subscription_state` so `error`/`off` thresholds work no matter which instance failed; `hydrate()` rebuilds the projections from persisted resources at boot through non-handshaking registration helpers (N restarting instances never re-handshake; stored runtime state overlays the resource's status field, while a **live** write keeps the shared `eventNumber` but re-seeds status from the fresh resource — the client's re-request intent wins); binding tokens mint/redeem through the shared store (fail-closed); `$status`/`$events`/WS-bind serve a snapshot overlaid with shared state.

The topic registry is now **tenant-scoped** (`(tenant_id, canonical_url)`) in both modes — previously it was keyed by bare canonical URL, letting tenant B's Subscription resolve tenant A's topic (session decision: fix the leak everywhere, not just in the shared tables).

### Delivery: WS fan-out (B1) + durable outbox (B5)

Cluster mode splits per channel after building each notification:

- **websocket**: the bundle persists to `subscription_notification_events`, local sockets get it inline (lossless local leg), and a `ws-event` envelope fans out — every other instance loads the stored bundle and delivers to *its* sockets; the origin skips its own envelope. Remote delivery is best-effort-with-gap-detection by `eventNumber` (the FHIR WS channel contract).
- **rest-hook / email / message**: fully outbox-driven from attempt zero — enqueue + wake hint; per-instance workers claim one attempt per cycle (`run_next_subscription_delivery` is the deterministic test seam) and record complete / scheduled retry (reusing the existing backoff math, persisted) / terminal failure + shared thresholds. A worker that misses the subscription locally re-hydrates before dispatching, so any instance can deliver any row. **At-least-once** by contract. An enqueue failure falls back to inline dispatch — delivering without durability beats dropping.

Subscription/topic lifecycle changes and status flips publish envelopes; the fan-out listener re-reads the resource from storage and patches local projections (a vanished resource reads as a delete), and answers `Resync`/broadcast-lag with a full re-hydration. `build_app` wires the five seams under `HFS_SUBSCRIPTIONS_FANOUT=pg-notify` (unset resolves to `pg-notify` under `HFS_CLUSTER`); backends without the seams warn and stay per-instance, and the validator refuses the genuinely unsafe combinations: explicit `memory` fanout under cluster, and subscriptions on a non-Postgres primary.

## HTS cache coherency (C3) + bootstrap lock (D1) + composite durable sync outbox (E1) (Phase 4)

**User decision this session**: HTS's new cache-invalidation toggle is a **standalone `HTS_*`-prefixed env var**, not coupled to `HFS_CLUSTER` — HTS can scale independently of the `hfs` FHIR server, and this follows its existing all-`HTS_*` convention.

- **D1 — HTS bootstrap advisory lock**: `bootstrap_sync_postgres` wraps the whole terminology-directory sync in a dedicated `pg_advisory_lock` (`HTS_BOOTSTRAP_LOCK`, distinct from the pre-existing schema-DDL lock), unconditional — no gating needed, advisory locks are near-free when uncontended. The per-file ledger check runs *inside* the lock, so a loser naturally skips whatever the winner already imported.
- **C3 — HTS terminology cache epoch**: a new single-row `terminology_epoch` table + `EpochGuard`, shared between the `AppState`-layer handler caches and the `PostgresTerminologyBackend`-layer response caches (each independently tracking its own last-cleared epoch against one shared memoized DB fetch, so neither layer's clear silently suppresses the other's within the same memo window). Bumped alongside the existing cache-clear calls on every terminology write; checked before every `$expand`/`$lookup`/`$validate-code`/`$translate`/`$subsumes` read. Opt-in via `HTS_TERMINOLOGY_CACHE_INVALIDATION = local | epoch` (default `local`, i.e. unchanged single-instance behavior).
- **E1 — composite sync durable outbox**: a new `CompositeSyncOutbox` trait + Postgres implementation (migration **v18**), denormalized to one row per `(event, backend_id)` pair, covering `Create`/`Update`/`Delete` (bulk resync stays on the existing in-memory path — its all-or-nothing batch semantics don't cleanly denormalize). **Capability-based wiring, not `HFS_CLUSTER`-gated**: `CompositeStorage` wires the outbox unconditionally whenever the primary is Postgres, mirroring F5's unconditional-fix precedent — no new env var. Workers claim with `FOR UPDATE SKIP LOCKED` + fencing, persisted retry/backoff, and a same-process `tokio::sync::Notify` for fast-path wakeup (no cross-instance fan-out needed — crash recovery is governed by lease expiry + a 1s poll floor regardless). A warn-only validator row (`resolve_composite_sync_durability`, mirroring C2's JWKS posture) flags a composite secondary on a non-Postgres primary under cluster mode, rather than refusing boot — that configuration already behaves this way single-instance today, so cluster mode isn't introducing a new regression.

## T3 harness — cluster smoke + nightly kill-9

- `cluster-smoke.yml` (merged to main as the dispatchable skeleton, run with `ref=<branch>`): two `hfs` processes + shared Postgres + nginx round-robin front (`X-Hfs-Upstream` proves distribution). Checks: health A/B/front, round-robin, write-on-A/read-on-B, **SoF export A → poll via front → download via B**, and — new in Phase 3 — **check 5, the strategy's one mandatory two-process test**: a websocat client bound to instance B receives the notification for an Encounter written to instance A, after proving lifecycle propagation (`$status` active via B for a Subscription created on A) and the cross-instance binding-token mint/bind, and closing with the sticky-session negative (the consumed token re-presented to A is rejected — redeem-once across instances), and — new in Phase 4 — an **HTS smoke stage**: two `hts` processes sharing the same Postgres, import via A, pre-warm B's caches with the stale display, update via A, assert B self-corrects (not just a cold read, which would pass even without C3). The smoke build gains the `subscriptions` feature and the pinned websocat install; the nginx WS `Upgrade` plumbing was already in place.
- **Nightly kill-9 recovery (A1)**, gated on `schedule` / `-f nightly=true`: stop B, seed 200k patients, kick off on A, wait for the claim, `kill -9` A, restart B — assert B reclaims the orphaned lease after expiry under a new `worker_id` with a bumped fencing token, completes, and serves the full-count shard. Passed on first dispatch.
- **Nightly kill-9 recovery (E1)**, same gating, dedicated `postgres-elasticsearch` instance pair + Elasticsearch container: stop B, POST a 2000-Patient transaction Bundle via A (durably enqueues one outbox row per resource), poll for a healthy queued/applying backlog, `kill -9` A, restart B, wait for full drain, then verify via a live Elasticsearch search through B that every row landed — not just that the outbox says so. Adapted from A1's shape for E1's many-small-rows-not-one-big-job failure profile. Passed on first real CI dispatch.
- **New `redis-cluster-tests.yml`** (Phase 2): runs the `RUN_REDIS_CLUSTER_TESTS=1`-gated Redis twin suite on PRs touching `crates/auth`, nightly, and on dispatch — the Redis seam gets the identical assertions as the CI-default DB backend without gating unrelated PRs.

# Testing

- **T1**: validator refusal table (7 collected violations) + the C2 resolution table (warn-only rows, feature-gated redis arms), job-store state machine vs the in-memory reference (6), refresh-cache semantics vs its in-memory reference (6 — watermark, staleness, own-max-age, error paths), `JwksCache`×coordination contract tests with a fake coordinator (6 — reuse without IdP contact, watermark threading, fallback, propagation), config selector tests.
- **T2** (two fresh handles, one shared container; deterministic — no sleeps): bulk-export calibration (5), D3 cold-start, F5 races (update/update, PUT/PUT, update/delete + Mongo CAS twin), job-store DoD (visibility/isolation/exclusivity/fencing/durability/cross-instance cancel), two-controller SoF export (`rest/tests/sof_export_cluster_pg.rs`), reindex DoD (4), **refresh-cache DoD (4: raced single-flight = exactly one fetch, cross-handle durability, watermark refetch, error releases the lock)**, **JWKS cluster suite (`rest/tests/jwks_cluster_pg.rs`, wiremock IdP hit counter: boot herd = 1 fetch, late joiner = 0, rotation race = exactly 1 more)**, and the **Redis twin (`auth/tests/jwks_cluster_redis.rs`, identical assertions, `RUN_REDIS_CLUSTER_TESTS=1`)**. The wrong-tenant isolation row rides every suite except the refresh cache, where it is N/A by design (server-global public key material; documented in the migration and test modules).
- **Phase 3 T1**: validator table grew the two subscriptions rows (worst case now collects 9 violations) + subscriptions-on-Postgres acceptance; fanout/enabled config resolution tests; 18 unit tests in the new `testing` reference models; `tests/cluster_engine.rs` — two engines over shared in-memory stores: gap-free counters, scattered failures reaching the `error` threshold on both engines' snapshots, mint-on-A/redeem-once-on-B, non-handshaking hydration with state overlay, WS fan-out across engines with the origin no-double-delivery negative, retryable-failure requeue-then-deliver, lifecycle propagation incl. delete closing the remote instance's socket clients — **plus the asserted-unsafe memory twin** (two engines without handles duplicate eventNumbers: the disease the seams cure).
- **Phase 3 T2** (two fresh backend handles, one shared container): 10-test persistence suite (raced gap-free increments, scattered failure accumulation, mint-on-A/redeem-once-on-B with a zero-TTL expiry row, outbox claim exclusivity / reclaim-after-worker-death / fencing / retry schedule, LISTEN/NOTIFY fan-out across two independent handles — `ready()`-gated, event-driven, no sleeps — cross-tenant registry-independent hydration, wrong-tenant isolation throughout, under `SUBSCRIPTION_OUTBOX_TEST_LOCK` where claims are cross-tenant); 6-test engine-over-Postgres suite (`tests/subscriptions_cluster_pg.rs`, behind a new `postgres` marker feature so it rides `--all-features` CI): the B3 create-on-A-matchable-on-B row, B4 gap-free alternated events asserted on wiremock bodies, counter durability across dropped handles, B2, B5 cross-instance claim incl. the local-miss re-hydration path, and the B1 in-process twin over real LISTEN/NOTIFY.
- **T3**: cluster-smoke **green on first dispatch with check 5** on the Phase 3 head ([run 29459960052](https://github.com/HeliosSoftware/hfs/actions/runs/29459960052)); previously green on the Phase 2 head ([run 29441743065](https://github.com/HeliosSoftware/hfs/actions/runs/29441743065)); nightly kill-9 green ([run 29428583874](https://github.com/HeliosSoftware/hfs/actions/runs/29428583874)). No smoke check for C2 (deliberate — its only observable is the upstream hit count, which the T2 wiremock counter owns).
- **End-to-end boot checks** (local, two binaries + one Postgres): C2 — exactly one IdP fetch across both boots; Phase 3 — a Subscription created and activated via A answers `$status` via B, a binding token minted on A binds a WebSocket on B, an Encounter written to A is delivered to B's socket over LISTEN/NOTIFY, the consumed token re-presented to A closes 1008, and both new refusal messages are exact.
- Full CI green on the Phase 3 head `e9407064` ([run 29459963541](https://github.com/HeliosSoftware/hfs/actions/runs/29459963541), attempt 2 — attempt 1 hit a self-hosted-runner disk-full on `github-agent5`, infrastructure only); Redis Cluster Tests green on the same head; previously green on the Phase 2 head ([run 29441746666](https://github.com/HeliosSoftware/hfs/actions/runs/29441746666)).
- **Phase 4 T1**: `EpochGuard` unit tests (memo-window respected, bump detected after expiry, `enabled=false` a total no-op); `CompositeSyncOutbox` state-machine tests against its `InMemorySyncOutbox` reference model (claim/fence/retry/fail/reap); `resolve_composite_sync_durability` resolution table.
- **Phase 4 T2**: `postgres_bootstrap_lock.rs` (D1 — a read-sleep-write race proving the advisory lock serializes two independently constructed HTS backend handles, verified red-without-fix / green-with-fix); `postgres_epoch_cluster.rs` (C3 — two independent `AppState`/backend pairs over real HTTP prove a pre-warmed stale lookup on B is invalidated by an update via A across both cache layers, plus the `local`-mode unsafe-contract negative); `composite_sync_cluster_suite` in `postgres_tests.rs` (E1 — exclusivity, fencing + durability-after-worker-death, retry-schedule + isolation, and an end-to-end `CompositeStorage` wiring test with a genuinely separate throwaway secondary database proving a write via A durably enqueues and B's worker claims and applies it).
- **Phase 4 T3**: cluster-smoke green on first dispatch with the new HTS C3 stage ([run 29510624382](https://github.com/HeliosSoftware/hfs/actions/runs/29510624382)); full CI green on the Phase 4 head `864b8cad` ([run 29511136310](https://github.com/HeliosSoftware/hfs/actions/runs/29511136310)); nightly kill-9 (E1) green on first real dispatch ([run 29524686368](https://github.com/HeliosSoftware/hfs/actions/runs/29524686368) — the drain-and-recover step completed in 71s with zero failed rows and full Elasticsearch convergence).
- **Infra note (Phase 4)**: `Test Rust` failed twice on **github-agent5** (`Stevens-Mini`) with a macOS dyld shared-cache mapping failure (`SIGABRT` before any test ran) — zero test assertions failed either time, purely an OS-level runner fault (the same runner flagged for disk space during Phase 3). Green on a later retry once the runner recovered externally.

# Notes

- **Migrations**: Postgres schema v14→v15 (`cluster_jobs`), v15→v16 (`cluster_refresh_cache`), v16→v17 (`subscription_state`, `subscription_notification_events`, `subscription_delivery_outbox`, `subscription_ws_tokens`), and v17→v18 (`composite_sync_outbox`). Applied automatically under the existing advisory-locked init. HTS's own schema (no versioned migration runner) gained a `terminology_epoch` table as idempotent DDL, plus the new `HTS_BOOTSTRAP_LOCK` advisory-lock key.
- **Zero change for single-instance defaults**: with `HFS_CLUSTER` unset everything keeps its in-process behavior; the new tables sit idle unless the database modes are selected. One deliberate behavioral fix in both modes: the topic registry is tenant-scoped, closing a pre-existing cross-tenant topic-visibility leak (single-tenant deployments unaffected).
- **Env vars**: `HFS_CLUSTER`, `HFS_JOB_STORE_BACKEND`, `HFS_EXPORT_CONTROLLER` (now read), `HFS_AUTH_JWKS_COORDINATION`, `HFS_AUTH_REDIS_URL` (opt-in Redis coordinator only), `HFS_SUBSCRIPTIONS_FANOUT` (new; `memory`/`pg-notify`, unset follows `HFS_CLUSTER`), and `HFS_SUBSCRIPTIONS_ENABLED` hoisted onto `ServerConfig` with its historical truthiness preserved; plus **`HTS_TERMINOLOGY_CACHE_INVALIDATION`** (new, Phase 4 C3, `local | epoch`, default `local`) — deliberately a standalone `hts`-binary env var, not `HFS_*`/`HFS_CLUSTER`-coupled. E1's composite sync outbox needs no new env var (capability-based). Documented in ch15.
- **Subscriptions semantics under cluster** (documented in ch15): remote WebSocket delivery is best-effort with `eventNumber` gap detection (envelopes published while an instance's LISTEN connection is down are lost; it re-syncs on reconnect); push-channel delivery is at-least-once (endpoints should dedupe by `eventNumber`); a transaction-pooling proxy (pgbouncer in transaction mode) between HFS and Postgres breaks LISTEN/NOTIFY — connect directly or use session pooling. The subscription heartbeat interval remains unwired (must be lease-guarded if it ever lands).
- **Cargo features**: helios-auth re-gains `redis` (redis 0.27, exactly the pre-#205 shape; moka stays gone — it was jti-only); helios-hfs re-gains the pass-through. Neither is in any default set.
- **Upstream absorptions**: main's #233 breakage was fixed upstream by #273 (we closed our duplicate #276); main's **#205 removed the jti subsystem**, so the validator's former C1 check was deleted with it and **Phase 2 rescoped to C2 (JWKS refresh coordination), now landed**. Also resolved: the Redis-vs-DB-only question — the database mode is the default, Redis is an opt-in scaffold.
- **Latent-gap fix**: `CompositeStorage` now forwards the cluster seams (including Phase 1/2's `cluster_job_store`/`cluster_refresh_cache`, which it had silently dropped) to its primary, so `postgres-elasticsearch` modes participate in every cluster feature.
- **Known limits / follow-ups**: the cluster job store, refresh cache, subscriptions, and composite sync outbox seams are all Postgres-primary (Mongo/S3 primaries warn and stay per-instance, per the documented posture); the nightly schedules (kill-9 A1/E1, Redis suite) execute main's copies of the workflows, so they become effective on merge; HTS's cache epoch is Postgres-only by the same reasoning (SQLite can't be shared across a cluster anyway).


mauripunzueta:
@aacruzgon — heads-up on an overlap: draft PR #1109 also edits `crates/persistence/src/search/reindex.rs`. The changes are confined to error reporting and retry policy:

- `ReindexProgressError` gains `retryable`, and `push_error` takes it.
- `ReindexProgress` gains `tenant_id`, set in `start_tracked`.
- `to_parameters` lists per-resource errors.
- `AutomaticGenerationOutcome` gains a `PermanentErrors` variant, and the outcome decision moved into `automatic_outcome()`.

The rebuild path itself (page loop, writers, bulk index rebuild) is unchanged. Whichever PR merges second will likely need a small rebase in that file. Happy to do it on my side if #269 lands first.

PR #277, opened 2026-07-14
feat(ui): schema-driven resource editor — a real Guided form (#264)

### LINKED ISSUE #264: Research & design: schema-driven FHIR Resource Editor prototype (FHIR Schema + extension editing)

angela-helios:
Closes #264. **Prototype**, as the issue asked (design in #265) — built and driven end-to-end in a real browser.

> **Stacked on `feat/fhir-validator` (#232).** The editor consumes that PR's FHIR Schema IR. It cannot merge before #232 does, and it will need a re-target once #232 lands.

## What this is

Brett's "Resources - Edit" frame gave the shell — the **Guided form / JSON** toggle, Save & Delete, the editable tree. What the mock could not do is the thing Steve flagged in his notes:

> *"if you wanted to create an address? People have multiple addresses. **How do you do that? There's no concept here.**"*

He was right about the mock: it is eight hand-picked fields. This is a Guided form that is **generated from the resource's schema**, so it can add a second address, an extension on a name, a `value[x]` the designer never thought of — anything the schema allows, at any depth.

loaded

## The thing no other editor does

An **arbitrary extension, by URL, at an arbitrary node, on a resource with no profile** — with the full `value[x]` type pick. Here is the "add" panel open on `Patient.name[0]` (the "extension on a name" Steve called crazy), offering the name's elements *and* an Extension-URL field:

addable on a name

Type a URL, and the new extension offers all 51 `value[x]` arms:

value x pick

This is the union the whole field is missing (from the #265 survey): **Medplum** needs a profile to touch an extension; **Simplifier, Aidbox, HAPI** send you to raw JSON. We do it from the base `Extension` type, so it needs no profile and works nested.

## How it is built

**The document is the state.** It rides in a hidden field; every structural mutation posts the whole resource back, the server applies it, re-validates, and re-renders. No client model, no session. Three things fall out of that:

- The editor's hard rules — cardinality, choice groups, the `[x]`-artifact and choice-declarer traps — live **once, in Rust**, tested (17 unit tests against the real R4 pack). The projection is in the **validator crate** (`helios_fhir_validator::editor`), not the UI, because #237 and #238 want the same "what can I add here?" query.
- Validation runs on **every mutation** (the pure ~300 µs sync pass) and errors come back **anchored to their node** by an exact dotted path — no fuzzy FHIRPath string-matching:

error anchored

- **Whatever we do not render, we do not destroy.** The `_birthDate` sibling of an extended primitive round-trips here — the quiet data-loss bug every editor surveyed for #265 has. There is an HTTP test that asserts exactly this.

## Two safety touches no surveyed editor has

- A **`modifierExtension` is flagged** — an unrecognised one means *"do not process this resource"*, and no editor surveyed says so.
- A **key the schema does not know is shown and kept**, marked "not in schema", rather than silently dropped.

## What is deliberately not here (prototype scope, per the issue)

- **Profiled extensions** wait on #232's extension catalogue (`extension-definitions.json` is not in the pack yet — see #265 §5). The **ad-hoc** path works today; the profiled one lights up when that lands.
- **No `mustSupport` emphasis or field labels** — the IR carries neither (#265 §8). Field names are shown as-is.
- **Slicing** beyond plain arrays, and the terminology-backed pickers for bound fields, are staged, not built.
- Live terminology (a bad *code*) is the async effects pass, so it is checked on save, not per-keystroke — by design (#265 §6).

## Verification

`cargo fmt`, `clippy` clean. 17 projection unit tests + 7 HTTP round-trip tests + locale parity (en/es/de). Driven in headless Edge: loads Donald Duck, adds an ad-hoc extension to his name, picks a `value[x]`, anchors a structural error — **no JS errors**.


angela-helios:
## Update: JSON + guided form in parallel, and the JSON is now foldable

Per the feedback that Brett's Resource Editor shows the JSON and the guided form **side by side** (not a toggle), with **line numbers** and **expand/collapse arrows** on objects — the editor now does exactly that.

parallel

- **JSON and the guided form sit side by side** over one document, with a **version rail** to the right — three panels on one screen, matching Brett's frame. The old `Guided form / JSON` toggle is gone.
- The **JSON is a real code view**: line numbers in the gutter, a **fold arrow on every object and array**, and syntax highlighting (keys / strings / numbers / bools). A textarea can do none of that, so it is rendered server-side (`json_view` module, 6 unit tests) into line records a few lines of JS fold. **Collapse all / Expand all** in the header; **Edit raw** is one toggle away and re-renders the form on apply.
- **Collapse all** folds each container to its `{ … }` / `[ N ]` summary:

collapsed

- The **version rail** is hydrated from the ordinary `_history` API (etag / lastModified, `resource.meta` fallback); clicking a version loads it into the editor.

Verified end to end in a real browser: 26 JSON lines with numbers, 8 fold arrows, folding hides the right lines and shows the summary, collapse-all works, no JS errors. An HTTP test asserts the parallel line-numbered foldable render; locale parity across en/es/de. `fmt`/`clippy` clean.

angela-helios:
Stacked per #321: base is now #324 (`feat/fhir-validator`), so this PR reviews as the editor layer only. The #257 (nl-search) files still appear in the diff until that PR merges to main and is folded into the base.

Review order: #257 ? #324 ? this PR.

smunini (approved):

PR #280, opened 2026-07-15
fix(hts): create the concepts_search_fts FTS5 table it uses (#274)

### LINKED ISSUE #274: main is broken: `concepts_search_fts` FTS5 table is used but never created (semantic merge conflict between #194 and #233)
## Summary

`main` currently fails `Test Rust`. `crates/hts/tests/value_set_ops.rs::expand_is_a_filter_with_text_filter_still_nests` errors with:

```
no such table: concepts_search_fts
```

The `concepts_search_fts` FTS5 table is **read from and written to** in `crates/hts/src/backends/sqlite/value_set.rs`, but its `CREATE VIRTUAL TABLE` **does not exist anywhere in the repository**:

```
$ grep -rn "concepts_search_fts" --include=*.rs .
crates/hts/src/backends/sqlite/value_set.rs   (10 usages: SELECT / MATCH / bm25 / DELETE / INSERT)

$ grep -rn "CREATE VIRTUAL TABLE.*concepts_search_fts" .
(no matches)
```

The sibling tables (`concepts_fts`, `concepts_word_fts`, `implicit_expansion_fts`, …) are all declared in `crates/hts/src/backends/sqlite/schema.rs`. This one was never added.

`populate_concepts_search_fts_for_system()` propagates the error with `?`, so the first text-filtered expansion for any system fails hard and surfaces as an `OperationOutcome` exception.

## Why CI did not catch it — semantic merge conflict

Neither PR could have caught this alone; it only breaks in combination.

| PR | what it did | why it was green |
|----|-------------|------------------|
| **#194** (`feat/hts-ig-publisher-validate-code`, merged `4df65385b`) | Added the code that uses `concepts_search_fts`, **without** the DDL | Nothing in its test set reached the build path |
| **#233** (`fix/227-hts-expansion-collapse`, merged `e5320c5f2`) | Added `expand_is_a_filter_with_text_filter_still_nests`, which **does** reach it | Its CI (run `29021783522`) ran against a base that predated #194 |

Both merged green. The combination on `main` is broken.

## Suggested fix

Add the missing table to `crates/hts/src/backends/sqlite/schema.rs`, mirroring `concepts_fts`. Inferred from usage (`INSERT INTO concepts_search_fts(rowid, system_id, code, term)`, `MATCH` on the term, `bm25()` ranking, `system_id` used only as a post-filter):

```sql
CREATE VIRTUAL TABLE IF NOT EXISTS concepts_search_fts
USING fts5(system_id UNINDEXED, code UNINDEXED, term,
           tokenize='trigram case_sensitive 0');
```

Flagging rather than fixing: the tokenizer and which columns are `UNINDEXED` are **my inference**, not the author's intent — @sandhums should confirm the shape, since `term` carries both preferred displays and synonym designations.

Also worth considering: the startup path clears/prebuilds the other FTS tables; `concepts_search_fts` likely needs the same treatment.

Found while rebasing #261 onto `main`.

mauripunzueta:
Fixes #274.

## The bug

`crates/hts/src/backends/sqlite/value_set.rs` reads from and writes to a `concepts_search_fts` FTS5 table (the typeahead ranker `fts_candidates_ranked_for_system` and its populate path `populate_concepts_search_fts_for_system`), but the table's `CREATE VIRTUAL TABLE` **existed nowhere in the repository**:

```
$ grep -rn "CREATE VIRTUAL TABLE.*concepts_search_fts" .
(no matches)
```

So the first text-filtered expansion of any system fails hard with `no such table: concepts_search_fts`, surfaced as an OperationOutcome exception. `main` currently fails `value_set_ops::expand_is_a_filter_with_text_filter_still_nests` on exactly this.

## Why CI didn't catch it — semantic merge conflict

| PR | added | green because |
|----|-------|---------------|
| **#194** | the code that uses the table (no DDL) | nothing in its tests reached the build path |
| **#233** | `expand_is_a_filter_with_text_filter_still_nests`, which does reach it | its CI ran against a base predating #194 |

Both merged green; the combination on `main` is broken.

## The fix

Add the table next to its siblings in `schema.rs`, and clear it at startup in the same two blocks that reset `concepts_fts` / `concepts_word_fts`.

The shape is derived from how `value_set.rs` uses it, not guessed:
- **`term`** is the only tokenised column — the INSERTs fill it from both the concept's `display` and every designation `value`, which is what lets a query match a concept by any synonym and rank with `bm25`.
- **`code`** is retrieved (`SELECT code … GROUP BY code`) and **`system_id`** is a post-filter (`WHERE system_id = ?`), so both are `UNINDEXED`.
- Same `trigram case_sensitive 0` tokenizer as `concepts_fts`, its empty-index fallback, so substring matching is consistent between the two.

```sql
CREATE VIRTUAL TABLE IF NOT EXISTS concepts_search_fts
USING fts5(system_id UNINDEXED, code UNINDEXED, term,
           tokenize='trigram case_sensitive 0');
```

## Verification

Against a real SQLite (3.45): the DDL applies, both populate INSERTs run, the `EXISTS` populated-probe returns true, and a trigram `MATCH` on a designation value returns the owning concept's code ranked by `bm25`.

@sandhums (author of #194) — please sanity-check the column/tokenizer choice against your intent for the search ranker; the behaviour is inferred from the queries, but you'd know if `term` was meant to carry anything else.

Found while rebasing #261 (issue #200) onto main; #261's CI is blocked on this because GitHub tests the PR *merged into main*.

### DIFF
--- a/crates/hts/src/backends/sqlite/mod.rs
+++ b/crates/hts/src/backends/sqlite/mod.rs
@@ -369,7 +369,8 @@ impl SqliteTerminologyBackend {
                 let _ = conn.execute_batch(
                     "DELETE FROM concepts_fts;
                      DELETE FROM concepts_fts_built;
-                     DELETE FROM concepts_word_fts;",
+                     DELETE FROM concepts_word_fts;
+                     DELETE FROM concepts_search_fts;",
                 );
 
                 // Update query-planner statistics for large tables.
@@ -461,7 +462,8 @@ impl SqliteTerminologyBackend {
         let _ = conn.execute_batch(
             "DELETE FROM concepts_fts;
              DELETE FROM concepts_fts_built;
-             DELETE FROM concepts_word_fts;",
+             DELETE FROM concepts_word_fts;
+             DELETE FROM concepts_search_fts;",
         );
         let _ = conn.execute_batch(
             "ANALYZE concept_hierarchy; ANALYZE concepts; ANALYZE concept_closure; \
--- a/crates/hts/src/backends/sqlite/schema.rs
+++ b/crates/hts/src/backends/sqlite/schema.rs
@@ -227,6 +227,19 @@ CREATE VIRTUAL TABLE IF NOT EXISTS concepts_word_fts
 USING fts5(system_id UNINDEXED, code, display,
            tokenize='unicode61 remove_diacritics 1');
 
+-- ── FTS5 ranked search index (preferred display + synonym designations) ───────
+-- Backs the typeahead ranker `fts_candidates_ranked_for_system`: `term` carries
+-- both the concept's preferred display and every designation value, so a query
+-- matches a concept by any of its synonyms, and bm25(concepts_search_fts) ranks
+-- them. Only `term` is tokenised; `code` is retrieved (SELECT/GROUP BY) and
+-- `system_id` is a post-filter, so both are UNINDEXED. Same trigram tokenizer as
+-- `concepts_fts`, which is the fallback when this index is empty, so substring
+-- matching behaves consistently between the two. Populated lazily per system_id
+-- (`populate_concepts_search_fts_for_system`); cleared on startup.
+CREATE VIRTUAL TABLE IF NOT EXISTS concepts_search_fts
+USING fts5(system_id UNINDEXED, code UNINDEXED, term,
+           tokenize='trigram case_sensitive 0');
+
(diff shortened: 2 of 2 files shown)


mauripunzueta:
Superseded by #273, which merged the same fix (create `concepts_search_fts` FTS5 table + valid ranked query) into `main` on 2026-07-15. `main` is green and no longer needs this. Closing as redundant.

Issue #288, opened 2026-07-16
HTS: ConceptMap/$translate rejects the R5-spec 'sourceCoding' parameter (accepts only code+system)


mauripunzueta:
## Problem

`hts.heliossoftware.com` rejects the R5-spec-defined `sourceCoding` parameter on `ConceptMap/$translate`, and also rejects the R4-era `coding` parameter. It accepts **only** `code` + `system`.

Reproduction against live HTS:

```bash
# R5 spec-correct: sourceCoding -> 400
curl -X POST https://hts.heliossoftware.com/ConceptMap/'$translate' \
  -H 'Content-Type: application/fhir+json' -d '{"resourceType":"Parameters","parameter":[
    {"name":"url","valueUri":"http://hl7.org/fhir/ConceptMap/cm-address-use-v2"},
    {"name":"sourceCoding","valueCoding":{"system":"http://hl7.org/fhir/address-use","code":"home"}}]}'
```
```json
{"resourceType":"OperationOutcome","issue":[{"severity":"error","code":"invalid",
 "diagnostics":"Missing required parameter: code or sourceCode (or targetCode for reverse)"}]}
```

The same request with `coding` (instead of `sourceCoding`) also returns 400. Only `code` + `system` succeeds:

```bash
curl -X POST https://hts.heliossoftware.com/ConceptMap/'$translate' \
  -H 'Content-Type: application/fhir+json' -d '{"resourceType":"Parameters","parameter":[
    {"name":"url","valueUri":"http://hl7.org/fhir/ConceptMap/cm-address-use-v2"},
    {"name":"code","valueCode":"home"},
    {"name":"system","valueUri":"http://hl7.org/fhir/address-use"}]}'
```
→ `200`, `match.part(concept).valueCoding.code = "H"`.

## Why this matters

Per the [R5 `$translate` definition](http://hl7.org/fhir/R5/conceptmap-operation-translate.html), `sourceCoding` is a valid way to supply the concept to translate. A conformant R5 client that sends `sourceCoding` gets a 400 from us. The error message (`"Missing required parameter: code or sourceCode"`) does not mention `sourceCoding` at all, so the client has no hint that a supported alternative exists.

`$validate-code` **does** accept `coding`, so this is inconsistent within HTS itself.

## How this surfaced

[...]

PR #289, opened 2026-07-16
fix(ci): fail the FHIRPath job when conformance tests silently skip (#287)

### LINKED ISSUE #287: CI: FHIRPath tx conformance tests skip silently instead of failing when HTS is unreachable
## Problem

The Windows "Test FHIRPath" job runs the official FHIR R5 test cases against `fhirpath-server`. The runner reports an errored test as `Assert.Inconclusive`, which the job records as **SKIPPED, not failed**.

That means terminology coverage can disappear from CI without the job ever going red. This is not hypothetical — it already happened. #262 removed the evaluator's default terminology server, which made `txTest01`-`txTest03` error out; the runner downgraded all three to skips, and the job stayed green while losing the only end-to-end coverage of `%terminologies`. It was found by reading the diff, not by CI.

#262 restores the coverage by pointing the job at `https://hts.heliossoftware.com`, but it does not close the underlying hole: if HTS is unreachable, the three tests skip again and the job is green again.

## Why this should fail, not warn

`hts.heliossoftware.com` is **our own infrastructure**, not a third-party service. An HTS outage is not ambient flake to be tolerated — it is real signal about our own product, and CI going red is the correct response to it. A green run must mean the terminology tests actually ran and passed.

This is the "no silently-skipped checks" rule: a check that stays green when its coverage vanishes provides no signal at all.

## Proposed work

- [ ] Post-process the runner's results in the Windows "Test FHIRPath" job and **fail the job** if `txTest01`, `txTest02`, or `txTest03` did not actually execute (skipped/inconclusive counts as not executed).
- [ ] Make the failure message distinguish the two causes — HTS unreachable vs. an evaluator regression — so the on-call read is immediate.
- [ ] Confirm the assertion catches the real regression: temporarily unset `FHIRPATH_TERMINOLOGY_SERVER` and verify the job goes red rather than green-with-skips.
- [ ] While the live-HTS assertion is being added, use it to re-validate the `txTest03` `$translate` expectation. #262 asserts it against an in-process stub whose response body was hand-verified against live HTS once; this job is where that claim gets checked continuously.

## Context

- Follow-up to #262 (which closes #217).
- Related: #183 / #184 (migration off the public `tx.fhir.org`).

mauripunzueta:
Closes #287.

## The hole

The Windows "Test FHIRPath" job could stay green while its coverage vanished. The validator exits `1` only when `(failed > 0 || knownFailuresFixed > 0)` — so a test that errored, was downgraded to inconclusive, or **never ran at all** exits `0` and the job passes. Green never meant "the suite ran".

**This is not hypothetical.** Replaying the gate added here against the last green run on `main` turns it red on **three real evaluator defects** that were being masked. Investigating the tx path turned up a fourth.

## What was actually broken

Two of the issue's premises turned out to be wrong, and the fix is shaped by the corrections:

- **The results JSON is fine.** The 422 path *does* record `Result=false` before `Assert.Inconclusive`. Skipping is a console/exit-code phenomenon, not a JSON one — so this needs **no change to the validator fork** and no release cycle.
- **`$translate` was genuinely broken against HTS.** Our client sent a `coding` parameter; terminology servers require `code`/`sourceCode` as a named parameter and reject a lone `coding` with `400 Missing required parameter: code or sourceCode`. `tx.fhir.org` tolerated it, our own HTS does not — so txTest03 failed on **every** call and was recorded as a skip. Verified end-to-end against live HTS.

### Defects fixed (the silence was hiding these)

| Defect | Was | Now |
|---|---|---|
| `ConceptMap/$translate` sent `coding` | 400 from HTS on every call → skipped | sends `code`+`system`, returns `H` against live HTS |
| `lowBoundary(-1)` / `highBoundary(-1)` | `Err("precision must be >= 0")` | empty, per the test cases |

`lowBoundary(32)` already returned empty, so the same out-of-range condition was handled two different ways in one function.

HTS also rejects the R5-spec `sourceCoding` — that's a server-side gap, filed separately as **#288** rather than silently absorbed by this client fix.

### The gate

- **Audit conformance coverage** — every test declared in `tests-fhir-r5.xml` must produce a passing result or be listed in `known-test-failures.json`. Derived from the XML rather than hardcoding `txTest01-03`, so it covers all 1046 declared tests and needs no edit when upstream adds a `txTest04`. It refuses to run if the manifest scrape comes back implausibly small — a gate that checks nothing passes loudest.
- **Purge `fhirpath-results/`** — the validator *merges* into that file, the runner checks out with `clean:false`, and the directory is untracked. A test that stopped executing kept its stale **passing** entry indefinitely. This is a precondition for any assertion here being meaningful.
- **Pre-flight HTS probe** — the server returns 422 + OperationOutcome for *both* "terminology unreachable" and "expression legitimately failed", so nothing downstream can tell an outage from a regression on its own. The probe supplies that signal out of band and asserts **values, not liveness** (an HTS that is up but expands `administrative-gender` to 3 concepts is an HTS regression, and a `/health` check would misreport it). It also re-validates txTest03's `$translate` expectation against live HTS on every run — issue item 4.
- **Fail when `fhirpath-server` never starts** — that step warned and exited 0, a second silent-skip hole of the same shape.
- **Pin the validator release** — with `getLatestRelease`, cutting a release in the validator repo changed this job's verdict with no commit here to review or bisect to.
- **`timeout-minutes`** — the job had no ceiling on a scarce self-hosted Windows runner.

### Declared what was already invisible

- `HTMLChecks/*` — `htmlChecks()` is unimplemented, the server answers 501, and the 4 tests were absent from the results with nothing recording that. Identical to the existing `testConformsTo` entry, just never written down.
- **txTest03's reason was factually false** on both servers (`"ConceptMap cm-address-use-v2 not available on test server"` — both `tx.fhir.org` and HTS resolve it). The real limitation is output type fidelity: `.value.code` returns `string`, not `code` — same family as `testContainedId`. It stays a known failure for that reason; the probe covers the `$translate` behaviour independently.

## Validation

The audit logic was replayed locally against the **real artifact** from green run `29528354772` and the real upstream XML:

| Scenario | Result |
|---|---|
| Last green run on `main`, replayed | **RED** on exactly the 3 boundary defects, nothing else |
| HTS unreachable | **RED** `HTS-UNREACHABLE`, tx tests *not* blamed on the evaluator |
| tx tests vanish entirely (503 path) | **RED** |
| Stale results file | **RED** `RESULTS-STALE` |
| Post-fix, HTS healthy | **GREEN**, with a notice asserting 3 tx tests ran |

That last row is the point: green now asserts something. A missing notice means the gate itself has rotted.

Three real bugs in the manifest scraper were caught by replaying against real data rather than by CI: one test is commented out in the XML, `mode="cda"` is skipped by the runner by design, and `mode="html"` hits the 501 path. The arithmetic now closes exactly (`NOT_IMPLEMENTED -> 7` = conformsTo 3 + html 4).

## Sequencing

**This should land before #262.** #262 restores tx coverage by pointing at HTS, but its `$translate` wiremock stub matches on method+path only with no body matcher — so it answers `H` to a request live HTS rejects with a 400, certifying a broken path as working. This PR fixes the underlying `$translate` bug, so #262 can rebase on it and tighten the stub.

## Not in scope (deliberately)

- **The validator fork's exit code** (`Program.cs` catches `AssertInconclusiveException` before `AssertFailedException`, and `skipped` never gates `return 1`). Architecturally the right fix, but it needs a release cycle, and since the JSON already carries `Result=false` we don't need it. Worth its own issue.
- **HTS blackbox monitoring.** CI should not be how we learn HTS is down at 2am — detection would be human-triggered, unbounded in latency, and pages the PR author who can't fix it. A better gate makes it *more* tempting to rely on CI as the alarm, so this deserves its own issue.
- **`ServerConfig.terminology_server` / `--terminology-server`** is stored and never read; only the env var works. Left alone here.

## Acceptance (issue item 3) — NOT done in this PR

Issue #287 asks for a run that proves the gate goes red when `FHIRPATH_TERMINOLOGY_SERVER` is unset. **That CI-side proof has not been performed**, so item 3 of the issue stays open.

What *has* been demonstrated, by replaying this exact audit logic against the real artifact from green run `29528354772`:

- HTS unreachable → **red** (`HTS-UNREACHABLE`, and the tx tests are explicitly *not* blamed on the evaluator)
- tx tests vanish entirely from the results → **red**
- stale results file → **red** (`RESULTS-STALE`)
- last green run on `main`, replayed → **red** on exactly the 3 boundary defects

So the negative paths are evidenced, but on a replay rather than on the runner. Worth doing as a one-off `workflow_dispatch` or a temporary commit before this is relied on.

## CI evidence (run `29537560963`)

Read from the job log rather than the job status, since "the job was green" is precisely the inference this PR exists to distrust:

- `FhirPath Validator release: v0.1.15` — pinned, via `getReleaseByTag`
- `Removing stale results directory from a previous run` — the staleness hazard was **real**, not theoretical; it fired on the very first run
- `HTS pre-flight: 3/3 green`, including `ConceptMap/$translate cm-address-use-v2/home -> ... = H` (the call that was 400ing)
- `Manifest: 1046 tests declared, 3 mode="tx"` — matches the local replay exactly
- **`Category: SKIPPED -> 0`** (was `3` on `main`) — the boundary fix, confirmed on the runner
- `KNOWN_FAILURE_RESOLVED -> 0` — txTest03 correctly stays known-failing on type fidelity
- `##[notice] All 1046 declared FHIRPath tests are accounted for`

Coverage: Codecov went 0% → **98.45%** patch. The 2 remaining uncovered lines are the async `.send()` round-trip in `translate()`, which needs a mock server (`wiremock` is #262's dev-dep, not on this branch). The request-construction logic where the defect actually lived is fully covered, and the live probe exercises the real round-trip every run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -421,6 +421,20 @@ jobs:
     name: Test FHIRPath
     runs-on: [self-hosted, Windows]
     needs: [test-rust]
+    # This job had no ceiling at all, so a hang held a scarce self-hosted Windows
+    # runner until GitHub's 6h default fired. Green runs are ~15-20 min (#287).
+    timeout-minutes: 45
+    env:
+      # The FHIR test cases include mode="tx" tests (txTest01-03) that exercise
+      # %terminologies. Point them at our own HTS rather than the public tx.fhir.org
+      # (#183). HTS serves operations at the root (no /r4, /r5 path segment).
+      # Job-level, not step-level: the pre-flight probe and the server must agree on
+      # the target, and two definitions would drift.
+      FHIRPATH_TERMINOLOGY_SERVER: https://hts.heliossoftware.com
+      # Pinned deliberately. With getLatestRelease, cutting a release in the validator
+      # repo changed this job's verdict with no commit here to review or bisect to.
+      # A test oracle must be a stable reference; bump this in a reviewed PR (#287).
+      FHIRPATH_VALIDATOR_TAG: v0.1.15
     steps:
       - name: Checkout code
         uses: actions/checkout@v5
@@ -434,14 +448,16 @@ jobs:
           script: |
             const fs = require('fs');
 
-            console.log('Fetching latest FhirPath Validator release...');
+            const tag = process.env.FHIRPATH_VALIDATOR_TAG;
+            console.log(`Fetching pinned FhirPath Validator release ${tag}...`);
 
-            const { data: release } = await github.rest.repos.getLatestRelease({
+            const { data: release } = await github.rest.repos.getReleaseByTag({
               owner: 'HeliosSoftware',
-              repo: 'Hl7.Fhir.FhirPath.Validator'
+              repo: 'Hl7.Fhir.FhirPath.Validator',
+              tag
             });
 
-            console.log(`Latest FhirPath Validator release: ${release.tag_name}`);
+            console.log(`FhirPath Validator release: ${release.tag_name}`);
 
             // Find the Windows executable archive
             const asset = release.assets.find(asset =>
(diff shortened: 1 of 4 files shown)


smunini:
Heads-up on some changes I made directly to this branch — apologies for touching your PR without asking first.

### 1. Unpinned the validator again

I reverted the `FHIRPATH_VALIDATOR_TAG: v0.1.15` pin back to `getLatestRelease` (cee09de). Your reasoning was sound — a test oracle that moves under you is genuinely hard to bisect — but we own the `Hl7.Fhir.FhirPath.Validator` fork, so pinning means a bump PR here every time we cut a release there, and that maintenance cost outweighs the stability for us. The new coverage gate also makes a validator-side change surface as a specific `COVERAGE-LOST` list rather than a silent flip, which takes some of the sting out.

### 2. Split the two `crates/fhirpath` fixes into their own PRs

This PR had grown to cover three separable things. The source fixes now live separately:

- #301 — `boundary_functions.rs`, out-of-range precision returns empty
- #302 — `terminology_client.rs`, `$translate` sends `code`+`system` not `coding`

Both are cut from `main`, not stacked, so they land in any order. This PR keeps `ci.yml` and `known-test-failures.json`. I left the JSON whole here rather than splitting it three ways — the `txTest03` reason really belongs with #302, and that's noted there.

### 3. This PR is now expected to fail, and that's the point

With the boundary fix moved out, `LowBoundaryDecimal3`, `LowBoundaryNegDecimal3` and `HighBoundaryDecimal3` fail and aren't declared as known failures, so the audit reports `COVERAGE-LOST`. Red here is the gate demonstrating the bug it was written to catch. Once #301 lands and this rebases, it goes green.

### One correction worth having

The comment on the new boundary test said these three were "downgraded to an inconclusive by the conformance runner and recorded as a skip." I pulled the `fhirpath-test-results` artifact from run 29625134263 to check, and it's actually worse than that — all three are recorded as outright failures:

```
LowBoundaryDecimal3     Result = False   Error (Processing) ... Invalid Argument: lowBoundary...
LowBoundaryNegDecimal3  Result = False   Error (Processing) ... Invalid Argument: lowBoundary...
HighBoundaryDecimal3    Result = False   Error (Processing) ... Invalid Argument: highBoundary...
```

Three tests sitting in the results file as explicit failures with error messages, and the job still concluded **success**. Not skipped — recorded, and ignored by the exit code anyway. I've corrected the comment in #301. It's a sharper case for this gate than the one that was written down.

### Separate finding, not addressed anywhere yet

While verifying the boundary behaviour I found six decimal cases that disagree with the spec on **scale** — e.g. `(-1.587).highBoundary()` returns `-1.5865` where the spec wants `-1.58650000`. That expression is `HighBoundaryDecimal4`, it expects `-1.58650000`, and it reports `Result: true` in that same artifact. The validator compares decimals numerically rather than by scale, so the trailing zeros — the precision information these functions exist to convey — are dropped without complaint.

Worth noting that **this gate as written would not catch it**: the audit asks "did every declared test pass," and this one passes. Detail is in #301; probably wants its own issue.

PR #283, opened 2026-07-15
feat(ui): Resources workspace — nav submenu, search, edit modal (#282)

angela-helios:
Closes #282.

The primary read/write workspace, tying together the pieces built across the UI branches — natural-language search + visual builder + saved queries (#255/#234), the schema-driven editor (#264), and the History & Versions diff (#236).

> **Targets `main`.** The schema-driven editor needs the FHIR Schema IR from `helios-fhir-validator`, which lived only on the experimental `feat/fhir-validator` branch (#232). Rather than stack on it, this PR **absorbs that crate** so it can target `main` directly — it supersedes #232. The write path stays **off by default** (`HFS_VALIDATION_MODE=off`): `$validate` is an additive endpoint and CRUD is unchanged unless an operator enables enforcement.

## Offline required-binding validation

This PR also wires **terminology binding checks with no terminology server**. The FHIR core value sets that expand offline (every `compose.include` names a `complete` in-bundle CodeSystem, no filters) are generated from the spec bundles into committed gzipped packs and embedded per version; a `CoreTerminology` provider checks `required` bindings against them. So `$validate` now rejects a code outside its required binding — e.g. `gender: masculino`, a Spanish *display* rather than the FHIR code `male` (codes are language-independent; the display can be localised, the code cannot):

validate binding

*(Real `OperationOutcome`s from `POST /Patient/$validate` against this build — no terminology server running.)*

The **resource editor runs the same check inline**: its structural pass already defers the required bindings, so the embedded value sets flag an out-of-value-set code on its row as you edit — no `$validate` round-trip:

editor binding

`HFS_VALIDATION_TERMINOLOGY` defaults to `embedded`; `remote` (HTS `$validate-code`) and `off` remain. Value sets that need a live server (SNOMED, LOINC, filtered includes) are absent from the packs and reported as not-checked rather than guessed.

## Left nav: a flat "Resources" entry

Following Brett's sidebar, Resources is a **single flat nav entry** — not a submenu. Searching, editing, and version history all live *inside* Resources (on the page and in its edit modal), so they have no separate nav entries; their routes stay reachable by URL. The sidebar is **collapsible** to an icon rail (the toggle by the logo), a preference that roams per-user via `/_user/settings`.

## The Resources page

[...]

Issue #249, opened 2026-07-10
UI testing strategy: adopt Playwright + axe-core for spec-driven browser tests in CI


smunini:
## Summary

`crates/ui` (helios-ui) has no browser-level test coverage. We should add one, wire it into CI, and structure it the way this repo already structures every other conformance concern: **an external normative spec, an executable conformance harness, and a CI gate.**

This issue captures a tool evaluation (Playwright vs. a Rust-native CDP driver — both actually built and run against a live `hfs`), a recommendation, and a phased plan.

A placeholder workflow, `.github/workflows/ui-tests.yml`, is being added to `main` so there is a CI seat to grow into. It is a **no-op-friendly skeleton**: it runs `cargo test -p helios-ui`, then boots the UI and runs the browser suite *only once `crates/ui/e2e/package.json` exists*. Until then the browser steps skip and the job stays green, so it gates nobody.

---

## The gap

Today `crates/ui` is covered by `tests/router_http.rs` and `tests/i18n_http.rs` — `tower::oneshot` calls that assert over the response **string**. They're fast and valuable and should stay. But they never construct a DOM and never execute `assets/theme.js`, so an entire class of behavior is structurally unreachable:

| Behavior | Where it lives | Reachable by `oneshot`? |
|---|---|---|
| `data-theme` applied **before first paint** (the FOUC guard) | `assets/theme.js`, loaded without `defer` | ❌ |
| `localStorage` cache + OS `prefers-color-scheme` precedence | `assets/theme.js` | ❌ |
| Toggle issues `PATCH /_user/settings` with an RFC 7386 merge-patch | `assets/theme.js` | ❌ |
| Graceful degradation when `/_user/settings` is unavailable | `assets/theme.js` | ❌ |
| Progressive enhancement — the README's core promise that the UI works **with JS off** | `layouts/base.html`, handlers | ❌ |
| htmx fragment swap actually replacing the target | `hx-*` attributes | ❌ |
| WCAG 2.2 AA conformance | rendered DOM + CSS | ❌ |

That last row is not hypothetical.

## This evaluation already found two real accessibility defects on `/ui`

Both reproduce on the current `main`, in **both** light and dark themes, and are rated `serious` by axe-core:

**1. `aria-prohibited-attr` — `crates/ui/templates/pages/index.html:42`**
```html
<span class="pill pill--square" aria-label="{{ i18n.t("chart-expand") }}">
```
[...]

PR #290, opened 2026-07-17
test(ui): Playwright + axe-core browser suite + fix 2 WCAG defects (#249)

angela-helios:
Addresses #249 (Phase 0 + Phase 1): a browser-level test ring for `crates/ui`, wired into CI, plus the two accessibility defects the issue's evaluation found.

> **Base.** Targets `feat/282-resources-page` so the suite runs against the current Resources workspace (the a11y scan covers `/ui/resources`, not just `/ui`). Re-target to `main` once #283 lands.

## What's here

A self-contained Playwright + axe-core harness under `crates/ui/e2e/` — the outer ring of the pyramid the issue describes. Nothing touches the cargo workspace; all Node lives in that folder.

| Spec | Covers |
|------|--------|
| `tests/a11y.spec.ts` | **axe-core WCAG 2.2 AA** over `/ui`, `/ui/resources`, `/ui/compartments`, `/ui/search-parameters`, in **light × dark** |
| `tests/theme.spec.ts` | `theme.js`: FOUC guard before first paint, OS-preference precedence, the `/_user/settings` **RFC 7386 PATCH**, server-roam, graceful degradation |
| `tests/no-cdn.spec.ts` | no off-origin requests (no CDN), no uncaught `pageerror`, no inline `<script>` blob |
| `tests/resources-editor.spec.ts` | Resources edit flows: Create targets the picked type, inline binding validation, Save blocked on invalid, raw-edit round-trips to the FHIR API |
| `tests/nojs/*.spec.ts` | the README's core promise — the UI **works with JavaScript disabled** (`nojs` project, `javaScriptEnabled: false`) |

**24 tests, green locally** (20 chromium + 4 nojs).

## The two defects (now fixed, and now caught)

Both were `serious` axe findings the issue called out; neither was reachable by the `tower::oneshot` tests.

1. **`aria-prohibited-attr`** — the chart-expand affordance was a `<span aria-label>`. `aria-label` is prohibited on a generic span, so AT discards the name and it isn't keyboard-operable. → made a real `<button>` (with a `button.pill` reset so it looks unchanged).
2. **`target-size` (SC 2.5.8)** — the theme-toggle buttons rendered 20×20; WCAG 2.2 AA requires ≥24×24. → grew the hit area to 24×24 (the icon stays small).

The a11y spec fails on these before the fix and passes after — the spec-first loop the issue asks for.

## CI

`.github/workflows/ui-tests.yml`: runs `cargo test -p helios-ui` (the fast Rust ring, always), then builds `hfs --features ui` and runs the browser ring. Gated on `crates/ui/**`; the browser steps skip cleanly if the e2e harness is absent, so it's safe on branches that predate it. Uploads the Playwright HTML report on failure.

## Also fixed: WCAG color-contrast (now strictly gated)

The muted body text (`#a0a0a0`, ~2.4:1) and white-on-celeste primary buttons / selected rail item (`#33b8ff`, ~2.2:1) failed contrast. Fixed while staying close to the design: `--muted` darkened to `#6a6a6a` in light (kept light on dark), a `--accent-strong` (`#0f77bd`) fills primary buttons behind white text, and a theme-specific `--accent-text` colors the selected item. **All four scanned routes are clean in both themes**, so the axe gate has **no exclusions** — every WCAG 2.2 AA rule (color-contrast included) is a hard failure.

> The exact shades are open to design refinement — they were chosen as the minimal move to clear 4.5:1 while keeping the brand blue.

## Scope / follow-ups (from the issue's open questions)

- **Not in this PR** (later phases): `playwright-bdd` + `.feature` files (Phase 2), `toHaveScreenshot()` visual baselines (Phase 3), cross-browser WebKit/Firefox (Phase 4).
- **Runner note:** the workflow installs the browser with `npx playwright install --with-deps chromium` on the self-hosted runner. If deps/sudo are a problem there, switching the job to the `mcr.microsoft.com/playwright` container is the drop-in alternative (issue's open question).

## Run it

```bash
cargo build -p helios-hfs --features ui
cd crates/ui/e2e && npm ci && npx playwright install chromium
npx playwright test
```


angela-helios:
The three production fixes this suite uncovered were extracted to #325 (per #323); once that merges to main and flows back through the chain, this PR drops to the test-only surface.

Auth-enabled coverage for the conformance self-fetch is tracked in #320.

smunini (changes requested):
This PR also doesn't target a branch that targets main.

angela-helios:
Same story as #283 — intentional stacking (#322, #323): the browser suite drives the Resources workspace, editor modal, and history pages, so it needs #283's UI underneath it to run at all. Basing it there keeps this diff to the test suite itself (~40 files) instead of dragging the whole UI chain back in.

**Don't merge while it targets `feat/282-resources-page`** — it would land into that branch, not main. It's the last link of the chain: once #283 lands on main, I retarget this one to main and it merges clean.

Worth the wait, in my view: this suite runs green today across the per-PR job and all five backends of the matrix, and it's the one that caught the ES-composite conformance bugs (#325), the S3 seeding/dashboard costs (#329), and a css truncation nothing else could see. If you want it flat against main sooner, I can retarget any time.

smunini (dismissed):

smunini (approved):

PR #291, opened 2026-07-17
feat(tenants): auto-register a tenant on its first write (#240)

### LINKED ISSUE #240: Auto-register tenants on first write
Follow-up to #213 / #220.

With the tenant registry in place, tenants provisioned via `POST /admin/tenants` get a registry row with `created_at` and an optional display name. But tenants that come into existence **organically** — the first time a new tenant id is used on a write — never get a row: they show up in `GET /admin/tenants` merged in from `count_by_tenant` as `registered: false, created_at: null`.

## Proposal

On the resource write path, when a write lands for a tenant id that has no registry row, register it automatically (no display name) so `created_at` is captured.

## Considerations

- **Hot-path cost**: this must not add a registry lookup to every write. Needs a cheap per-process seen-tenants cache (the registry is append-mostly, so a positive cache never invalidates), or an async/best-effort registration off the write path.
- **Race semantics**: double-register races already surface as constraint/duplicate-key/precondition errors on every backend — auto-registration should treat those as success.
- **Scope**: deliberately left out of #220 to keep the hot write path untouched (see PR discussion).

angela-helios:
Closes #240. Follow-up to #213 / #220.

## The gap

Tenants provisioned via `POST /admin/tenants` get a registry row with `created_at` (and an optional display name). Tenants that appear **organically** — the first time a new tenant id is used on a write — never did: they showed up in `GET /admin/tenants` only as data-only rows (`registered: false, created_at: null`), merged in from `count_by_tenant`. #220 left this out deliberately to keep the hot write path untouched.

## What this does

On the resource write path, when a write lands for a tenant id with no registry row, register it automatically (no display name) so `created_at` is captured.

- **No per-write registry lookup.** `AppState` gains a per-process, append-only seen-tenants cache (`DashMap`). The hot path pays only a cheap in-memory probe; only the *first* write for an id this process touches the registry at all. (The registry is append-mostly, so a positive cache never needs invalidating — exactly the shape the issue called for.)
- **Idempotent + best-effort.** A duplicate (a concurrent registration won the race) or an unsupported backend surfaces as an error that is simply dropped — it never blocks or fails the write, and never overwrites an existing row's display name.
- **Where it's wired:** `create` (including conditional create), `update`, conditional update, and `batch`/`transaction`. **Patch is intentionally not** wired: it reads an existing resource first, so the tenant was already written to — and therefore already registered.

## Design note: inline, cache-guarded

The issue offered two shapes — a seen-tenants cache, or async/off-path registration. I took the cache and register **inline** (a single `await` on the first write per tenant), rather than a detached `tokio::spawn`. It's deterministic (no spawn-timing races in tests or in `GET /admin/tenants` right after a write) and the cost is one insert per tenant per process; the cache is what keeps it off every write.

## Tests

`crates/rest/tests/admin_tenants.rs`:
- `first_write_auto_registers_the_tenant` — a brand-new tenant's first write yields `registered: true` with a `created_at`.
- `writing_to_an_already_registered_tenant_does_not_clobber_it` — a write to a pre-provisioned tenant keeps its display name; the duplicate registration is a silent no-op.
- `list_includes_data_only_tenants_with_counts` — updated: since writes now auto-register, it deregisters the tenant to keep exercising the discovered-but-unregistered merge path (still reachable via deregister).

All 26 tenant tests green; `cargo build -p helios-hfs` clean.

### DIFF
--- a/crates/rest/src/handlers/batch.rs
+++ b/crates/rest/src/handlers/batch.rs
@@ -93,6 +93,10 @@ where
                 message: "Bundle must have a type".to_string(),
             })?;
 
+    // A batch/transaction may create the tenant's first resource — capture it
+    // in the registry (#240). Cache-guarded, so only the first write pays it.
+    state.ensure_tenant_registered(tenant.tenant_id()).await;
+
     match bundle_type {
         "batch" => process_batch(&state, tenant, &prefer, &bundle, principal.as_ref()).await,
         "transaction" => {
--- a/crates/rest/src/handlers/create.rs
+++ b/crates/rest/src/handlers/create.rs
@@ -104,6 +104,11 @@ where
         });
     }
 
+    // Organic tenants — a tenant id first seen on a write — get a registry row
+    // so their created_at is captured (#240). Guarded by a per-process cache, so
+    // only the first write for an id touches the registry; best-effort otherwise.
+    state.ensure_tenant_registered(tenant.tenant_id()).await;
+
     // Check for conditional create
     if let Some(search_params) = conditional.if_none_exist() {
         debug!(search_params = %search_params, "Processing conditional create");
--- a/crates/rest/src/handlers/update.rs
+++ b/crates/rest/src/handlers/update.rs
@@ -106,6 +106,9 @@ where
         });
     }
 
+    // Capture an organically-appearing tenant in the registry (#240).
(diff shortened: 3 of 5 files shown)


smunini:
Close because I think I addressed the correct way to handle tennants in https://github.com/HeliosSoftware/hfs/pull/252

Issue #279, opened 2026-07-15
Postgres composite search is I/O-bound at volume — needs a denormalized one-row-per-group storage layout


mauripunzueta:
## Summary

At full data volume the Postgres composite search shapes (`code-value-quantity`, `combo-code-value-quantity`, `component-code-value-quantity`) run at ~12s median under 30 concurrent VUs and time out at the 30s `statement_timeout`. This is the one search shape that PR #260 could **not** fix — it is I/O-bound, and no query rewrite or index shape addresses it. It needs a storage-layer change.

Spun off from #224 / PR #260.

## Why no query fix works

The composite is extremely sparse: only **222 of 656,737** Observations match a typical `code-value-quantity` query. But each component of a composite instance is stored as its **own `search_index` row** sharing a `composite_group`, so answering "code = X AND value > Y **within one group**" requires reading many rows per resource and aggregating.

The shipped query (PR #260) prefilters and then `GROUP BY resource_id, composite_group HAVING MAX(CASE …)`. Measured against the real dataset, its cost is dominated by random heap I/O:

```
Bitmap Heap Scan on search_index (actual rows=113014)
  Buffers: shared hit=212 read=111367
```

~112k buffer **reads** to fetch ~113k rows — close to one random page per row, because a parameter's rows are scattered across a ~10M-row table. That is I/O, not CPU or plan choice.

Every alternative was measured against the real 656,737-Observation dataset in CI and is worse (documented in `build_composite_condition` in `crates/persistence/src/backends/postgres/search/query_builder.rs`):

| candidate | real cost |
|---|---|
| shipped (OR-prefilter + IN + GROUP BY/HAVING) | ~1.0s cold / ~12s at 30 VUs |
| correlated `EXISTS` | 8.9s on a match, 20s on a zero-match |
| drop prefilter + covering INCLUDE index | 16.5s |
| self-join correlated on `composite_group` | flattened → ordered-scan trap |

## Proposed fix: one denormalized row per composite group

Write **one `search_index` row per `(resource, composite_group)`** with *every* component's value column populated, instead of one row per component. Then a single composite query becomes a plain multi-column predicate on one row:

```sql
WHERE param_name = 'code-value-quantity'
  AND value_token_code = '8867-4'
  AND value_quantity_value > 100
```

which a single covering index can answer directly — no per-group aggregation, no fan-out, no scattered heap reads. The sparse match becomes an index range scan returning ~222 rows.

## Work required

[...]

PR #319, opened 2026-07-21
fix pg statement_timeout on every connection

### LINKED ISSUE #285: Postgres: statement_timeout is applied to one pooled connection, so ~9 of 10 connections run uncapped
## Summary

`PostgresConfig::statement_timeout_ms` is applied to exactly **one** pooled connection and then that connection is returned to the pool. `SET` is session-scoped, so the other `max_connections - 1` connections (default: 9 of 10) never receive it and run with the server's default — usually no timeout at all.

Found while reviewing #212 / #275, which fixed a sibling defect in the same struct (`connect_timeout_secs` was declared, defaulted, serialized and never read).

## The code

`crates/persistence/src/backends/postgres/backend.rs`, in `PostgresBackend::new`:

```rust
let pool = Self::create_pool(&config)?;

// Verify connectivity
let client = pool.get().await…?;

// Set statement timeout
client
    .execute(&format!("SET statement_timeout = {}", config.statement_timeout_ms), &[])
    .await…?;

drop(client);   // ← back to the pool, still the ONLY connection carrying the SET
```

The connection checked out to verify connectivity is the one that gets the `SET`. Every connection deadpool creates afterwards, on demand up to `max_connections`, starts a fresh session with no `statement_timeout`.

So the timeout is enforced *nondeterministically*, on whichever fraction of requests happens to land on that one connection.

## Why this is more than a config nit

**The benchmark data in #224 and #227 may be partly an artifact of this bug.**

#227 reports for PG EX02: `6 rps, ~13% err, p95 = 30s (statement_timeout)`. #224 reports a composite search "that hits the 30s `statement_timeout` and returns 500".

With `default_max_connections() == 10` and exactly one connection carrying the `SET`, roughly **1 request in 10** should be killed at 30s while the other nine run unbounded. A measured error rate of ~13% sits right on top of that prediction, and `30s` is exactly `default_statement_timeout_ms()`. There is no server-side `statement_timeout` configured anywhere in `.github/` or the Dockerfiles, so the 30s kills observed in those runs can only be coming from this one connection.

If that reading is right, those suites are not measuring "the query takes >30s" uniformly — they are measuring a coin flip over which connection served the request. Worth confirming before either issue's numbers are trusted as a baseline.

## Why the fix needs care (and its own PR)

The obvious fix — a deadpool `post_create` hook that issues the `SET` on every new connection — would **newly enforce a 30s cap on queries that have never been capped**. Per #224 and #279, real composite searches at volume run for 12s median and one shape reaches ~500s. Making the timeout real would convert those from slow-but-successful into hard failures.

That may well be the correct outcome (a 500s query is not serving anyone), but it is a behaviour change that needs its own load validation and its own decision about the default — which is why #275 deliberately left it alone rather than smuggling it in.

## Related dead config in the same struct

`PostgresConfig::schema_name` (`backend.rs`, "Optional schema name for schema-per-tenant isolation") is likewise never read by any query path — only written as `None` in `Default` and asserted in one test. The actual schema-per-tenant machinery lives in `crates/persistence/src/strategy/schema_per_tenant.rs` and has its own separate config (`shared_schema`), so nothing is broken today; there is also no env wiring that could set it. It should either be wired to `search_path` or deleted, so the struct stops advertising isolation it does not provide.

That makes three fields in `PostgresConfig` declared-but-not-honoured, one of which (`connect_timeout_secs`) is fixed in #275. A sweep for the pattern is probably worth more than three separate fixes.

## Acceptance

- `statement_timeout` is applied to every pooled connection, not just the first.
- The change is load-validated against the shapes in #224 / #279 before merge, with an explicit decision on the default.
- #224's and #227's error rates are re-measured afterwards, since they may move once the timeout applies uniformly.
- `schema_name` is either honoured or removed.

bomanaps:
I Moved the `statement_timeout` setting into the connection startup packet so every pooled connection picks it up, not just the first one, and dropped the unused `schema_name` field while I was there, the rest of #285 needs the FHIR benchmark workflow run against this branch to check the #224/#279 shapes and re-measure #227's error rate, can someone with self-hosted runner access kick it off?


mauripunzueta:
Maintainer note on scope (pushed a follow-up commit, `d1c47cc29`, on top of yours — authorship preserved).

Quick clarification so the record is accurate: the startup-packet change that actually applies `statement_timeout` to every pooled connection (`cfg.options = "-c statement_timeout=…"`) already landed on `main` in `d93ea0884` (the #224 work), so it isn't introduced by this branch. What this PR genuinely contributes is exactly the part of #285 that was still missing:

- the **regression test** proving every pooled connection carries the GUC (not just the one borrowed inside `PostgresBackend::new`), and
- removal of the dead `PostgresConfig::schema_name` field — never read by any query path, satisfying #285's "schema_name is either honoured or removed".

My follow-up commit keeps both and makes two small adjustments:
- `get_client` is now `#[doc(hidden)] pub` rather than `pub`. It hands out a raw pooled connection that bypasses tenant scoping, so it shouldn't be stable API; `#[doc(hidden)]` keeps it callable from the out-of-crate integration test without advertising it (its siblings `pool()`/`base_registry()` are `pub(crate)`).
- restored the test's rationale comment and added a diagnostic assert message naming which connection lacked the GUC on failure.

On the remaining #285 acceptance — I kicked off the FHIR benchmark against this branch: https://github.com/HeliosSoftware/hfs/actions/runs/29882664448 (postgres, search) to check the #224/#279 shapes and re-measure #227's error rate. We still owe an explicit decision on the default `statement_timeout` before this fully closes #285.

bomanaps:
Pulled the artifact p50 42ms / p95 149ms / p99 216ms / 0 failures across 63,267 requests, so nothing is being clipped at the current 30s  but the import suite loaded bulk_1k 19MB search_index, composite cardinality 0, which is well below the 656k-Observation scale where #227's error rate showed up. I would hold the default at 30s: p99 sits at 140× under the cap, the shapes that legitimately need more room live in #279's rewrite, and `HFS_PG_STATEMENT_TIMEOUT_MS` already lets operators lift it during a migration.

mauripunzueta:
Agreed on holding the default at 30s.

Worth separating the evidence from the conclusion, though: the 63,267-request run is good proof that the startup-packet change doesn't regress normal traffic, but it can't speak to the cap itself — as you noted, composite cardinality was 0, and p99 216ms is two orders of magnitude short of 30,000ms. The load-bearing argument is your second one: #279 measures the composite shapes at ~12s median / 500s worst against 656,737 Observations, so no sane default accommodates them, and `HFS_PG_STATEMENT_TIMEOUT_MS` is the escape hatch for migrations. Fail-fast at 30s plus #279's storage rewrite is the right shape.

One thing I'd like recorded in the PR before this lands: at full volume, this fix makes composite timeouts **deterministic**. Per #285 the GUC previously only reached ~1 of 10 pooled connections, so most of those slow requests were completing (slowly) on uncapped connections. After fbeb115e2 every connection carries it, so the #279 shapes go from intermittently timing out to always timing out. That's the intended effect of the fix, not a regression — but whoever runs the next full-volume benchmark will see composite errors jump and read it as one. Worth putting in the PR description, and I'll note it on #279.

Related, and it's why the above matters: there's no handling for SQLSTATE 57014 (`query_canceled`) anywhere in the workspace. `crates/persistence/src/error.rs:712` maps every `tokio_postgres::Error` to `BackendError::Internal`, which `crates/rest/src/error.rs:741-744` returns as a 500 with the raw `canceling statement due to statement timeout` string. Before this fix that path was mostly unreachable; after it, it's the normal outcome for the #279 shapes at volume. A fired statement timeout should be a 504 with a clean OperationOutcome, not a 500 leaking the driver message. I've filed that as #353 rather than widening this PR — it adds a `BackendError` variant and a new arm in the REST error map, which is a different review surface from a pool config change. Shout if you'd rather fold it in here.

I also pushed one small commit to the branch (0a8feda4c): `HFS_PG_STATEMENT_TIMEOUT_MS` and `HFS_PG_POOL_WAIT_TIMEOUT_SECS` weren't in the README's `HFS_PG_*` table, only in a doc comment on `PostgresConfig`. Since the case for keeping the default at 30s rests on operators being able to lift it, that row needed to exist.

Two remaining notes, neither for me to action:

- The branch is 86 commits behind main. It'll need a merge before landing, and ideally a benchmark re-run on the merged state — the current numbers are against a fairly stale base. That's a maintainer call.
- The red "Test Rust" check is infra, not your code: it fails in 3s at step 2, `Verify DOCKER_HOST is set`, because fork PRs don't receive repo secrets, so the self-hosted job can't start. That's what has the PR showing as blocked.

smunini (approved):

PR #340, opened 2026-07-21
fix(hts): align $validate-code with tx.fhir.org for IG conformance (#339)

### LINKED ISSUE #339: HTS IG conformance: 12 residual gate failures (98% pass) — 3 root-cause groups
## Summary

The **HTS Terminology IG Conformance** suite (HL7 `fhir-tx-ecosystem-ig`) currently passes **585/600 (98.0%)** with **12 gate failures, identical across R4 and R5**. These are genuine HTS behavior deltas vs the reference server (tx.fhir.org) — reproducible, not flakes, and **not** caused by the recent R4B build fix (PR #332).

### Timeline / context
- Conformance was **green through 2026-07-10**, went **red on 07-11 with 62 failures** (a real test failure; build was fine that day).
- Failures were then whittled from **62 → 12** by 07-21.
- A **separate build break** (the `FhirVersion::R4B` `E0004` non-exhaustive-match, fixed in **PR #332** / merge `09091eb4b`) masked conformance runs 07-19 → 07-21. That build fix is unrelated to these 12 — it merely let the suite reach these tests again.

The 12 failures cluster into **3 root-cause groups** (authoritative messages from `report.json` → `.action[].operation.message`):

---

### Group A — retired+inactive concept emits 2 issues instead of 1 merged (5 tests)
`validation/validation-contained-good`, `inactive/inactive-3-validate`, `inactive-3a`, `inactive-3b`, `batch/batch-validate`

HTS emits **two separate** OperationOutcome issues — `"...has a status of inactive..."` and `"...has a status of retired..."` — for a concept that is both retired and inactive. The reference server merges them into a **single** issue: `"The concept 'X' has a status of retired and inactive and its use should be reviewed"`. Result: the `issue` array has one extra element (e.g. Expected `1`, Actual `2`).

Code: `crates/hts/src/backends/{sqlite,postgres}/value_set.rs` and `code_system.rs` (search `"has a status of"`, `INACTIVE_CONCEPT_FOUND`).

### Group B — version precedence in overload (1 test)
`overload/validate-good2a`

With multiple versions of a CodeSystem loaded, HTS returns version `1.0.0`; the expected result is `2.0.0` (newest when the request does not pin a version). Also surfaces as `Display 2` vs expected `Display #2`.

### Group C — CodeableConcept coding-selection order (6 tests)
`permutations/good-cc2-{all,enumerated,exclude-filter,exclude-import,exclude-list,import}-request`

Validating a `CodeableConcept` with multiple valid codings, HTS reports `code3` / `"Display 3"` as the chosen coding; the expected result is `code1` / `"Display 1"` — HTS appears to report the **last** matching coding instead of the **first**.

---

### Not in scope
`tx-ecosystem-ig/tests/icd-11/vs1.json` shows in logs but is in the workflow's `IGNORED_TESTS` (server-private-content) and is excluded from the gate.

No open PR currently addresses any of these groups (verified via file-level and keyword sweep; PR #338 touches `value_set.rs` but only for the incremental-FTS rebuild).

### Reproduce / triage
```
gh workflow run hts-ig-conformance.yml --ref main -f backend=sqlite
```
Then download the `tx-test-output-sqlite-R5` (and `-R4`) artifact:
- `report.json` → per-test expected-vs-actual (`.action[].operation.message`)
- `expected/` vs `actual/` → full response bodies (note `$optional$` markers = tolerated diffs)

### Suggested order
Group A first — largest (5 tests), most localized (message construction), lowest risk — then B, then C.

mauripunzueta:
## Summary

Fixes the **12 residual HTS IG-conformance gate failures** from #339 (585/600 → target 597/600), grouped into 3 root causes. Every change is in the shared, version-agnostic operations layer (`crates/hts/src/operations/validate_code.rs`), so **both the SQLite and PostgreSQL backends are fixed by the same code** — no backend/SQL edits, and `cs_precedence_order_by` is deliberately left untouched.

Root cause across all three: commit `f6008b29d` ("IG Publisher $validate-code compatibility", Jul 6) adapted to a fixture snapshot that upstream then changed on 07-11. These changes re-converge on the spec / reference-server behavior.

### Group A — retired+inactive → one merged issue (5 tests)
`validation/validation-contained-good`, `inactive/inactive-3{,a,b}-validate`, `batch/batch-validate`

A concept that is both retired and inactive now yields **one** merged warning `"The concept 'X' has a status of retired and inactive and its use should be reviewed"` instead of two separate issues. The generic backend-emitted `INACTIVE_CONCEPT_FOUND` issue is rewritten **in place** in `build_validate_response_async` (the sole chokepoint both backends flow through — backends only ever emit the generic "inactive" wording and a boolean). The top-level `concept_status` parameter is preserved (batch-validate depends on it). Works for `deprecated`/`withdrawn` too, not just `retired`.

### Group C — CodeableConcept first-coding-wins (6 tests)
`permutations/good-cc2-{all,enumerated,exclude-filter,exclude-import,exclude-list,import}-request`

When multiple codings validate, the response now echoes the **first** coding in input order (reference-server document order), not the last. Removes the deliberate `.rev()` at all three iteration sites (CS path, VS path, inline-VS) and fixes the reverse-order fallbacks (inline-VS `.last()`→`.first()`, drop `last_miss`).

### Group B — version precedence in overload (1 test)
`overload/validate-good2a`

Falls out of the Group C first-wins change **once the `coding_displays` / `coding_versions` / `coding_index` maps are rebuilt keep-first** (`.fold(..or_insert..)` instead of last-wins `.collect()`), so a duplicate `(system,code)` echoes the first coding's version/display (`2.0`/`Display #2`). This restores `d1f28f711`, which `f6008b29d` had reverted. No version-precedence heuristic is introduced (`CodeSystem.version` is an opaque string per spec; the two versions are pinned per-coding).

## Testing

- **4 new backend-agnostic regression tests** (run in the sqlite `check` CI leg): first-matching-coding echo for the CS and VS paths, keep-first version echo for a duplicate coding, and the single merged retired+inactive status issue. These lock the new behavior so a future upstream shift trips CI rather than silently regressing.
- **PostgreSQL parity** is proven end-to-end by the `tx-ecosystem-test` conformance legs (its `check` leg is `cargo check`-only by design).
- No local C linker in the dev env, so behavior is verified via `hts-ig-conformance.yml` (sqlite + postgres × R4 + R5). Verification: dispatch the workflow on this branch and confirm the 12 targeted tests flip green with zero new reds.

```
gh workflow run hts-ig-conformance.yml --ref fix/339-hts-ig-conformance -f backend=all
```

## Notes / scope
- The lexicographic-vs-semver ordering in `cs_precedence_order_by` (e.g. `10.0.0` < `2.0.0`) is a real but **separate** latent bug, not exercised by these fixtures; fixing it would need dialect-specific SQL and is intentionally out of scope.
- Any remaining IG-Publisher fixture that still expects last-wins / a second status issue would be a genuine upstream inconsistency to escalate, not re-patch here.

Closes #339

### DIFF
--- a/crates/hts/src/operations/validate_code.rs
+++ b/crates/hts/src/operations/validate_code.rs
@@ -1170,13 +1170,16 @@ async fn build_validate_response_async<B: TerminologyBackend>(
     lenient_display_validation: bool,
 ) -> Value {
     // For inactive concepts whose underlying status is more specific than
-    // "inactive" (e.g. `retired`, `deprecated`, `withdrawn`), the IG
-    // `inactive/validate-inactive-3*` fixtures expect TWO warning issues:
-    // one with text "...has a status of inactive..." (the canonical wording
-    // already emitted by the backend) AND a second with text using the
-    // specific status code (e.g. "...has a status of retired..."). Detect
-    // that case here by looking up the concept's `status` property and
-    // appending a second issue when needed.
+    // "inactive" (e.g. `retired`, `deprecated`, `withdrawn`), merge the two
+    // facts into the single canonical warning IN PLACE: rewrite the generic
+    // "...has a status of inactive..." issue the backend emitted to
+    // "...has a status of {specific_status} and inactive..." (e.g. "retired and
+    // inactive"). The reference server (tx.fhir.org) emits ONE combined
+    // `INACTIVE_CONCEPT_FOUND` warning whose status token is the join of the
+    // property status and the literal `inactive`, not two separate issues. We
+    // also surface the specific status as the top-level `concept_status`
+    // parameter. Idempotent: once merged, `already_has_specific` matches and
+    // re-entry is a no-op.
     if resp.inactive == Some(true) {
         let inferred_system = resp.system.clone();
         let lookup_system: Option<&str> = system.or(inferred_system.as_deref());
@@ -1195,23 +1198,17 @@ async fn build_validate_response_async<B: TerminologyBackend>(
                             .contains(&format!("has a status of {specific_status} and"))
                 });
                 if !already_has_specific {
-                    let inactive_issue = resp.issues.iter().find(|i| {
+                    // Rewrite the generic issue's text in place — preserving its
+                    // severity / fhir_code / tx_code / expression / location /
+                    // message_id — so a single merged warning names both the
+                    // specific status and `inactive`. Push nothing.
+                    if let Some(issue) = resp.issues.iter_mut().find(|i| {
                         i.message_id.as_deref() == Some("INACTIVE_CONCEPT_FOUND")
                             && i.text.contains("has a status of inactive")
-                    });
-                    if let Some(template) = inactive_issue.cloned() {
-                        let new_text = format!(
-                            "The concept '{cd}' has a status of {specific_status} and its use should be reviewed"
+                    }) {
+                        issue.text = format!(
+                            "The concept '{cd}' has a status of {specific_status} and inactive and its use should be reviewed"
                         );
-                        resp.issues.push(ValidationIssue {
-                            severity: template.severity,
-                            fhir_code: template.fhir_code,
-                            tx_code: template.tx_code,
-                            text: new_text,
-                            expression: template.expression,
-                            location: template.location,
-                            message_id: template.message_id,
-                        });
                     }
                 }
             }
@@ -2461,16 +2458,17 @@ async fn process_validate_code_inner<B: TerminologyBackend>(
             .find(|p| p.get("name").and_then(|v| v.as_str()) == Some("codeableConcept"))
             .and_then(|p| p.get("valueCodeableConcept"))
             .cloned();
-        // The IG fixtures expect the LAST matching coding to win (when several
+        // The IG fixtures expect the FIRST matching coding to win (when several
         // codings in a CodeableConcept all validate, the response echoes the
-        // last one). Iterate in reverse so the earliest "yes" we find is the
-        // last entry in the input.
+        // first one in input order, matching the reference server's
+        // document-order processing). Iterate forward and return on the first
+        // "yes".
         let cc_explicit_version = find_str_param(&params, "version");
         let cs_lenient = params
             .iter()
             .find(|p| p.get("name").and_then(|v| v.as_str()) == Some("lenient-display-validation"))
             .and_then(|p| p.get("valueBoolean").and_then(|v| v.as_bool()));
-        for (system, code) in codings.into_iter().rev() {
+        for (system, code) in codings.into_iter() {
             // Resolved per coding: the version pins are keyed by system canonical
             // (`system-version=<url>|<version>`), and a CodeableConcept may carry
             // codings from several systems, so the effective version can differ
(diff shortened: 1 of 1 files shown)


smunini (commented):
## Reconciliation with #194 — and conformance confirmation

Flagging that this PR overlaps with #194 (`f6008b29d`, "IG Publisher $validate-code compatibility"), since at a glance it can look like it *unwinds* that work. After digging through the history, it doesn't — the overlap is narrow, and this PR is on the correct side of it. Documenting the analysis here, plus the conformance run that settles the open question.

### The overlap is two behaviors, not #194's feature set

#194 was mostly IG-Publisher **parameter-shape** work (16 files): `system`-as-`url` alias, `valueCoding`/`valueCodeableConcept`, SNOMED `version=current`, BCP-13/mimetypes, inline ValueSet, UCUM composed units. **This PR touches only `validate_code.rs` and preserves all of that** — it still calls `coding_entries_from_codeable_concept`, still handles `valueCodeableConcept`/`valueCoding`, and doesn't touch UCUM/mimetypes/`version=current`.

What it actually reverses are two behavioral choices #194 *bundled in* but never called out in its summary:

1. **Coding selection: last-wins → first-wins.** #194 added `.rev()` at three iteration sites and reverted the keep-first maps (`.fold(..or_insert..)` → last-wins `.collect()`). This PR removes the `.rev()` and restores keep-first.
2. **Retired+inactive: two issues → one merged issue.**

### Why this PR is the correct side of the collision

The git archaeology on behavior #1 is decisive:

- **Jul 2 — `d1f28f711`:** "keep first coding's display/version for duplicate (system,code)" — first-wins, with a detailed rationale about `overload/validate-good2a`.
- **Jul 6 — #194 reverted it** back to last-wins (`.rev()`), as a side effect of chasing a different fixture; its description never mentions coding order.
- **This PR restores the `d1f28f711` design.**

So on coding order this isn't new behavior — it's putting back the earlier fix that #194 clobbered inadvertently. The root cause is real and visible in the diff: #194's own (now-removed) comment said the fixtures *"expect TWO warning issues"* — it was coding to the **tx-ecosystem-ig fixture snapshot as of Jul 6**. Upstream changed those fixtures on **07-11**, and this PR re-converges on the newer fixtures / tx.fhir.org reference-server behavior (first coding in document order; one merged status issue).

They can't both be right *on those two behaviors*, and this is the right resolution: it re-aligns with current upstream and restores the earlier fix, while keeping 100% of #194's actual features. Less "unwinding #194," more "un-bundling a regression #194 smuggled in."

### Conformance run — the caveat is now closed

The only open question was whether current upstream actually wants first-wins + merged-issue. Dispatched `hts-ig-conformance.yml` on this branch (`418bf3fd3`), all four legs:

| Leg | Result |
|-----|--------|
| sqlite / R4 | ✅ pass |
| sqlite / R5 | ✅ pass |
| postgres / R4 | ✅ pass |
| postgres / R5 | ✅ pass |

`FAILED=0` on every leg — *"All applicable tx-ecosystem tests passed (validator exit 0)"*. The 12 targeted tests flip green with zero new reds, on both backends and both FHIR versions. 👍

Run: https://github.com/HeliosSoftware/hfs/actions/runs/29882647368

*(Note: this conformance workflow is manual-dispatch / nightly, not part of the standard PR checks — worth being aware that a green PR-check set alone doesn't exercise it.)*

smunini (approved):
Approving. Conformance run 29882647368 is green on all four legs (sqlite + postgres × R4 + R5), FAILED=0. This correctly un-bundles the last-wins/split-issue regression from #194 while preserving its feature work, and restores the first-wins design from d1f28f711.

Issue #320, opened 2026-07-21
test(ui): exercise the conformance self-fetch and the UI under every auth mode


angela-helios:
From the 2026-07-20 review call: the UI now self-fetches `GET /SearchParameter` and `GET /CompartmentDefinition` over HTTP, and those calls must honor whatever auth mode the server runs in — no carve-outs. Today the browser suite (#249 / #290) only runs with auth disabled, and the auth-enabled path is untested.

### Tasks
- [ ] E2E leg with `HFS_AUTH_MODE` enabled and `HFS_OUTBOUND_BEARER_TOKEN` provisioned: conformance pages render with real data.
- [ ] E2E: auth enabled **without** an outbound token: the pages degrade to their warning state (no crash, no empty 404s), and the degraded fetch is retried on the next request.
- [ ] Smoke against a real IdP (Keycloak) with out-of-the-box defaults: first login should need no extra client configuration (no consent-popup loops).
- [ ] Document the supported modes (off / static bearer / IdP) for the UI self-call in the book, including the planned `JwtAssertionOutboundAuthProvider` follow-up noted in `crates/hfs/src/main.rs`.

Relates to #249, #290.

PR #346, opened 2026-07-22
feat(ui): wire the sidebar FHIR version selector

### LINKED ISSUE #343: ui: wire the sidebar FHIR version selector (static "R4" today)
The sidebar's version selector is a visual placeholder: it renders a hardcoded `"R4"` with no dropdown and no read path (`base.html`: *"Static until version switching has a server-side read path"*).

The pieces mostly exist:
- `search_params::enabled_versions()` lists the FHIR versions compiled into the build.
- The SearchParameter viewer already has a working per-page selector via `?version=`.

### Scope
- [ ] Render the compiled-in versions in the sidebar selector (a `<details>` menu like the tenant selector shell, links not JS).
- [ ] Decide the semantics: per-page `?version=` links (cheap, matches the sp viewer) vs. a persisted "current version" in `/_user/settings` that pages read (roams like the theme). Recommend starting with the former.
- [ ] Default label from the server's `HFS_DEFAULT_FHIR_VERSION` instead of the literal.
- [ ] e2e: selector renders the enabled set and navigation keeps the choice.

Relates to #282, #290.

angela-helios:
The sidebar's footer selector was a static chip hardcoding `R4`. This wires it per the v1 semantics agreed in #343:

- A `<details>` disclosure (same pattern as the tenant selector shell — works without JavaScript, opens upward from the footer) listing the FHIR versions compiled into the build, from the existing `enabled_versions()`.
- The label is the **server's default version** (`HFS_DEFAULT_FHIR_VERSION`), no longer a literal, and that option is marked `aria-current`.
- Each option is a plain link reloading the current page with `?version=` — the pages with a version dimension (SearchParameters viewer, Compartments) switch; pages without one ignore it. No persisted "current version" yet; that upgrade path (`/_user/settings`, like the theme) stays open in #343.
- `Status` (already rendered by every page) carries the default version, so no template struct grew a field.

Single-version builds render a one-option menu — harmless, and it goes away on multi-version builds.

Router test asserts the options, the `aria-current` mark, and the server-derived label. Browser-level coverage (open the menu, switch on the sp viewer) belongs to the e2e suite and lands on #290's chain once this merges.

Closes #343.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,7 +145,7 @@ jobs:
         timeout-minutes: 15
         env:
           RUN_MINIO_S3_TESTS: "1"
-          EXPECTED_MINIO_SETTINGS_TESTS: "4"
+          EXPECTED_MINIO_SETTINGS_TESTS: "5"
           # The single self-hosted Windows runner is a virtualized guest whose
           # CPU does not advertise SSE4.1 + PCLMULQDQ. The AWS S3 SDK's default
           # integrity protection computes a CRC32 on every upload via the
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3605,6 +3605,7 @@ dependencies = [
  "similar",
  "tokio",
  "tower",
+ "tracing",
  "unic-langid",
 ]
 
--- a/crates/hfs/src/main.rs
+++ b/crates/hfs/src/main.rs
@@ -543,6 +543,7 @@ async fn start_mongodb(
     // settings store: it keeps ownership of the backend Arc and wires the
     // settings-capable builder (like the SQLite/Postgres backends).
     let settings_store: Option<Arc<dyn SettingsStore>> = Some(backend.clone());
+    let ui_settings = settings_store.clone();
 
     // MongoDB primary; embedded SQLite sidecar for bulk-export job state.
     let export_bundle = {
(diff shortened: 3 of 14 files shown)


smunini (changes requested):
Is this selection stored in user settings so they don't have to change it every time they log in if they are using something other than the default FHIR version?

angela-helios:
Done — `8e69103db` adds the persistence.

Selecting a version now POSTs to `/ui/version`, which merge-patches `{ "fhirVersion": … }` into the same `/_user/settings` document the theme roams in, and bounces back to the page you were on. A per-request middleware resolves the effective version once (stored choice when it's compiled into the build, server default otherwise): the sidebar label reflects it everywhere, and the versioned pages (SearchParameters, Compartments) render under it — an explicit `?version=` still wins per page.

Details worth knowing:
- The user key mirrors the REST `UserKey` derivation (`issuer|subject`, or the local fallback) — with `/ui` outside the auth layer today, the authenticated modes remain #320's territory, same as the theme.
- Backends without a settings store (bucket-per-tenant S3 with no system bucket) degrade to a one-shot `?version=` redirect instead of failing.
- Covered by a round-trip test against a real SQLite settings store: POST → 303 back to the referrer → document contains the choice; unknown versions 400.

I also updated the PR body's claim — the "no persisted version yet" caveat no longer applies.

smunini (approved):

PR #354, opened 2026-07-23
fix(subscriptions): rehydrate the engine from storage at startup

### LINKED ISSUE #305: Subscriptions engine is not rehydrated at startup — stored active subscriptions stop delivering after a restart
## Summary

The Subscriptions engine holds all of its state in process memory and is populated **only** by live write handlers. Nothing reads previously-stored `Subscription` / `SubscriptionTopic` resources back into the engine at startup. After a server restart, subscriptions that are active in the database are inert — they will not match, evaluate, or deliver — until something happens to re-write the resource.

## Evidence

`SubscriptionEngine` is constructed once at `crates/rest/src/lib.rs:809` (`state.with_subscription_engine(...)`).

Every path that populates it is a write handler:

- `crates/rest/src/handlers/create.rs:206`
- `crates/rest/src/handlers/update.rs:189`
- `crates/rest/src/handlers/patch.rs:137`
- `crates/rest/src/handlers/delete.rs:89`

A grep across `crates/subscriptions/src`, `crates/rest/src`, and `crates/hfs/src` for `rehydrate` / `restore_subscription` / `load_subscription` returns nothing.

Structurally, rehydration isn't currently possible: the `SubscriptionEngine` struct (`crates/subscriptions/src/engine/mod.rs:35-47`) holds `topic_registry`, `manager`, `evaluator`, the channels, config and base URL — but **no storage handle**. All registries (`InMemoryTopicRegistry`, `SubscriptionManager`, `WebSocketManager`, `WsBindingTokenManager`) are `DashMap`-backed.

## Relationship to #170

#170 covers *cross-instance* WebSocket delivery and states that "Subscription resources themselves are database-backed". That's true of the resources, but it's orthogonal to this: the resources persist, the **engine's view of them does not survive a process restart**. Even a single-instance deployment loses all subscription behavior on restart. Worth tracking separately, though a shared registry may end up being the fix for both.

## Impact

- Restart / redeploy / crash silently stops all notification delivery for existing subscriptions, with no error surfaced — the resources still read back fine over the REST API, so the server looks healthy.
- Compounds with #170 in multi-replica deployments: each replica has an independently-populated view, so a subscription registered against replica A is invisible to replica B.

## Suggested fix

At startup (after storage is up, before serving), page through stored `Subscription` and `SubscriptionTopic` resources for each provisioned tenant and register the active ones with the engine — the mirror of what the write handlers already do. That needs a storage handle threaded into the engine or an explicit `rehydrate(&dyn ResourceStorage)` entry point called from `crates/hfs/src/main.rs`.

## Notes

Found during a full-codebase read; static analysis only, not reproduced against a running server.

mauripunzueta:
Closes #305.

## The bug

The subscription engine's registries are in-memory and were populated **only** by live write handlers (`create`/`update`/`patch`/`delete`). Nothing read stored `Subscription` / `SubscriptionTopic` resources back at startup, so a restart, redeploy, or crash left every stored subscription inert — no matching, no evaluation, no delivery. Nothing surfaced an error: the resources still read back fine over the REST API, so the server looked healthy while silently notifying nobody.

## What the issue understated

The issue proposes "register the active ones". Verifying against the code turned up something that changes the fix:

**`SubscriptionManager::update_status` is in-memory only — it never writes back to storage** (`manager/mod.rs:299-328`).

So a subscription a client POSTed as `requested`, which handshook successfully and became `Active` in memory, still reads `"requested"` from the database forever. Rehydrating only `status == "active"` resources would therefore have missed essentially *every* subscription the server itself activated — it would have looked like a fix while restoring almost nothing.

`requested` subscriptions are consequently re-activated through the same handshake path the write handler uses, bounded by a concurrency cap so a restart cannot stampede subscriber endpoints.

## Approach

`SubscriptionEngine::rehydrate` is the startup mirror of the write path. Per tenant, in dependency order:

1. `SubscriptionTopic` — native topics (R4B/R5/R6)
2. `Basic` — R4 backport topics, which are stored as `Basic` resources
3. `Subscription`

The order is a **correctness constraint**, not a preference: `register` rejects a subscription whose topic is not already in the registry (`TopicNotFound`). A test seeds the subscription *before* the topic so storage ordering can't accidentally produce the right answer.

### All backends, no DB work

Driven by `ExportDataProvider::fetch_export_batch` — the one enumeration primitive **every** primary backend implements: SQLite, PostgreSQL, MongoDB, S3, and `CompositeStorage`. One implementation covers all of them with **no schema change, no migration, and no per-backend query to drift**. It also already filters deleted resources and is cursor-paged, so peak memory stays bounded.

`SearchProvider::search` was the alternative — it carries a richer `StoredResource` — but standalone S3 reports search unsupported and would have silently rehydrated nothing.

| Primitive | All backends? | Carries `fhir_version`? | Per-DB work |
|---|---|---|---|
| `SearchProvider::search` | No (S3 standalone) | Yes | none |
| new `list_resources_of_type` | only if hand-written ×5 | Yes | **5 backends + composite** |
| `ExportDataProvider::fetch_export_batch` | **Yes** | No — recovered, see below | **none** |

### FHIR version is recovered, not guessed

`fetch_export_batch` yields raw resource JSON, so the persisted `fhir_version` is unavailable. It doesn't need to be: every version-dependent branch in the crate is a binary R4-vs-native split, and both sides are directly observable in the resource.

| | R4 (backport) | R4B / R5 / R6 (native) |
|---|---|---|
| topic | `criteria` / `backport-topic-canonical` ext | `topic` |
| channel | `channel` object | `channelType` coding |

Native markers win (they cannot appear in a genuine R4 backport resource); ambiguous resources fall back to the server default, matching what the write path stores for a request that omits `fhirVersion`.

### Status policy

| Stored status | Action | Why |
|---|---|---|
| `off` | skipped | Terminal — `can_transition_to` permits no target |
| `active` | registered, no handshake | Established in a previous process lifetime |
| `error` | registered, dormant | Never matches, but keeps `$status` honest |
| `requested` | registered **and activated** | See above — this is the common case |

### Where it's wired

In `build_app`, not `main.rs` as the issue suggested. `build_app` is the single funnel all eight backend start paths share, and the only place holding both the engine and the storage handle — `main.rs` cannot reach the engine, which is constructed and moved into `AppState` here, so wiring it there would need eight call sites plus an ownership change.

It runs in a background task: registration is fast, but activating `requested` subscriptions performs handshakes with retry/backoff against possibly-dead endpoints, which could otherwise delay serving by minutes. The spawn is guarded with `Handle::try_current()` so a non-async embedder gets a warning instead of a panic.

## Config

| Variable | Default | Description |
|---|---|---|
| `HFS_SUBSCRIPTION_REHYDRATE` | `true` | Reload stored subscriptions/topics at startup |
| `HFS_SUBSCRIPTION_REHYDRATE_BATCH_SIZE` | `500` | Resources per storage round-trip |
| `HFS_SUBSCRIPTION_REHYDRATE_HANDSHAKE` | `true` | Re-run the handshake for stored `requested` subscriptions |
| `HFS_SUBSCRIPTION_REHYDRATE_HANDSHAKE_CONCURRENCY` | `8` | Max handshakes in flight, across all tenants |

## Tests

`crates/subscriptions/tests/rehydrate_integration.rs` — against a **real** SQLite backend, simulating a restart by seeding storage and then constructing a *fresh* engine. Each test asserts the fresh engine is inert **before** rehydration, which is what makes these regression tests rather than feature tests: without that half, a no-op `rehydrate` would still pass.

- active subscription survives a restart, and genuinely matches events afterwards
- topics registered before subscriptions (subscription seeded first)
- `off` skipped; deleted subscriptions not resurrected
- every tenant covered; rehydrated subscriptions stay tenant-isolated
- one bad resource doesn't stop the scan (and doesn't mark the pass incomplete)
- disabling is a true no-op; rehydration is idempotent

Plus unit tests for version inference and report accounting.

## Notes / limitations

- **Not fixed here:** persisting status transitions, which would remove the re-handshake entirely. That needs a write path from the engine back into storage, changes the write handlers, and raises cross-replica conflict questions — closer to #170. Worth a follow-up issue.
- On R4 builds the `Basic` scan pages through *all* `Basic` resources in a tenant, not just topics (they're indistinguishable at the storage layer). Background work, bounded by batch size, but a tenant with many unrelated `Basic` resources pays some startup I/O.
- Does not address #170: each replica still rehydrates its own independent view. It does mean every replica now *has* a complete view, where previously each had only what it personally served.
- **Not compiled locally** — this environment has no C linker (`ld` only, no `cc`), so `cargo check`/`test` cannot run. `cargo fmt` passes and the code was hand-reviewed carefully, but CI is the first real compile.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3565,6 +3565,7 @@ dependencies = [
  "async-trait",
  "chrono",
  "dashmap",
+ "futures",
  "helios-auth",
  "helios-fhir",
  "helios-persistence",
--- a/README.md
+++ b/README.md
@@ -363,12 +363,14 @@ JWT/bearer auth (`HFS_AUTH_*`, e.g. `HFS_AUTH_ENABLED`, `HFS_AUTH_JWKS_URL`, `HF
 | `HFS_SUBSCRIPTION_MESSAGING_ENABLED` | `false` | Enable the FHIR Messaging subscription channel |
 | `HFS_SUBSCRIPTION_MESSAGE_SOURCE_ENDPOINT` | `HFS_BASE_URL` | Source endpoint URL used in outbound FHIR message headers |
 | `HFS_SUBSCRIPTION_ALLOW_PRIVATE_ENDPOINTS` | `false` | Allow subscription deliveries to private or loopback endpoints; intended for local development and CI only |
+| `HFS_SUBSCRIPTION_REHYDRATE` | `true` | Reload stored `Subscription`/`SubscriptionTopic` resources into the engine at startup, so subscriptions keep delivering across a restart |
+| `HFS_SUBSCRIPTION_REHYDRATE_HANDSHAKE` | `true` | Re-run the activation handshake for stored `requested` subscriptions during rehydration |
 | `HFS_BULK_EXPORT_ENABLED` | `true` | Enable the [Bulk Data Export](crates/rest/README.md#bulk-data-export) `$export` operation; when `false`, all `$export` endpoints return `501` |
 | `HFS_BULK_EXPORT_OUTPUT_BACKEND` | `local-fs` | Bulk export output store: `local-fs` or `s3`. See the [rest crate README](crates/rest/README.md#bulk-data-export) for the full `HFS_BULK_EXPORT_*` table |
 | `HFS_BULK_SUBMIT_ENABLED` | `true` | Enable the Bulk Data Submit `$bulk-submit` operation (HFS as Data Consumer — fetches a provider manifest and ingests it); when `false`, all `$bulk-submit` endpoints return `501`. Available on `sqlite`/`postgres` (+ `-elasticsearch` composites). See the "Bulk Data Submit" section of `CLAUDE.md` for the full `HFS_BULK_SUBMIT_*` table |
 | `HFS_BULK_SUBMIT_OUTPUT_BACKEND` | `local-fs` | Bulk submit status-artifact store: `local-fs` or `s3` |
 
-The SMTP delivery channel (`HFS_SUBSCRIPTION_SMTP_*`) and delivery-retry tuning (`HFS_SUBSCRIPTION_HANDSHAKE_*`) are documented in the [helios-subscriptions README](crates/subscriptions/README.md).
+The SMTP delivery channel (`HFS_SUBSCRIPTION_SMTP_*`), delivery-retry tuning (`HFS_SUBSCRIPTION_HANDSHAKE_*`), and the remaining startup-rehydration knobs (`HFS_SUBSCRIPTION_REHYDRATE_*`) are documented in the [helios-subscriptions README](crates/subscriptions/README.md).
 
 For detailed backend setup instructions (building from source, Docker commands, and search offloading architecture), see the [persistence crate documentation](crates/persistence/README.md#building--running-storage-backends).
 
--- a/crates/rest/src/lib.rs
+++ b/crates/rest/src/lib.rs
@@ -187,6 +187,8 @@ use tower_http::{
     trace::TraceLayer,
 };
 use tracing::info;
+#[cfg(feature = "subscriptions")]
+use tracing::warn;
 
 /// Creates the Axum application with default configuration.
 ///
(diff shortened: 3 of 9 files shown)


mauripunzueta:
Filed the follow-up flagged under *Notes / limitations*: #357 — subscription status transitions are never persisted.

That is the root cause of the re-handshake this PR has to perform: because a successful handshake never reaches storage, an activated subscription still reads `"requested"`, so rehydration cannot distinguish it from one that was genuinely never activated. Fixing #357 would let rehydration trust the stored status and drop the re-handshake entirely.

Writing up #357 also surfaced two consequences beyond this PR's scope: the delivery circuit-breaker is amnesiac (a subscription auto-turned `off` after repeated failures resumes delivering after a restart), and `GET /Subscription/{id}` permanently disagrees with `GET /Subscription/{id}/$status`.

smunini (approved):

Issue #390, opened 2026-07-24
Test suites can pass vacuously: env-gated tests silently skip in CI (sibling of #389)


aacruzgon:
> **Update 2026-08-04 — the `s3_es_tests.rs` row below is resolved and has been struck through.** `ci.yml:186` now sets `RUN_MINIO_S3_ES_TESTS: "1"`, added in `9376bf6e2` (2026-07-24) which cites this issue. The other two rows still hold: `HFS_TEST_MONGODB_URL` and `RUN_AWS_S3_TESTS` remain unset in every workflow. Verification in [this comment](https://github.com/HeliosSoftware/hfs/issues/390#issuecomment-5185702107).

## Summary

Sibling of #389. While investigating the Docker outage there, an audit of the test suites and workflows found several other places where CI reports green while tests **silently skip** — the suite passes whether or not the gated tests ran. #389 covered the email-channel instance; this issue tracks the rest.

Rust's libtest has no runtime "skip" outcome, so availability guards must choose between passing and failing when a dependency is missing — and everywhere below they chose "pass," which converts missing infrastructure or unset variables into green checkmarks.

## Confirmed gaps — green in CI, tests never run anywhere

| Suite | Gate | Status |
|---|---|---|
| ~~`crates/persistence/tests/s3_es_tests.rs` (S3+ES composite)~~ | `RUN_MINIO_S3_ES_TESTS=1` | ~~**Never set in any workflow** — the entire suite has never run in CI~~ — **RESOLVED 2026-07-24** in `9376bf6e2`: `ci.yml:186` sets it unconditionally on the main test step. Un-skipping revealed the suite was born broken; fixed in #391 (see comment below) |
| 8 tests in `crates/persistence/tests/mongodb_tests.rs` (`count_by_types`, `count_all_types`, `count_by_day`, `count_deltas_by_bucket`, `activity_histogram`, `count_by_tenant`, `tenant_registry_crud`, `purge_tenant_data`) | `HFS_TEST_MONGODB_URL` | **Never set in any workflow.** Main CI runs `cargo test --workspace --all-features`, so these compile and pass vacuously while their sibling tests in the same file run against a testcontainer |
| `crates/persistence/tests/s3_tests.rs` (real AWS S3) | `RUN_AWS_S3_TESTS=1` + `HFS_S3_TEST_BUCKET_TENANT_A/B` | Never set — presumably intentional (needs live AWS credentials), but undocumented, so it reads as coverage that doesn't exist |
[...]

PR #391, opened 2026-07-24
ci+deps+test(persistence): run the S3+ES composite suite on every agent, fix its latent defects, and bump past the crc-fast x86_64 panic

aacruzgon:
# Run the S3+ES composite suite on every agent, fix its latent defects, and remove the crc-fast panic at the root

## Summary

`crates/persistence/tests/s3_es_tests.rs` self-skips unless `RUN_MINIO_S3_ES_TESTS=1`, and no workflow ever set it. The suite had **never run in CI** — every green run passed it vacuously (#390).

Un-skipping it uncovered two further layers:

1. **The suite was born broken.** 3 of its 10 tests fail deterministically anywhere they actually execute. Fixed here.
2. **The AWS S3 SDK panicked on the self-hosted Windows agent.** This is the same defect #345 hit, and it is an upstream `crc-fast` bug rather than a property of the runner. Rather than gate the suite to Linux, this PR fixes it at the root by bumping the SDK past the fixed `crc-fast` — which also lets #345's checksum workaround be removed.

Net result: the suite runs **unconditionally on every agent**, verified 10/10 on the Windows agent that previously failed 8/10, and the Windows MinIO leg regains the checksum coverage it has been forfeiting since #345.

## Why

Three separate ways this was silently verifying nothing:

- **A suite that never ran.** `RUN_MINIO_S3_ES_TESTS` was unset everywhere, so `s3_es_tests` skipped in every CI run since it was written. The `finished in 0.01s` signature was there the whole time.
- **A suite that couldn't have worked.** Even when run, its harness never indexed anything into Elasticsearch, so every positive search assertion was failing — and its delete/isolation assertions could pass vacuously on timing.
- **A production bug hiding as a CI quirk.** The `crc-fast` panic was treated as a Windows-runner peculiarity. It is not: *any* HFS deployment writing to S3 on a CPU without SSE4.1+PCLMULQDQ panicked on every upload.

### `crates/persistence/tests/s3_es_tests.rs`

1. **The composite never indexed anything into ES.** The harness used the default sync config — `SyncMode::Asynchronous` — but never called `start_sync_workers()`. With an S3 primary (no durable outbox), every secondary sync failed with *"async worker was never started"*, and `composite.create/update/delete` swallow sync errors by design (primary succeeded, `warn!` only). Nothing ever reached ES, so all positive search assertions failed. The harness now builds the composite with `SyncMode::Synchronous`, which is also the semantics the tests assert. Verified by writing to the ES backend directly (worked all along) — the bug was purely the composite wiring in the harness.
2. **Fixed 100ms sleeps → bounded polling.** ES refresh visibility on a loaded runner can lag past any fixed sleep, so a `search_until()` helper polls (10s bound). The delete test now waits for the create to be *searchable* before deleting, so its absence assertion can't pass vacuously; the tenant-isolation test waits for tenant-a's visibility before asserting tenant-b sees nothing, making the isolation check meaningful rather than timing-dependent.

### `Cargo.lock`, `crates/sof/Cargo.toml`

- `aws-sdk-s3` 1.119.0 → **1.122.0**, pulling `aws-smithy-checksums` 0.63.12 → 0.64.4 → **`crc-fast` 1.6.0 → 1.9.0**, past the upstream fix.
- `helios-sof`'s `lru` 0.12 → **0.16**, following the SDK's own pin (`^0.12` → `^0.16`) so the lockfile keeps a single `lru`. The comment there already prescribed exactly this: *"Bump to the latest once the AWS SDK updates its lru pin."* Usage is `new`/`get`/`put` only; no source change needed.

### `.github/workflows/ci.yml`

- `RUN_MINIO_S3_ES_TESTS: "1"` on the existing "Run tests with all features" step — no new step, no extra build, no OS gate.
- Removed `AWS_REQUEST_CHECKSUM_CALCULATION` / `AWS_RESPONSE_CHECKSUM_VALIDATION` from the Windows MinIO step. See [Relationship to #345](#relationship-to-345). #345's job/step timeouts are untouched.

### 1. Reproducing the Windows failure

An earlier revision of this PR gated the suite to Linux on two assumptions: that the AWS SDK panics on the Windows agent, and that the ES container was unproven there. To replace conjecture with evidence, [run 30303350382](https://github.com/HeliosSoftware/hfs/actions/runs/30303350382) (commit `2717104ce`, reverted) removed the gate and pinned `test-rust` to `[self-hosted, Windows]`. [The job failed](https://github.com/HeliosSoftware/hfs/actions/runs/30303350382/job/90101442626) as intended:

```
thread 's3_es_test_create_then_search' panicked at
  ...\crc-fast-1.6.0\src\arch\mod.rs:121:13:
x86 features missing (SSE4.1 && PCLMULQDQ)

test result: FAILED. 2 passed; 8 failed; finished in 13.07s
```

**Assumption 1 confirmed.** The only two passing tests are precisely the two that never PUT an object to S3.

**Assumption 2 disproved — ES was never a blocker.** `s3_es_test_read_missing_returns_none` passed, and it runs the whole `make_harness()` path: ES testcontainer start, `ElasticsearchBackend` construction, and `initialize().await.expect("initialize ES backend")`. The container works on that agent.

### 2. Root cause

`crc-fast` 1.6.0, `src/arch/mod.rs`:

```rust
ArchOpsInstance::SoftwareFallback => {
    #[cfg(target_arch = "x86")]
    crate::arch::x86_software_update(state, bytes, params);

    // This should never happen, but just in case
    panic!("x86 features missing (SSE4.1 && PCLMULQDQ)");
}
```

A table-based software fallback has existed since 1.2.0, but the call to it is gated `#[cfg(target_arch = "x86")]` — **32-bit x86 only**. On `x86_64` that line compiles to nothing and control drops into the `panic!`. (The helper it would have called is broken the same way: its inner software call is also cfg'd to `x86`, and it discards the return value, falling through to an identical panic.) The `// This should never happen` comment is the tell — the author assumed every x86 CPU advertises SSE4.1+PCLMULQDQ; the agent is a virtualized guest that does not. The SDK reaches this on every upload via `aws-sdk-s3` → `aws-smithy-checksums` → `crc-fast`.

Upstream fixed it in **1.7.1**, under *"Ensure Miri passes on x86_64"* (PR #26) — not advertised as a CPU-compat fix, which is why the changelog doesn't flag it:

| version | panics | clean software fallback |
|---|---|---|
| 1.6.0 | yes | no |
| 1.7.0 | yes | no |
| **1.7.1** | **no** | **yes** |
| 1.8.0 – 1.10.0 | no | yes |

### 3. Why the fix needs an SDK bump

Not reachable by `cargo update` alone: `aws-smithy-checksums 0.63.12` requires `crc-fast ~1.6.0` (`>=1.6.0, <1.7.0`), there is no 1.6.1, and a `[patch.crates-io]` override cannot satisfy a tilde requirement either. It has to arrive through the SDK.

**Why 1.122.0 and not latest:** `aws-sdk-s3 1.140.0` needs `aws-smithy-types 1.6.1` → `serde ^1.0.228`, and the workspace pins `serde = { version = "=1.0.224" }` (`Cargo.toml:40`). 1.122.0 is the earliest release carrying the fix and it resolves against that pin.

### Relationship to #345

cc @smunini — #345 hit this exact panic in the Windows MinIO step and diagnosed it correctly down to the line number and dependency chain. Its workaround was the right call at the time: the fix genuinely wasn't reachable, for the tilde-pin reason above, and #345 said so explicitly, naming the root fix as *"expose host CPU features (SSE4.1 + PCLMULQDQ) to the Windows runner's hypervisor."*

What's new here is the mechanism. #345 described crc-fast 1.6.0 as having *"no software fallback on x86_64"* — true in effect, but the fallback **exists and is merely cfg'd out**, which makes this an upstream bug with an upstream fix rather than a hardware requirement. That opens a third option beyond "workaround" and "change the hypervisor": bump the SDK. With the panic gone, #345's env opt-out is unnecessary and is removed here.

This matters for coverage, not just tidiness. `WHEN_REQUIRED` doesn't merely route around a crash — it disables the SDK's default integrity protection, so the Windows leg was running a *different SDK configuration* than the Linux leg, and the checksum path (the thing that was broken) was the one part Windows never exercised. Both legs now run identical defaults and differ only by shell.

**#345's remaining caveat still stands:** *"any other hardware-CRC/crypto path on that runner could still hit this."* This fixes `crc-fast` specifically, not the general class of hardware-accelerated paths on a guest with a restricted CPU feature set.

### CI, on the Windows agent that originally failed

Each run below was temporarily pinned to `[self-hosted, Windows]` to guarantee it landed on `github-agent4`; every pin has been reverted, and the branch ships `runs-on: [self-hosted]`.

| run | change under test | `s3_es_tests` | MinIO settings |
|---|---|---|---|
| [30303350382](https://github.com/HeliosSoftware/hfs/actions/runs/30303350382) | gate removed, no SDK bump | **8 failed** / 2 passed | not reached |
| [30310519486](https://github.com/HeliosSoftware/hfs/actions/runs/30310519486) | + SDK bump | **10 passed** / 0 failed | opt-out still in place |
| [30318227484](https://github.com/HeliosSoftware/hfs/actions/runs/30318227484) | + #345 opt-out removed, `main` merged | **10 passed** / 0 failed | **5 passed** / 0 failed |

Final run ([job](https://github.com/HeliosSoftware/hfs/actions/runs/30318227484/job/90162379012)), with the SDK's default CRC32 path active on the CPU that used to panic:

```
running 10 tests
test s3_es_test_create_then_search ... ok
test s3_es_test_delete_then_search ... ok
test s3_es_test_multi_tenant_isolation ... ok
test s3_es_test_history_from_s3 ... ok
test s3_es_test_read_from_s3 ... ok
test s3_es_test_read_after_delete_returns_gone ... ok
test s3_es_test_s3_search_returns_unsupported ... ok
test s3_es_test_read_missing_returns_none ... ok
test s3_es_test_update_then_search ... ok
test s3_es_test_vread_from_s3 ... ok

test result: ok. 10 passed; 0 failed; finished in 14.45s
```

```
running 5 tests
test test_minio_settings_create_only_single_winner ... ok
test test_minio_settings_concurrent_patches_never_lose_an_update ... ok
test test_minio_settings_delete_is_idempotent ... ok
test test_minio_settings_round_trip ... ok
test test_minio_settings_stale_if_match_conflicts ... ok

test result: ok. 5 passed; 0 failed; finished in 0.80s
```

14.45s of real container work, not the `finished in 0.01s` skip signature — the first genuine full pass of this suite on a non-Linux agent. Because the eight previously-panicking tests had never reached their ES assertions on Windows, passing them also proves ES indexing, search visibility, and the `search_until()` bound work there — not merely that the panic is gone. And because real uploads now compute CRC32 on a CPU lacking those instructions, the `crc-fast` 1.9.0 fallback is genuinely exercised rather than just present.

### Local

- `RUN_MINIO_S3_ES_TESTS=1 cargo test -p helios-persistence --test s3_es_tests --features s3,elasticsearch` — **10/10** against real MinIO + ES containers.
- `cargo clippy --all-targets --all-features` with CI's exact `-A` flags — **exit 0**, re-run after merging `main`.
- `cargo test -p helios-sof --lib` — **127/127** (the `lru` bump).
- `cargo check -p helios-persistence --features s3,elasticsearch --tests` and `cargo check -p helios-sof --all-targets` — clean.
- `cargo metadata` exits 0 with no lockfile rewrite after the `main` merge, confirming the merged `Cargo.lock` is consistent.
- `cargo fmt --all` applied.

## Notes

- **This is a product fix, not just a CI fix.** Any HFS deployment writing to S3 on a CPU without SSE4.1+PCLMULQDQ panicked on every upload. CI merely surfaced it.
- **macOS is untested.** Linux and Windows agents are both proven now; `ci.yml` has a macOS linker branch, so an agent may exist. crc-fast's aarch64 path needs NEON+AES, which Apple Silicon has, so the same failure is unlikely — but it has not been demonstrated.
- **Future SDK bumps** are bounded by two constraints worth remembering: the `serde = "=1.0.224"` workspace pin caps `aws-sdk-s3` below 1.140.0, and `helios-sof`'s `lru` must track whatever the SDK pins or the lockfile carries two versions whenever `s3` is enabled.
- **Possible product-level footgun surfaced** (not addressed here): a composite configured with async sync but no started workers fails *every* secondary sync and only `warn!`s — a production misconfiguration would silently lose the entire search index. Flagged on #390 for triage.
- **History of the failure**: earlier revisions of this PR appeared green because the Test Rust job landed on the Windows agent, where the env var was empty and the suite skipped — a live demonstration of the exact vacuous-pass failure mode #390 describes.
- The anti-vacuous count assertion originally proposed here was dropped per the no-more-steps constraint; its guarantee moves to #390 PR B (guards fail loudly in CI at the code level).
- Remaining #390 items land separately: PR B (guard hardening, MongoDB testcontainer migration, `#[ignore]` for the AWS suite, MinIO container `run_id` label) and PR A2 (scheduled OIDC workflow for `s3_tests.rs`, pending IAM setup). The Redis twin-suite contract check rides with #269.


.github/workflows/ci.yml
```diff
@@ -82,6 +82,16 @@ jobs:
...
+        env:
+          # Un-skips the S3+ES composite suite (s3_es_tests.rs), which had
+          # never run in CI — nothing set this variable, so every run passed it
+          # vacuously (#390). Linux-only for now: the suite drives the AWS S3
+          # SDK, whose default checksum path (crc-fast) panics on the Windows
+          # agent's CPU (no SSE4.1/PCLMULQDQ — see the Windows MinIO step's
+          # AWS_REQUEST_CHECKSUM_CALCULATION workaround), and the ES container
+          # is unproven on that agent. On non-Linux landings the suite still
+          # skips silently; loud-failure hardening is #390 PR B.
+          RUN_MINIO_S3_ES_TESTS: ${{ runner.os == 'Linux' && '1' || '' }}
```
smunini: Let's make this work for non Linux operating systems.

smunini (changes requested):

aacruzgon:
### Windows failure reproduced deliberately — evidence

To stop justifying the Linux gate from conjecture, [run 30303350382](https://github.com/HeliosSoftware/hfs/actions/runs/30303350382) (commit `2717104ce`, now reverted) removed the gate and pinned `test-rust` to `[self-hosted, Windows]`. [The Test Rust job](https://github.com/HeliosSoftware/hfs/actions/runs/30303350382/job/90101442626) failed on `github-agent4`:

```
test result: FAILED. 2 passed; 8 failed; finished in 13.07s

thread 's3_es_test_create_then_search' panicked at
  ...\crc-fast-1.6.0\src\arch\mod.rs:121:13:
x86 features missing (SSE4.1 && PCLMULQDQ)
```

Two outcomes, one of which corrects this PR's earlier reasoning:

1. **The checksum panic is real**, and `mod.rs:121` is the `panic!` that crc-fast 1.6.0 reaches on x86_64 because its software fallback is gated `#[cfg(target_arch = "x86")]` and never compiles there. The only two passing tests are the two that never PUT to S3.
2. **Elasticsearch is fine on the Windows agent** — the gate's second justification was wrong. `s3_es_test_read_missing_returns_none` passed, and it runs the full `make_harness()` path including ES testcontainer start and `initialize()`.

Fixed upstream in crc-fast **1.7.1**, but unreachable via `cargo update` — `aws-smithy-checksums 0.63.12` pins `crc-fast ~1.6.0`. It arrives through `aws-sdk-s3 >= 1.122.0` → `checksums 0.64.4` → `crc-fast 1.9.0`, which I verified resolves and compiles clean. That bump is deliberately **not** in this PR; details and its two constraints (the `serde = "=1.0.224"` pin caps the SDK at 1.122.x; sof's `lru` pin needs to follow) are in the updated description.

The branch is back to `runs-on: [self-hosted]` with the Linux gate intact; the only lasting change from this experiment is the `ci.yml` comment, which now records what was measured instead of what was assumed.

aacruzgon:
### crc-fast fix verified on the Windows agent — 10/10

Follow-up to the [reproduction above](https://github.com/HeliosSoftware/hfs/pull/391#issuecomment-5097367331). This PR now carries the dependency bump, and it was verified on the same agent that failed.

[Run 30310519486](https://github.com/HeliosSoftware/hfs/actions/runs/30310519486) / [Test Rust](https://github.com/HeliosSoftware/hfs/actions/runs/30310519486/job/90124813897), temporarily pinned to `github-agent4` (pin reverted in `058da95e6`):

```
running 10 tests
test s3_es_test_create_then_search ... ok
test s3_es_test_delete_then_search ... ok
test s3_es_test_history_from_s3 ... ok
test s3_es_test_multi_tenant_isolation ... ok
test s3_es_test_read_after_delete_returns_gone ... ok
test s3_es_test_read_from_s3 ... ok
test s3_es_test_s3_search_returns_unsupported ... ok
test s3_es_test_read_missing_returns_none ... ok
test s3_es_test_update_then_search ... ok
test s3_es_test_vread_from_s3 ... ok

test result: ok. 10 passed; 0 failed; finished in 14.09s
```

Same agent, same suite: **8 failed → 0 failed**. 14.09s of real container work, so this is not the `finished in 0.01s` vacuous skip — it is the first genuine full pass of this suite on a non-Linux agent.

What changed: `aws-sdk-s3` 1.119.0 → 1.122.0 → `aws-smithy-checksums` 0.64.4 → `crc-fast` 1.9.0, plus `helios-sof`'s `lru` 0.12 → 0.16 to follow the SDK's pin and keep a single `lru` in the lockfile. `RUN_MINIO_S3_ES_TESTS` is now an unconditional `"1"` — no OS gate at all.

The eight previously-panicking tests had never executed their ES assertions on Windows, so passing them also proves ES indexing, search visibility, and the `search_until()` bound work on that agent — not just that the panic is gone.

Local: CI-style `cargo clippy --all-targets --all-features` exit 0, `helios-sof` 127/127, `cargo fmt` applied.

Two things deliberately left out, noted in the description: the Windows MinIO step's now-redundant `AWS_REQUEST_CHECKSUM_CALCULATION` opt-out (kept so this verification run had one variable), and macOS, which remains unproven.

aacruzgon:
### Checksum opt-out removed and verified on Windows — 5/5 and 10/10

[Run 30318227484](https://github.com/HeliosSoftware/hfs/actions/runs/30318227484) / [Test Rust](https://github.com/HeliosSoftware/hfs/actions/runs/30318227484/job/90148395245) on `github-agent4`, temporarily Windows-pinned (pin now reverted). This run removed #345's `AWS_REQUEST_CHECKSUM_CALCULATION` / `AWS_RESPONSE_CHECKSUM_VALIDATION` opt-out, so the SDK's **default CRC32 path was active** on the CPU that used to panic.

MinIO conditional-writes step — the suite #345 was fixing:

```
running 5 tests
test test_minio_settings_create_only_single_winner ... ok
test test_minio_settings_concurrent_patches_never_lose_an_update ... ok
test test_minio_settings_delete_is_idempotent ... ok
test test_minio_settings_round_trip ... ok
test test_minio_settings_stale_if_match_conflicts ... ok

test result: ok. 5 passed; 0 failed; finished in 0.80s
```

S3+ES composite suite, same run, on top of a fresh merge of `main`:

```
test result: ok. 10 passed; 0 failed; finished in 14.45s
```

So the `crc-fast` 1.9.0 fallback is genuinely exercised, not merely present: real S3 uploads now compute CRC32 on a CPU without SSE4.1/PCLMULQDQ and succeed. The Windows leg is back to the SDK's default integrity protection, which it had been opting out of since #345.

**On #345** (cc @smunini): the diagnosis there was right down to the line number, and the workaround was the correct call at the time — the fix genuinely wasn't reachable, since `aws-smithy-checksums 0.63.12` pins `crc-fast ~1.6.0` and a `[patch.crates-io]` override can't satisfy a tilde requirement. What changed is the mechanism: crc-fast 1.6.0 *does* ship a software fallback, it's just gated `#[cfg(target_arch = "x86")]` and unreachable on x86_64. Upstream fixed that in 1.7.1, so bumping the SDK removes the panic without touching the hypervisor. #345's timeouts are untouched.

#345's remaining caveat still holds: this fixes `crc-fast`, not the general class of hardware-accelerated CRC/crypto paths on a guest with a restricted CPU feature set.

The branch is unpinned (`runs-on: [self-hosted]`) and `main` is merged in.

smunini (approved):

Issue #406, opened 2026-07-27
sqlite: chained search through multi-target reference params and the :Type qualifier return zero results


angela-helios:
Found while manually testing the #394 builder against sqlite. Chained search works for the common single-hop cases but silently returns zero results in two situations, both reproducible on a fresh server:

```bash
curl -X PUT localhost:8080/Practitioner/gp-smith -H "Content-Type: application/fhir+json" -d '{"resourceType":"Practitioner","id":"gp-smith","name":[{"family":"Smith"}]}'
curl -X PUT localhost:8080/Patient/pat-linked -H "Content-Type: application/fhir+json" -d '{"resourceType":"Patient","id":"pat-linked","name":[{"family":"Linked"}],"generalPractitioner":[{"reference":"Practitioner/gp-smith"}]}'
curl -X PUT localhost:8080/Observation/obs-linked -H "Content-Type: application/fhir+json" -d '{"resourceType":"Observation","id":"obs-linked","status":"final","code":{"coding":[{"code":"1234-5"}]},"subject":{"reference":"Patient/pat-linked"}}'

# works:
curl 'localhost:8080/Observation?subject.family=Linked'                      # -> obs-linked
curl 'localhost:8080/Patient?_has:Observation:patient:code=1234-5'           # -> pat-linked
curl 'localhost:8080/Patient?general-practitioner=Practitioner/gp-smith'     # -> pat-linked

# zero results:
curl 'localhost:8080/Patient?general-practitioner.name=Smith'                # (1)
curl 'localhost:8080/Observation?subject:Patient.family=Linked'              # (2)
curl 'localhost:8080/Patient?general-practitioner:Practitioner.name=Smith'   # (1)+(2)
```

### (1) Multi-target reference params resolve to an invalid target type

`resolve_target_type` (`crates/persistence/src/backends/sqlite/search/chain_builder.rs:256-296`) falls back to `infer_target_type` when the param declares multiple targets. The inference map (`crates/persistence/src/search/chain_resolver.rs:590`) has no entry for `general-practitioner`, so the default branch capitalizes the param name and joins against a resource type named `General-practitioner` — which matches nothing, silently. The comment above `infer_target_type` already names the fix ('pick the first declared target from the SearchParameter registry'); spec-wise the server should search across every declared target that defines the leaf param, or at least an existing type from the registry.

### (2) The `:Type` qualifier on a chained param is not honored

[...]

PR #407, opened 2026-07-27
feat(ui): forward and reverse chaining in the visual query builder

### LINKED ISSUE #394: Implement Forward Chaining and Reverse Chaining in the Search visual builder
## Summary

Add support for **Forward Chaining** and **Reverse Chaining** (`_has`) to the visual query builder under **Search** in the web UI.

FHIR search supports chaining references across resources:

- **Forward chaining** — constrain a search by a property of a referenced resource. Example: `Patient?general-practitioner.name=Smith` (find Patients whose general practitioner is named Smith).
- **Reverse chaining** (`_has`) — constrain a search by a property of a resource that references the target. Example: `Patient?_has:Observation:patient:code=1234-5` (find Patients referenced by an Observation with a given code).

Today these can only be expressed by editing the raw query. This issue tracks exposing both capabilities as first-class controls in the visual builder.

## Scope

- Add UI controls in the Search visual builder to construct forward chains (reference param → target resource type → target param).
- Add UI controls to construct reverse chains via `_has:<Type>:<reference-param>:<param>`.
- Support chained parameter selection driven by the relevant resource type's search parameters.
- Round-trip: chained expressions built in the visual builder should serialize to the correct query string, and existing chained query strings should parse back into the visual builder where feasible.

## Notes

- Do not implement as part of this issue — this is a tracking/record item only.
- Follow up on multi-level chaining and reference-type disambiguation (`:TypeName` modifier on reference params) as needed.

angela-helios:
## Summary

Closes #394. First-class chaining controls in the Search visual builder, per Brett's Figma designs (the compact-row variants; the expanded multi-level panels are the follow-up the issue notes anticipate).

## What it does

- **Forward chains** — a condition row whose parameter is a reference (per the registry metadata) shows a drill-into affordance; clicking it converts the row to `ref-param › target-type › target-param`, serializing to `ref.param=` or `ref:Type.param=`. The target-type select is fed by the param's declared targets.
- **Reverse chains** — a '⧉ Filter a resource that links here' button adds a has-a-related row (type, linked-via reference param, leaf param), serializing to `_has:Type:ref:param=`. The linked-via suggestions are reference params of the chosen type that can point at the base type.
- **Round-trip** — chained keys in a pasted URL hydrate back into their rows; multi-level chains and nested `_has` stay as raw conditions that survive serialization untouched (follow-up).
- The `/ui/queries/params` catalog fragment now carries `data-type` and `data-targets` per option, promise-cached per type client-side.
- Fixes a latent builder papercut: the browser's dirty-value flag fired a native `change` after programmatic URL rewrites and rebuilt the rows mid-interaction (yanking focus and dropping edits); re-parsing now skips its own serialized echo.

Locale keys added for en, es, and de (the parity check runs in CI).

## Tests

- Params-catalog HTTP test asserts the new type/target metadata.
- Six builder e2e cases: forward hydrate + round-trip, typed chain, drill-into conversion, `_has` hydrate and scratch-build, multi-level raw fallback. Full local suite green.

Chained queries built here need #406's server fix to actually match (that PR carries the end-to-end run test); the two merge independently.


angela-helios:
Multi-level forward chains landed on this branch per the follow-up notes: hop-based rows (`ref › type` per hop), drill-deeper when the leaf resolves to a reference, any-depth hydration with per-hop :Type qualifiers. Nested `_has` stays a raw-condition fallback.

Chaining

smunini (approved):

Issue #366, opened 2026-07-23
ui(editor): make live-vs-on-save validation legible (terminology checks are save-time)


smunini:
Follow-up from #277 (resource editor prototype), noting current — deliberate — behaviour.

The editor runs the pure synchronous validation pass on every mutation (~300 µs), which is what anchors structural errors to their node. Anything requiring the async effects pass — notably live terminology checks, i.e. a code that is structurally fine but not in the bound value set — is only evaluated on save, by design (#265 §6).

That is the right default. This issue is to make it *legible* and to decide whether we want more:

### Scope
- Make it visible in the UI which errors are "checked live" vs "checked on save", so a clean-looking form that fails on save is not a surprise.
- Consider a debounced async terminology check on blur for bound fields (opt-in / configurable), rather than per-keystroke.
- Confirm on-save terminology failures anchor to their node the same way structural errors do.

### Notes
- Do not move terminology into the per-keystroke path — the latency budget does not allow it.
- Overlaps with the terminology picker work; if a picker only offers valid codes, the on-save surprise mostly disappears for that field.

angela-helios:
Implemented in #431 (stacked on the editor-depth chain; the Development link and auto-close activate when the stack collapses onto main and GitHub retargets the PR).

Issue #365, opened 2026-07-23
ui(editor): slicing support and terminology-backed pickers for bound fields


smunini:
Follow-up from #277 (resource editor prototype), deliberately out of scope there.

Two related gaps the prototype staged but did not build:

**Slicing beyond plain arrays.** The editor treats repeating elements as plain arrays. It does not understand slices — so it cannot offer "add an MRN identifier" vs "add an SSN identifier", cannot label existing entries by the slice they match, and cannot enforce per-slice cardinality.

**Terminology-backed pickers.** Fields with a value-set binding are plain text/token inputs. A bound field should offer a picker backed by the terminology server (HTS) — `$expand` for the bound value set, filtered as the user types — with required/extensible/preferred bindings treated differently.

### Scope
- Expose slicing (discriminators, slice names, per-slice cardinality) through the editor projection.
- Render sliced arrays as named add-choices, and match/label existing entries against their slice.
- Expose the binding (value set URL + strength) per element in the projection.
- Wire a picker in the Guided form to HTS `$expand`, degrading to free text when the value set is unavailable or the binding is weak.

### Notes
- These may split into two issues once someone picks them up; they are together here because both need richer per-element data out of the same projection call.

angela-helios:
Implemented in #430 (stacked on the editor-depth chain; the Development link and auto-close activate when the stack collapses onto main and GitHub retargets the PR).

Issue #363, opened 2026-07-23
ui(editor): profiled extensions in the Guided form (needs #232 extension catalogue)


smunini:
Follow-up from #277 (resource editor prototype), deliberately out of scope there.

The Guided form's ad-hoc extension path works today: you can add an extension by URL at an arbitrary node with the full `value[x]` pick, from the base `Extension` type, with no profile required.

What is missing is the **profiled** path — offering the extensions a profile actually declares, with their names, definitions, and constrained `value[x]` arms, instead of asking the user to type a URL.

This is blocked on the validator's extension catalogue: `extension-definitions.json` is not in the FHIR Schema pack yet (#232, and see #265 §5).

### Scope
- Get `extension-definitions.json` into the embedded pack(s) / IR so extension definitions are queryable.
- Extend the editor projection (`helios_fhir_validator::editor`) to return profiled extensions applicable at a node, alongside the ad-hoc URL entry.
- Surface them in the Guided form's "add" panel: named, described, with the constrained `value[x]` set rather than all 51 arms.

### Notes
- The ad-hoc path must keep working — this is additive.
- #237 and #238 want the same "what can I add here?" query, so the logic stays in the validator crate.

angela-helios:
Implemented in #432 (stacked on the editor-depth chain; the Development link and auto-close activate when the stack collapses onto main and GitHub retargets the PR).

PR #429, opened 2026-07-29
feat(validator,ui): mustSupport and short labels through the IR into the editor

### LINKED ISSUE #364: ui(editor): mustSupport emphasis and human field labels (needs IR support)

The Guided form currently shows raw element names as-is, because the FHIR Schema IR carries neither `mustSupport` nor human-readable labels/short definitions (#265 §8).

### Scope
- Carry `mustSupport` through the SD→schema conversion into the IR, per element.
- Carry the element `short` / `definition` text (or a label of some kind) through as well.
- In the editor: emphasise `mustSupport` elements (ordering and/or visual weight) and use the label instead of the raw element name, with the definition available on hover/expand.

### Notes
- Both are IR-level additions in `helios-fhir-validator`, so the editor projection and any other consumer (#237, #238) get them for free.
- Watch pack size — the definition text is not small; consider whether `short` alone is enough.

angela-helios:
## Summary

Closes #364. The SD→schema converter parsed neither the informational mirrors nor any human text — the IR fields (`modifier`/`mustSupport`/`summary`) existed but stayed empty in every pack, and the editor could only show raw element names.

## Changes

- **Converter**: reads `mustSupport` / `isModifier` / `isSummary` and the element `short`; the IR gains a `short` field; profile overlays carry it through layered schemas.
- **Packs regenerated**: gzipped size goes ~437 KB → ~830 KB total across the four versions with `short` alone — which is why the longer `definition` text is deliberately **not** carried, per the issue's size note.
- **Editor projection**: `Addable` exposes `must_support` + `short` (present rows read them off the `Arc<FhirSchema>` already in hand).
- **Guided form**: the short text becomes the row label with the raw element name beside it as the technical hint; must-support rows get weight and an MS badge (tooltip explains the contract); add-panel options show their short description and MS marker.

## Tests

New converter golden test pinning the mirrors and labels (set flags carried, unset stay absent). Full validator suite and helios-ui suites green; fmt clean.

First of the editor-depth stack (#364 → #365 → #366 → #363).


angela-helios:

Editor depth stack

smunini (approved):

PR #430, opened 2026-07-29
feat(validator,ui): slicing in the editor and terminology-backed pickers

### LINKED ISSUE #365: ui(editor): slicing support and terminology-backed pickers for bound fields

angela-helios:
Closes #365

## Summary

#365 **Stacked on #429** (#364's IR work). Two halves, per the issue:

**Slicing** — the IR already modeled it; nothing consumed it:
- The projection offers one named add-choice per slice, respecting per-slice cardinality against the document; adding into a slice seeds the item with the slice's pattern so it matches immediately ('add an MRN identifier' seeds `system`).
- Array items are labeled with the slice they match. `slice_matches` now also reads the slice schema's pattern/fixed keyword — the shape the converter actually emits for pattern discriminators — which the engine's own slicing sweep benefits from too.

**Terminology** — rows expose the binding's strength (chip) and canonical value-set URL; a new `/ui/editor/expand` proxies `ValueSet/$expand` on the configured `HFS_TERMINOLOGY_SERVER` (threaded through `mount` into `WebState`); bound inputs debounce a filtered expansion into a per-row datalist. The proxy answers 204 when no server is configured, so the picker degrades to the plain input.

## Tests

New projection tests: sliced arrays offer named add-choices with cardinality (spent slices vanish), slice-adds seed the pattern, items label by slice. Full validator suite (9 suites) and helios-ui suites green; fmt clean.


angela-helios:

Editor depth stack

smunini (changes requested):
Please address the Claude review issues identified

angela-helios:
Review findings addressed in ca25e8b18:

1. **Duplicated slice chip** — one of the two identical blocks removed.
2. **Duplicated binding chip** — same, removed.
3. **Datalist id collision** — the id was path-length + page-global fetch counter, which collides trivially and cross-feeds suggestions between fields. Both editor.js and resources.js now assign a dedicated incrementing id per input on first use; the id sticks to the input, so later fetches reuse it.
4. **`add_slice_element` un-merged lookup** — the seed lookup now goes through `merged_elements` like `schema_at`/`addable`, so a sparse derived profile that inherits the sliced element seeds the pattern correctly. Regression test added (`slice_add_seeds_through_an_inherited_element`) with a base + constraint-derivation profile whose own element map lacks the sliced element.

Validator lib 41 tests, UI crate suites, and a `--features ui` rebuild all green locally.

smunini (approved):

PR #431, opened 2026-07-29
feat(ui): make the editor's live-vs-save validation split legible

### LINKED ISSUE #366: ui(editor): make live-vs-on-save validation legible (terminology checks are save-time)

angela-helios:
Closes #366

## Summary

#366 **Stacked on the #365 PR.** The deliberate contract — structure, cardinality and required bindings per keystroke; FHIRPath constraints and configured terminology on save — becomes visible:

- A legend under the validity header states both halves.
- Bound fields carry an 'on save' chip, so a clean-looking form failing at save time is expected, not a surprise; where a terminology server is configured, the #365 picker already closes most of that gap live.
- On-save refusals now anchor like live errors: the server's OperationOutcome expressions map onto the same dotted row paths, each issue landing on its node (first also shown globally).

Per the issue's note, terminology stays out of the per-keystroke path.

## Tests

helios-ui suites green; locale keys in en/es/de with the parity check.


angela-helios:

Editor depth stack

smunini (approved):

smunini (approved):

PR #432, opened 2026-07-29
feat(validator,ui): profiled extensions in the Guided form

### LINKED ISSUE #363: ui(editor): profiled extensions in the Guided form (needs #232 extension catalogue)

angela-helios:
Closes #363

## Summary

#363 **Stacked on the #366 PR** — last of the editor-depth stack.

The blocker was the missing extension catalogue (`lib.rs` documented the gap). Now:

- `extension-definitions.json` (R4, 393 SDs, 4.5 MB raw) is vendored beside the other spec bundles; the pack generator folds it into the schema pack for **+23 KB gzipped**. The two-pass loader already resolves by URL, so the extensions-sugar `unknown-schema` gap closes as a side effect.
- The IR carries each extension's element applicability contexts (`StructureDefinition.context[].expression`), overlay-preserved.
- `SchemaRegistry::extensions_applicable(contexts)` answers 'what can I add here' — the same query #237/#238 want, living in the validator crate per the issue's note.
- Nodes that accept extensions list the applicable ones by name with their short description (the #364 labels); picking one runs the existing extension op with its URL. **The ad-hoc URL path keeps working** — this is additive.

## FHIR version coverage

The code here is version-agnostic — converter, `FhirSchema.context`, `extensions_applicable()`, the walk overlay, and the generator's `SOURCE_FILES` loop all four versions and skip the file when absent. Only the **data** is R4: `extension-definitions.json` is vendored for R4 alone and only `fhir_schemas_r4.json.gz` is regenerated, so R4B/R5/R6 packs carry zero `kind=extension` schemas and the picker renders nothing there. The ad-hoc URL entry still works on those versions, so this degrades rather than breaks.

Extending it is not uniform, tracked in #488:

- **R4B** is drop-in — `http://hl7.org/fhir/R4B/extension-definitions.json` returns 200; vendor it and re-run the generator, no code change.
- **R5/R6 are not.** Both 404 in the core spec — HL7 moved extensions to the cross-version Extensions Pack IG (`hl7.fhir.uv.extensions`), published under a different filename and package shape that `SOURCE_FILES` will not match. That needs a generator change plus a decision on mapping one cross-version pack onto per-version packs.

Separately and pre-existing on `main`: `crates/ui/src/editor.rs` resolves the registry with `FhirVersion::default()`, so the Guided form is R4-only regardless of pack contents.

Constrained `value[x]` arms inside a picked extension follow the extension's schema once resolved (deeper arm-narrowing in the add panel noted as follow-up).

## Tests

Registry test pins context-scoped offers against the R4 pack (birthPlace → Patient; own-name → HumanName.family; no cross-leak). Converter/validator/ui suites green.


angela-helios:

Editor depth stack

PR #440, opened 2026-07-29
feat(ui): Import/Export nav entries; the sidebar rail expands on hover

### LINKED ISSUE #438: ui: Import/Export nav entries, and a hover-expanding sidebar rail
Two sidebar updates from the current Figma designs:

**Batch & Data section.** The designs show four entries — Batch / Transaction, **Import**, **Export**, SQL-on-FHIR. The menu today has three: no Import, and the export entry is labeled "Bulk Export". Add the Import entry (placeholder like its siblings until the bulk-submit UI exists) and relabel Bulk Export to Export, in the three locales.

**Collapse behavior.** Drop the explicit expand/collapse toggle button from the brand row. The sidebar becomes a collapsed icon rail that expands when hovering the options (and on keyboard focus within it), overlaying the content instead of reflowing it. This replaces the persisted expanded/collapsed preference (nav.js + /_user/settings nav key).

e2e coverage in crates/ui/e2e/tests/chrome.spec.ts asserts the old toggle behavior and needs to follow.

### LINKED ISSUE #442: user-settings: a fresh user can never write conditionally — If-Match "0" is rejected against the ETag the GET itself served
`GET /_user/settings` serves the empty document as a real representation: `{}` with `ETag: "0"`. But `check_if_match` maps version 0 to "no current representation", so `If-Match: "0"` — the exact validator the GET handed out — always fails with 412. A client doing read-then-conditional-write (saved-queries.js does exactly this for the roaming recents/saved lists) deadlocks forever on a fresh user: every 412 retry re-reads `"0"` and fails again.

It went unnoticed because nav.js used to PATCH `{nav: …}` **without** If-Match on the first toggle, unconditionally creating the document — after which version ≥ 1 and conditional writes work. #438 removed nav.js, and the e2e suite now fails deterministically on a fresh database (`queries.spec.ts › a run is recorded under the Recent disclosure`, three `PATCH → 412` in the trace of PR #440's CI run).

Fix: in `crates/rest/src/handlers/user_settings.rs::check_if_match`, evaluate concrete tag lists against `version.to_string()` including version 0, so the advertised validator round-trips. `If-Match: *` keeps requiring an actually-stored document. The storage layer is already coherent — all four backends treat `Some(0)` as "expect absent" and insert.

angela-helios:
Closes #438
Closes #442

Two sidebar updates from the current designs:

**Batch & Data entries.** An **Import** entry joins the section (new mirrored icon), and *Bulk Export* becomes **Export** — placeholders like their siblings until those pages exist. Three locales updated.

**Hover-expanding rail.** The explicit collapse toggle is gone, together with nav.js, its `data-nav` state machine, and the persisted expanded/collapsed preference (the `/_user/settings` `nav` key is simply no longer read). The sidebar is now a fixed icon rail that expands as an overlay while hovered — or keyboard-focused, so it stays keyboard-accessible — and the content column never reflows. The narrow-viewport media block and the `has-nav-panel` grid variants collapsed into the one base state, since the rail is now the resting state everywhere. Labels stay in the accessibility tree while the rail is closed (same visually-hidden pattern as before).

## Screenshots

Resting rail:

resting rail

Hovered — expands as an overlay, with the new Import/Export entries:

hovered rail

Resources page (the three-column `has-nav-panel` layout keeps its own columns):

resources with rail

## Tests

- `router_http`: the page ships no toggle and no nav.js, and lists Import/Export.
- e2e `chrome.spec.ts` rewritten: rail width at rest, expands on hover and on keyboard focus, collapses on leave, content column doesn't move, no toggle in the DOM, and the Batch & Data section lists Import/Export/SQL-on-FHIR. Full suite green locally (57 passed), a11y and no-cdn included.

---

**Update:** the e2e `Recent` failure turned out to be a pre-existing server bug this PR uncovered (#442): `GET /_user/settings` hands out `ETag: "0"` for the never-written document, but `check_if_match` rejected that same validator on conditional writes, so a fresh user's read-then-conditional-write 412'd forever — previously masked by nav.js's unconditional first PATCH. The fix (one function in `handlers/user_settings.rs` + two regression tests, 25/25 in `user_settings`) is folded into this PR; `If-Match: *` keeps requiring an existing document, and all four storage backends already treated expected version 0 as insert-if-absent.


angela-helios:
CI status: everything is green except the one e2e failure — `queries.spec.ts › a run is recorded under the Recent disclosure` — which is the pre-existing server bug #442 that this PR merely uncovers (removing nav.js removed the unconditional PATCH that used to paper over it). #443 fixes it and is fully green; once it merges, re-running this PR's suite goes green. Suggested order: #443 first, then this.

PR #443, opened 2026-07-29
fix(rest): let If-Match "0" write the first user-settings version

### LINKED ISSUE #442: user-settings: a fresh user can never write conditionally — If-Match "0" is rejected against the ETag the GET itself served

angela-helios:
Closes #442

`GET /_user/settings` serves the empty document as a real representation — `{}` with `ETag: "0"` — but `check_if_match` mapped version 0 to "no current representation", so the exact validator the GET handed out could never satisfy a conditional write. A fresh user's read-then-conditional-write (saved-queries.js's roaming recents do exactly this) retried into 412 forever. It went unnoticed because nav.js used to create the document with an unconditional PATCH on the first toggle; #438 removes nav.js, and PR #440's CI then failed deterministically on `queries.spec.ts › a run is recorded under the Recent disclosure` — the trace shows the three `PATCH → 412` with `If-Match: "0"` against a fresh database.

Concrete tag lists now evaluate against the version number including 0, so `"0"` round-trips; `If-Match: *` keeps requiring an actually stored document. No storage changes: all four backends already treat an expected version 0 as insert-if-absent.

Tests: fresh-store round-trip (GET etag "0" → conditional PATCH creates version 1 → "0" is then genuinely stale again) and the `*`-still-requires-existence guard. 25/25 in `user_settings`.

Note for #440: its e2e `Recent` failure is this bug surfacing; once this merges, #440 rebases green.

### DIFF
--- a/crates/rest/src/handlers/user_settings.rs
+++ b/crates/rest/src/handlers/user_settings.rs
@@ -326,22 +326,28 @@ fn parse_if_match(conditional: &ConditionalHeaders) -> RestResult<EntityTagPreco
 }
 
 /// Checks a parsed precondition against the currently stored `version`, where
-/// version `0` means "no document exists yet".
+/// version `0` means "no document has been written yet".
 ///
-/// Version `0` is mapped to "no current representation" so the shared evaluator
-/// gives the semantics this endpoint already had: `*` requires an existing
-/// document, and a concrete tag never matches a document that does not exist.
+/// The GET handler serves that state as a *real representation* — `{}` with
+/// `ETag: "0"` — so the validator it hands out must round-trip: a concrete
+/// `If-Match: "0"` matches version 0, or a fresh user's read-then-conditional-
+/// write could never succeed (#442). `If-Match: *` keeps requiring an actually
+/// stored document, per its "unless it does not exist" semantics.
 fn check_if_match(precondition: &EntityTagPrecondition, version: i64) -> RestResult<()> {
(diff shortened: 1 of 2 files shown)


angela-helios:
Folded into #440 per review flow — its branch already carries this commit (`b3815439e`, merged as `5a7ea5a32`), so one PR delivers the rail and the settings fix it uncovered. Closing without merge.

PR #452, opened 2026-07-30
ci: reap testcontainer volumes, not just containers

smunini:
## The bug

Every CI cleanup path force-removes its containers with `docker rm -f`, which **does not remove anonymous volumes** — that requires `-v`.

The postgres, elasticsearch, mongo, minio and keycloak images all declare `VOLUME` in their Dockerfiles, so every testcontainer gets an anonymous volume, and every reap orphans it permanently. This has been leaking on every run, on every workflow, for as long as the reaps have existed.

## How it surfaced

The shared self-hosted Docker host is an LXC container on a ZFS dataset (`nvme1/subvol-106-disk-0`) with a 200G refquota. The orphaned volumes filled it until containerd could no longer write `io.containerd.metadata.v1.bolt/meta.db`, at which point *every* container start failed:

```
Failed to start Postgres container: Client(CreateContainer(DockerResponseServerError {
  status_code: 500,
  message: "write /var/lib/containerd/io.containerd.metadata.v1.bolt/meta.db: disk quota exceeded"
}))
```

Note `EDQUOT` ("disk quota exceeded"), not `ENOSPC` — the ZFS dataset quota, not a full disk. The failures are protean, because *anything* needing a container dies: see #449 for the Elasticsearch variant.

Seen in [run 30559660973](https://github.com/HeliosSoftware/hfs/actions/runs/30559660973/job/90929100594) (`Test Rust`, `postgres_bootstrap_lock.rs`), and it is not PR-specific — any run landing in that window fails the same way.

The fingerprint on the host, *after* a manual prune had already recovered it:

```
TYPE            TOTAL     ACTIVE    SIZE      RECLAIMABLE
Images          11        4         6.316GB   3.715GB (58%)
Containers      5         5         13.41MB   0B (0%)
Local Volumes   144       4         38.5GB    20.02GB (51%)
Build Cache     158       0         2.975GB   2.583GB
```

144 volumes, 4 active. 140 orphans.

## The change

`docker rm -f` → `docker rm -fv` at 84 call sites across 16 workflows, plus a comment at the canonical reap site in `ci.yml` recording why `-v` is load-bearing.

## Safety

`docker rm -v` removes **only** anonymous volumes; named volumes are never touched. No workflow in this repo runs `docker volume create` or mounts a named volume — every volume in play is anonymous, so all of them were pure leak.

`actionlint` was diffed against `main`: identical findings, no new ones. The only delta is three pre-existing `SC2086` warnings shifting column 16→17, from the extra character on those lines.

## Follow-ups

This stops *new* orphans, but an already-orphaned volume has no container left, so no `docker rm -fv` will ever reclaim it. The backlog still needs sweeping — see the per-run GC in the follow-up PR, which supersedes #451's nightly janitor.

#451 also carries an unrelated real fix (`ci(ui-matrix): route the Elasticsearch self-traffic around the runner proxy`) that is worth keeping.

### DIFF
--- a/.github/workflows/audit-events.yml
+++ b/.github/workflows/audit-events.yml
@@ -229,7 +229,7 @@ jobs:
         if: matrix.backend == 'sqlite-elasticsearch' || matrix.backend == 's3-elasticsearch'
         run: |
           ES_CONTAINER="es-audit-${{ matrix.backend }}-${{ github.run_id }}-${{ github.run_attempt }}"
-          docker rm -f "$ES_CONTAINER" 2>/dev/null || true
+          docker rm -fv "$ES_CONTAINER" 2>/dev/null || true
           docker run -d --name "$ES_CONTAINER" -p 0:9200 \
             -e "discovery.type=single-node" \
             -e "xpack.security.enabled=false" \
@@ -260,7 +260,7 @@ jobs:
         if: matrix.backend == 'postgres'
         run: |
           PG_CONTAINER="pg-audit-${{ github.run_id }}-${{ github.run_attempt }}"
-          docker rm -f "$PG_CONTAINER" 2>/dev/null || true
+          docker rm -fv "$PG_CONTAINER" 2>/dev/null || true
           docker run -d --name "$PG_CONTAINER" -p 0:5432 \
             -e POSTGRES_USER=helios \
             -e POSTGRES_PASSWORD=helios \
@@ -291,7 +291,7 @@ jobs:
         if: matrix.backend == 'mongodb'
         run: |
           MONGO_CONTAINER="mongo-audit-${{ github.run_id }}-${{ github.run_attempt }}"
-          docker rm -f "$MONGO_CONTAINER" 2>/dev/null || true
+          docker rm -fv "$MONGO_CONTAINER" 2>/dev/null || true
           docker run -d --name "$MONGO_CONTAINER" -p 0:27017 mongo:7.0 --replSet rs0 --bind_ip_all
 
           echo "MONGO_CONTAINER=$MONGO_CONTAINER" >> $GITHUB_ENV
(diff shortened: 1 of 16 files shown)


mauripunzueta:
Heads-up on an overlap — #453 (opened just before this, for the benchmark specifically) touches the same four lines in `fhir-benchmark.yml`.

We reached the same conclusion independently, which is reassuring. Your commit has a detail I didn't have and that I've now cited: the host's **200G ZFS quota**, and that the leak is **anonymous** volumes because the postgres/ES/mongo images declare `VOLUME`.

No disagreement in content — #453 already does `docker rm -f -v` at all four of the `fhir-benchmark.yml` sites this PR changes, so it's a strict superset *for that one file*. It also adds a named volume for the PG data (the benchmark's data was in the container's writable layer, so nothing could reclaim it independently of the container), a start-of-job reaper, a disk canary, and moves the artifact uploads ahead of the post-processing steps.

**Suggested order: merge this one first.** It's broader (16 files) and lower-risk, and it fixes workflows #453 doesn't touch. I'll rebase #453 on top — its `fhir-benchmark.yml` changes supersede the four lines here, so the conflict resolution is "take #453's version of that file".

One thing reading this PR fixed in mine: your `ci.yml` sweep uses an age guard, and I'd omitted one. Without it, two benchmark dispatches on *different refs* would reap each other's containers mid-run — the `concurrency` group only cancels same-ref runs. Now aligned with your idiom (`a101d0e6a`).

mauripunzueta:
Quantified evidence for this PR's leak, from a benchmark dispatch I just ran on the same host ([run 30576511473](https://github.com/HeliosSoftware/hfs/actions/runs/30576511473)):

```
Images           11       5       6.751GB    4.343GB (64%)
Containers       20      20     448.3MB         0B (0%)
Local Volumes   187      19      23.73GB   22.13GB (93%)
```

**187 volumes, 22.13 GB reclaimable — 93%.** Against the 200 G quota you cite, that is the bulk of the problem, and it is exactly the anonymous-volume leak `-fv` fixes. Containers are only 448 MB; the volumes are where it all went.

Also worth noting for the record: two containers orphaned by benchmark run 30550776427 were **still present 5½ hours later** (ages 333m/334m) when my reaper removed them. So the host recovering enough for CI to go green earlier today did not mean it had been cleaned — it just had enough headroom again. Some manual pruning on the host is still worth doing; neither PR retroactively reclaims those 22 GB.

Issue #448, opened 2026-07-30
test: manual sanity pass of search over Synthea data — SQLite first, then Postgres, MongoDB, Elasticsearch


angela-helios:
From the 2026-07-30 review call: the Playwright suite is green, but we want human eyes confirming it is not lying to us — real data, real searches, both search UIs.

### Setup
- Build `main` locally and start hfs with the default SQLite backend.
- Load data two ways, to exercise both ingestion paths:
  - a batch/transaction Bundle (~10 patients) POSTed to the server root;
  - `$bulk-submit` with [Synthea](https://github.com/synthetichealth/synthea)-generated NDJSON (`run_synthea -p 1000` — Doug has used it and can help).

### Sanity checklist (per backend)
- [ ] Search by name, by birthdate, and other common Patient parameters returns sensible results.
- [ ] The guided/NL search UI translates and the landed query runs correctly.
- [ ] The visual builder round-trips: conditions, OR values, modifiers, sort.
- [ ] Advanced features: `_include`, `_revinclude`, forward/reverse chaining (`_has`), multi-level chains.
- [ ] Result counts agree between the UI and direct API calls.

### Backends
- [ ] SQLite
- [ ] Postgres
- [ ] MongoDB
- [ ] Elasticsearch

Findings that look like bugs get filed as their own issues and linked here.

angela-helios:
## SQLite leg — done. Two real bugs found, which is the point of this exercise.

**Setup**: main built locally, fresh SQLite DB. Data: 27 Synthea patients generated with `synthea-with-dependencies.jar -p 25` (needs Java 17; a portable Temurin JDK works).

**Ingestion, both paths:**
- Batch/transaction: hospital + practitioner bundles (`batch`, 193+192 entries) then 10 patient bundles (`transaction`, 200–1086 entries each) POSTed to the root — all 200, ~5.9k resources.
- `$bulk-submit`: remaining 17 patients as per-type NDJSON (20 files, 11,286 resources) behind a local Bulk Export Manifest. Kickoff/status/poll flow works; ingestion hit **#457** (below).

[...]

PR #466, opened 2026-07-31
fix(bulk-submit): key entry results by their manifest output file

### LINKED ISSUE #457: bulk-submit: multi-file manifests collide in bulk_entry_results — the PK has no file discriminator, so every file after the first fails at its first line
Found during the #448 manual sanity pass (SQLite, main @ 2026-07-30), submitting a Bulk Export Manifest with one NDJSON file per resource type (20 files, 11,286 resources from 17 Synthea patients).

### What happens
The first file ingests fine (Patient, 17/17). Every subsequent file dies wholesale at its first entry:

```
failed to ingest file http://.../ImagingStudy.ndjson: internal error in sqlite:
Failed to store entry result: UNIQUE constraint failed:
bulk_entry_results.tenant_id, .submitter, .submission_id, .manifest_id, .line_number
```

### Why
`bulk_entry_results`'s primary key is `(tenant_id, submitter, submission_id, manifest_id, line_number)` — no file/output discriminator — so line 1 of the second file collides with line 1 of the first. Both the SQLite and Postgres schemas carry the same key shape.

A single-file manifest never trips it, which is presumably how it survived: the smoke tests submit one file.

### Fix sketch
Add the output file (url or index) to the table, the PK, the insert, and the FK-dependent aggregation queries (`countSeverity`, status output/outcome grouping already carry per-file `manifestUrl` — they need the per-file identity too), plus a migration for both backends.

Part of the #448 sanity pass. The remaining count checks after the failure: Patient 27 ✓ (10 transaction + 17 bulk), other bulk types partially or not ingested.

angela-helios:
Closes #457

Line numbers restart in every manifest output file, but `bulk_entry_results`'s primary key had no file discriminator — so in any multi-file manifest, only the first file ingested and every other died wholesale on a UNIQUE violation at its first entry (single-file smoke tests never trip it; the #448 Synthea submission with 20 files did, instantly).

- The worker names the file it is ingesting (`BulkProcessingOptions::file_url`, threaded per-file), both backends store it, and both primary keys include it.
- **SQLite**: schema v15 with a table rebuild (PKs are immutable there); pre-migration rows keep `''`.
- **Postgres**: in-place `ADD COLUMN IF NOT EXISTS` + a re-key guarded on the constraint's current shape, so it is idempotent.
- The insert is now an **upsert** on that key: the worker re-fetches a whole file after a transient failure, and the retry must overwrite its own earlier rows instead of colliding with them (found live when my test harness's file server dropped mid-ingest and the retry loop compounded the collisions).

Tests: new multi-file regression (two files, both starting at line 1, both must ingest and both results stored); persistence lib 759 green; rest `bulk_submit` 13 green. A full-scale e2e (17 Synthea patients, 20 NDJSON files, 11,286 resources through `$bulk-submit`) is running against this build — result to follow as a comment.


angela-helios:
Full-scale e2e against this build: the #448 Synthea submission — 20 NDJSON files, 11,286 resources, every file starting at line 1 — now ingests **completely**. The status manifest reports one output per file (AllergyIntolerance 26, CarePlan 54, Claim 1,073, Condition 491, DiagnosticReport 1,045, Observation 1,377, Procedure 1,906, …) with no error outcomes, and every resource type queries back with the expected counts. Before the fix the same submission died at the first line of the second file.

angela-helios:
The `terminology_integration::test_in_modifier_fails_open_on_hts_unavailable` failure is unrelated to this PR — it is the #449 runner-proxy signature hitting a third surface (the HTS client detours through the egress proxy on some pool runners and hangs instead of failing fast). Root-cause fix in #469; re-running the failed job meanwhile, which should pass on any runner without the proxy env.

smunini (approved):

Issue #476, opened 2026-07-31
ui: Batch/Transaction workspace — upload, preflight review, execute, results summary


angela-helios:
From the review call on Brett's Batch/Transaction designs, updated 2026-07-31 against the new Figma frames (`330:461` Upload, `330:2` Actions, `330:720` Bundle JSON, `330:1141` Response) — the designs cover the agreed flow; this issue is the implementation.

### Flow (per the frames)
1. **Upload** — drop/browse a Bundle JSON file.
2. **Preflight (Actions)** — request line (`POST [base] · Bundle·{type}·{n} entries`), one row per entry with its HTTP method chip and target URL, collapsible body per row (chevrons in the design), `Bundle JSON` tab with the raw payload, Cancel / Upload another / Execute. Entry methods are display-only.
3. **Execute** — POST to the server root.
4. **Response** — per-action outcomes with the status badges, `Execute again` / `Back to bundle`.

### Implementation requirements beyond the mock
- **Aggregate result summary** (Steve, not in the mock): the Response page states how many resources were created / updated / failed, not just per-row badges.
- **Batch-vs-transaction explainer**: one line of copy in the preflight stating the semantics of what is about to run — transaction: all-or-nothing; batch: independent entries, failures don't stop the rest.
- **Texts come from the real payload, not the mock**: the bundle type shown is the uploaded bundle's `type` (the mock mixes `batch` in the header with `transaction` in the JSON), and per-entry statuses render verbatim from the response bundle (`201 Created`, not the mock's `200 CREATED`). Flagged to Brett as example-text nits; implementation is the source of truth.
- **GET entries are in-scope**: per the R4 spec, transaction processing explicitly includes "Process any GET or HEAD interactions" (step 4), and the server already supports them (`BundleMethod::Get`).

### Out of scope (decided 2026-07-31)
- Single-resource upload path and a dedicated expanded-accordion mock (Edson's suggestions — dropped).
- Import ($bulk-submit) and Export remain separate nav entries and separate work.

PR #481, opened 2026-08-01
fix(rest): execute search-style GET bundle entries as searches

### LINKED ISSUE #478: batch/transaction: search-style GET entries are mis-parsed as instance reads
Found while building the Batch/Transaction workspace (#476). Bundle GET entries whose `request.url` is a **search** are treated as instance reads in both bundle types:

```
transaction entry {method: GET, url: "Patient?name=Nguyen"}
  → 400 "Transaction failed at entry N: Entry processing failed: invalid reference: Patient?name=Nguyen"

batch entry {method: GET, url: "Patient?name=GetById"}
  → entry response "404 Not Found" ("Resource not found") — the search URL parsed as a resource address
```

`GET Patient/{id}` works in both types (200 with the resource). Per the R4 transaction processing rules, GET entries cover both reads **and searches** — the entry URL is interpreted "as a request against the server's base", so `Patient?name=X` should execute the search and return a searchset in that entry's response.

The workspace UI (#476) renders whatever the server answers, so nothing blocks there — but a user uploading a legitimate search-in-bundle gets a confusing failure today.

angela-helios:
Closes #478

## The bug

Bundle GET entries were unconditionally parsed as `Type/id` reads. `GET Patient?name=Nguyen` became a read of resource type `Patient?name=Nguyen` — a 404 in batch bundles, and a 400 rejecting the **whole bundle** in transactions — even though the spec's processing rules allow any read *or search* URL in a GET entry (this is what blocked the #476 screenshot bundle and got the bug filed).

## The fix

- `parse_request_url` strips the query string before extracting type/id, so scope checks and reads see the real resource type.
- A GET entry addressing a type (`Patient?name=x`, or bare `Patient`) now runs through the **same search pipeline as the HTTP endpoint** (`execute_search_bundle`, factored out of the search handler): terminology expansion, `_list`/chain resolution, `_include`, paging clamps. The resulting searchset Bundle is embedded as the entry's resource with `200 OK`. Instance reads (`Patient/123`) are untouched.
- **Transactions:** the spec orders GETs after all writes, and a search cannot run inside the storage transaction, so search entries execute against the just-committed state — which also makes them see the bundle's own writes (tested: POST a patient + `GET Patient?family=Tran` in one transaction finds it). Their queries are validated up front, so a malformed search still rejects the whole bundle before anything executes; a post-commit execution failure surfaces as that entry's own error outcome rather than a misleading whole-bundle failure for writes that did commit. GET-by-id keeps running inside the storage transaction as before.

## Tests

New `search_entries` conformance tests: batch search entry returns a searchset, bare-type GET is a search, batch mixing read + search entries, transaction search seeing the bundle's own writes, transaction GET-by-id unchanged. Full helios-rest suite green (33 test binaries, incl. batch_conformance 44, search_integration 101, rest_conformance 84); clippy clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3584,6 +3584,7 @@ dependencies = [
  "mime",
  "p256 0.13.2",
  "p384",
+ "parking_lot",
  "rand 0.8.6",
  "regex",
  "reqwest",
--- a/crates/rest/Cargo.toml
+++ b/crates/rest/Cargo.toml
@@ -133,6 +133,12 @@ helios-observability = { path = "../observability" }
 # Key generation for the JWE round-trip tests
 rand = "0.8"
 
+# `SearchProvider::search_param_registry` returns a `parking_lot::RwLock`, and
+# the batch unit tests' mock storage has to satisfy that trait since #478
+# widened `process_batch`'s bound. Test-only: the crate's own code reaches the
+# registry through `helios-persistence` and never names the lock type.
+parking_lot = "0.12"
+
 # Temp files
 tempfile = "3"
 
--- a/crates/rest/src/handlers/batch.rs
+++ b/crates/rest/src/handlers/batch.rs
@@ -17,8 +17,8 @@ use helios_audit::{AuditAction, AuditCorrelation, AuditEventBuilder};
 use helios_auth::{FhirOperation, Principal, SmartScopePolicy};
 use helios_fhir::FhirVersion;
 use helios_persistence::core::{
-    BundleEntry, BundleEntryResult, BundleMethod, BundleProvider, ResourceStorage,
-    bundle_if_match_gate,
+    BundleEntry, BundleEntryResult, BundleMethod, BundleProvider, IncludeProvider, ResourceStorage,
+    RevincludeProvider, SearchProvider, bundle_if_match_gate,
 };
 use helios_persistence::error::{ResourceError, StorageError, TransactionError};
 use serde_json::Value;
@@ -60,7 +60,13 @@ pub async fn batch_handler<S>(
     request: Request,
 ) -> RestResult<Response>
 where
-    S: ResourceStorage + BundleProvider + Send + Sync,
+    S: ResourceStorage
+        + SearchProvider
+        + IncludeProvider
+        + RevincludeProvider
+        + BundleProvider
+        + Send
+        + Sync,
 {
     // Extract the Principal from request extensions (set by auth middleware).
     // If present, per-entry scope checks will be enforced.
@@ -251,7 +257,7 @@ async fn process_batch<S>(
     principal: Option<&Principal>,
 ) -> RestResult<Response>
 where
-    S: ResourceStorage + Send + Sync,
+    S: ResourceStorage + SearchProvider + IncludeProvider + RevincludeProvider + Send + Sync,
 {
     debug!(
         tenant = %tenant.tenant_id(),
@@ -371,7 +377,13 @@ async fn process_transaction<S>(
     principal: Option<&Principal>,
 ) -> RestResult<Response>
 where
-    S: ResourceStorage + BundleProvider + Send + Sync,
+    S: ResourceStorage
+        + SearchProvider
+        + IncludeProvider
+        + RevincludeProvider
+        + BundleProvider
+        + Send
+        + Sync,
 {
     debug!(
         tenant = %tenant.tenant_id(),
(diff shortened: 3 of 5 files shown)


aacruzgon:
Flagging a data-corruption interaction found while implementing #503 (PR #512) — it lands squarely
on this PR's `parse_request_url` change, and it is not visible from this diff alone.

### What this PR does

```rust
// A query string is not part of the type/id path (`Patient?name=x` is a
// search on Patient, not a read of a resource named `Patient?name=x`).
let path = url.split('?').next().unwrap_or(url);
```

That is correct, and it is exactly what #503 needs. The problem is what happens to the **write**
arms once the criteria are gone — this PR adds `parse_search_entry_url` for the GET path but does
not touch the `"PUT"` or `"DELETE"` arms.

### The mechanism

With the query stripped, `PUT Patient?identifier=x` parses to `("Patient", "")` and falls through to
`state.storage().create_or_update(tenant, "Patient", "", …)`. That does **not** reject:

- `sqlite/storage.rs:186-214` — on `Ok(None)` it does `obj.insert("id", Value::String(id.to_string()))`
  and calls `create`.
- `sqlite/storage.rs:95-99` — `create` does
  `resource.get("id").and_then(as_str).map(String::from).unwrap_or_else(|| Uuid::new_v4())`. The UUID
  fallback fires on `None`, **not** on `Some("")`.

So it writes a row whose id is the empty string, into the real `Patient` table. The next conditional
PUT of that type reads `Some(current)` and *overwrites it*. `DELETE Patient?identifier=x` then
targets that same empty-id row.

### Why it matters that it is this PR

Before the strip, the corruption is quarantined: the criteria ride along in the resource type, so the
row lands under `Patient?identifier=http:` — invisible to every legitimate search and to the type
routes. After the strip, it lands in `Patient`, where search can see it and where every conditional
PUT of that type collides on one row. Same bug, materially worse blast radius.

Verified empirically: with the empty-id guard disabled, a batch `PUT Patient` returns
**`201 Created`**, not a 4xx.

### Options

1. **Add the guard here** — `if id.is_empty() { return create_error_result(400, …) }` at the top of
   the `"PUT"` and `"DELETE"` arms. About 12 lines, keeps this PR self-contained and mergeable in any
   order.

   > **Correction (edited after posting):** those two guards alone are not sufficient.
   > `POST Patient?identifier=x` post-strip parses to `("Patient", "")`, and the `"POST"` arm never
   > reads `id` — it goes straight to `check_write` and then
   > `state.storage().create(tenant, "Patient", resource, …)`. So a conditional create becomes a
   > **real, unconditional create with the dedup criteria silently discarded**, reporting 201.
   > Pre-strip that entry landed under the bogus type `Patient?identifier=x`; post-strip it is a
   > silent duplicate in the real table. An `id.is_empty()` test cannot catch it, because an empty
   > id is legitimate for POST — the check has to key on the URL carrying a query instead. #512
   > handles this by refusing any non-GET entry whose type-level URL carries criteria.
2. **Sequence it** — land #512 first, after which this PR's strip becomes a no-op on rebase (#512
   rewrites `parse_request_url` to `split_once('?')` plus empty-segment filtering, and adds the
   guards). If this PR lands first, #512 should follow immediately.

Either is fine; option 2 needs no change here. Happy to do whichever you prefer — just flagging so
it is a decision rather than a surprise.

### Context

- #503 — the parser defect
- PR #512 — the fix, stacked on #501 → #489. Also carries the empty-id guards and a transaction-path
  refusal, and deliberately does **not** touch the GET arm so this PR's search work lands on an
  untouched dispatch path.
- #511 — resolving conditional interactions properly (this PR and #512 both only make them fail
  honestly).

aacruzgon:
## Rebased onto the batch stack

Heads-up @angela-helios — I force-pushed this branch and retargeted its base from `main` to `fix/504-batch-entry-issue-codes` (#516). **Your original is preserved** at [`backup/481-original-pre-restack`](https://github.com/HeliosSoftware/hfs/tree/backup/481-original-pre-restack) (`65f127a10`) if you want to diff against it.

Why: this PR and the #501→#489→#503→#502→#504 stack both touch `process_batch_entry`, and they conflict. Rather than leave it for whoever merged second, we took it — but the reconciliation needs `entry_failure`, which only exists after #516, so the only shape that keeps both diffs scoped was to put this branch on top of that stack. Your diff is unchanged in substance and still reads as 4 files.

**Resolution applied** (5 hunks in `batch.rs`, 1 keep-both append in `batch_conformance.rs`):

- **Both new failure sites now render through `entry_failure`.** They were written as `let (status, _, details) = e.client_response(); create_error_result(status.as_u16(), &details)` — the code-discard #504 deleted everywhere else. Left as-is they'd have made search entries the one path still answering `processing`.
- **`parse_request_url`'s query strip is dropped as redundant** — #503 landed the same fix, with the empty-id write guards that make it safe. #512 called this: *"if this lands first, #481's strip becomes a no-op on rebase."*
- **The GET arm keys off `BundleMethod::Get`** rather than the raw `"GET"` string, since #502 replaced the string matcher with an exhaustive enum match.
- **The rollback fan-out** keeps #504's status/code threading and gains your `.chain(&search_entries)`.
- **`DelayStorage` gains `SearchProvider`/`IncludeProvider`/`RevincludeProvider`** and `run_batch`'s bound widens to match `process_batch`'s — #505 added a unit-test harness that drives `process_batch` directly with a mock, so widening the bound here means the mock has to satisfy it. All `unimplemented!()`, matching how the mock's write methods already work. That adds `parking_lot` as a **dev**-dependency.

1114 tests pass (1109 on the base + your 5).

One gap worth flagging: your five tests are all happy-path `200 OK`, so neither new failure site had coverage. #518 stacks on this and pins the issue codes on both — including the transaction-arm one, which turns out to be the first *reachable* per-entry outcome on that arm.

Nothing here changes what this PR does; shout if you'd rather own the resolution yourself and I'll hand it back.

smunini (approved):

Issue #459, opened 2026-07-30
transaction: bundle references are stored unresolved — urn:uuid fullUrls are not rewritten and conditional references are never resolved


angela-helios:
Found by the #448 R4 parameter sweep (SQLite, main @ 2026-07-30), ingesting Synthea transaction bundles.

### 1. `urn:uuid` references survive into storage
Synthea bundles reference sibling entries by `fullUrl` (`urn:uuid:<id>`). After a successful transaction, the stored resources still carry the urn:
```
GET /Observation?_count=1 → .subject.reference == "urn:uuid:8712799b-3834-10cc-87b8-9474304f2d04"
```
Per the Bundle processing rules the server must rewrite references to bundle entries to the ids it assigned. Search mostly papers over it — the indexer evidently strips the prefix and the uuid happens to equal the assigned id, so `Observation?patient=<id>` works — but any client that *reads* the resource gets a dangling `urn:uuid`, and `Observation?subject=urn:uuid:<id>` (the literal in the body) matches nothing.

### 2. Conditional references are stored verbatim
```
GET /Immunization?_count=1 → .location.reference == "Location?identifier=https://github.com/synthetichealth/synthea|62163921-…"
```
Transaction processing is supposed to resolve conditional references against the server's data (or reject). Stored as-is they are unsearchable (`Immunization?location=…` can never match) and unresolvable by any client.

Both are the same subsystem (bundle reference resolution at commit); Synthea is a convenient repro because it uses both styles heavily. Practitioner/Organization-side references from the `batch` bundles show the same behavior.

Part of the #448 sanity pass.

angela-helios:
Scope correction after digging in: **facet 1 (urn:uuid rewriting) is a false alarm** — replaying both a minimal transaction and a real Synthea bundle against a clean server shows stored references correctly rewritten to `Type/id`. The dangling `urn:uuid` resources in my report came from the #448 $bulk-submit NDJSON, which I had built by extracting raw bundle-entry resources (urns and all) — test-data artifact, not a server bug. **Facet 2 (conditional references stored verbatim) is confirmed** and is what the fix addresses.

Issue #491, opened 2026-08-04
ci(inferno): the US Core gate ignores `skip` — postgres passed 25 of 544 tests and reported success for six weeks


aacruzgon:
> **Update 2026-08-04 (2) — still open; implementation deliberately deferred.** The gate weakness is unchanged, but ~37% of `us_core_v800` still skips on every backend and the numbers a baseline would be built from are about to move three more times (#495/#496 for postgres, #500 for s3-elasticsearch, #463 for #456). A table committed now would be stale on arrival. Plan: land those, take one clean 30-cell nightly, then set baselines from it. Current measurements and two implementation options in [this comment](https://github.com/HeliosSoftware/hfs/issues/491#issuecomment-5186687766).

> **Update 2026-08-04 — the underlying defects (#490, #494) are fixed; the gate weakness this issue describes is not.** Post-fix pass/skip counts for all six suites are in [this comment](https://github.com/HeliosSoftware/hfs/issues/491#issuecomment-5185569913) — v8.0.0 skips fall 511 → 202, so the blind spot is smaller but still live. The evidence below is left as recorded: it is an account of what six weekly runs actually reported.

## Summary

The Inferno US Core workflow fails a job only when a test reports `fail` or `error`. Inferno's third failure-ish verdict, **`skip`**, is invisible to the gate — and a test skips when a precondition search returns nothing, which is exactly what a broken server does.

The result: a backend can stop working almost entirely and CI still reports green.

## Evidence

In the 2026-08-02 nightly ([run 30740427391](https://github.com/HeliosSoftware/hfs/actions/runs/30740427391)), the `us_core_v800` **postgres** job passed **25 of 544** tests. Its own log says:

```
Failures (excluding 17 omitted tests): 0
All tests passed (with 17 omitted)
```

and the job is marked `success`. The other 511 tests skipped.

It is not a one-off. Postgres pass rates in that single run, latest result per test id:

| suite | pass | skip | pass rate | job |
|---|---:|---:|---:|---|
| us_core_v311 | 28 | 241 | 10.2% | success |
| us_core_v400 | 27 | 262 | 9.1% | success |
| us_core_v501 | 25 | 374 | 6.1% | success |
| us_core_v610 | 28 | 418 | 6.2% | success |
| us_core_v700 | 27 | 487 | 5.2% | success |
| us_core_v800 | 25 | 511 | **4.6%** | success |

[...]

PR #500, opened 2026-08-05
fix(s3): decline transaction Bundles S3 cannot honour, and bound bundle concurrency

### LINKED ISSUE #489: transaction: s3-elasticsearch advertises and accepts transaction Bundles it cannot honour — a timed-out bundle commits 466 of 473 entries behind a 408
> **Update 2026-08-04 (2) — root cause found; the "Suggested investigation" below is superseded.** Full analysis in [this comment](https://github.com/HeliosSoftware/hfs/issues/489#issuecomment-5186142379). In short: rollback *exists* but is unreachable — the HTTP `TimeoutLayer` **drops** the request future, so cancellation skips every error branch, and the compensation list is only populated after all writes resolve, so it is empty at that moment. The "writes entries serially" hypothesis is **refuted**: `join_all` fires all 473 writes at once, and that unbounded fan-out is what causes the timeout. Scope is S3 only — SQLite/PostgreSQL/MongoDB are genuinely atomic. Per the persistence README and design discussion #28, S3 is an archival tier and #28 already specifies the trait gate that `process_transaction` bypassed, so the fix is to stop advertising an atomicity this tier never had. One decision is open: the Inferno fixtures carry 2,058 inter-entry `urn:uuid` references, so plain `batch` cannot replace `transaction` without client-side resolution.

> **Update 2026-08-04:** the open question below — whether the aborted transaction left partial writes — has been answered by inspecting the CI bucket. It did: **466 of the 473 entries are durably committed**, with no rollback. The atomicity violation, not the 408, is the bug. See [this comment](https://github.com/HeliosSoftware/hfs/issues/489#issuecomment-5183448042) for the evidence. The original report below is left as written, except where marked.

## Summary

In the 2026-08-02 nightly [Inferno US Core Test Suite run](https://github.com/HeliosSoftware/hfs/actions/runs/30740427391), the **Inferno US Core v8.0.0 Tests (s3-elasticsearch)** job failed at the *Load Inferno test data into HFS* step. `POST /` of `crates/hfs/tests/inferno/uscore_bundle_patient_355.json` returned **HTTP 408** after ~30s:

```
Processing uscore_bundle_patient_355.json...
  Type: transaction bundle -> POST /
  FAILED (HTTP 408)
```

Request started `10:54:40.680`, failed `10:55:10.719` — **30.04s**, which is exactly our own request-timeout budget.

The 408 is emitted by HFS, not by an intermediary or the loader script: `crates/rest/src/lib.rs:1035` installs

```rust
.layer(TimeoutLayer::with_status_code(
    axum::http::StatusCode::REQUEST_TIMEOUT,
    std::time::Duration::from_secs(config.request_timeout),
));
```

with `request_timeout` defaulting to 30 (`crates/rest/src/config.rs:1296`, `HFS_REQUEST_TIMEOUT`). The CI job does not override it.

## Why this bundle

`uscore_bundle_patient_355.json` is the largest transaction bundle in the Inferno fixture set — **1.08 MB, 473 entries** (140 Observation, 129 Encounter, 31 Condition, 23 DiagnosticReport, 21 DocumentReference, 20 each Procedure/CareTeam/CarePlan). The next largest, `uscore_bundle_patient_85.json` (617 KB), succeeded in the same job in 5.5s; every other bundle finished in under 3s.

So we are writing 473 resources through the `s3-elasticsearch` composite backend — an S3 PUT plus an Elasticsearch index per resource — inside a single 30s HTTP request. That is close enough to the ceiling that ordinary S3 latency variance pushes it over.

The server log supports that reading. It is quiet for the whole request except one line, ~2.5s in:

```
2026-08-02T10:54:43.244028Z INFO aws_smithy_runtime_api::client::connection:
  Connection encountered an issue and should not be re-used. Marking it for closure
```

then nothing further — no completion, no error, no rollback record. A dropped S3 connection plus SDK retries plausibly consumed the remaining budget, but the log is at INFO with no per-request spans, so this is inference rather than a proven cause.

## Impact

Two separate problems.

**1. The suite is fragile at its largest fixture.** Only the `s3-elasticsearch` backend has failed this way so far; sqlite/postgres/mongodb legs load the same bundle comfortably. It is timing-dependent, so it will recur intermittently.

**2. A transaction was aborted mid-flight and its writes were kept.** ~~we do not know what state it left behind~~ — **confirmed 2026-08-04:** FHIR transactions are atomic, all-or-nothing. The timeout layer cancels the request at the HTTP boundary, and the composite backend does *not* unwind: 466 of the 473 entries, plus their `_history/1.json` versions, are durably present in S3 with no tombstone or compensating delete. The client was told the transaction failed while the server had committed 98.5% of it. This is a correctness bug independent of CI.

Worth noting the failure mode was benign *here* only by luck: the load step failed the job, so Inferno never ran against half-loaded data. Had the loader tolerated the error and continued, the suite would have tested against a partially-populated server and the failures would have looked like conformance bugs.

## Frequency

- 2026-08-02 (`30740427391`) — this failure.
- 2026-07-26 (`30195238202`) — v8.0.0 s3-elasticsearch **passed**; only the v5.0.1 legs failed (separate issue, fixed by syncing our Inferno fork).
- 2026-06-28 and 2026-07-05 runs had broad multi-suite failures that I have not attributed; they may or may not share this cause.

## Suggested investigation

- Confirm whether a timed-out transaction leaves partial writes on the `s3-elasticsearch` composite backend. This is the part that matters beyond CI — if it does, the 408 is a symptom and the atomicity gap is the real bug.
- Consider whether a 30s wall-clock cap is the right control for bundle writes at all, versus a per-entry or per-backend budget. A 473-entry transaction is a legitimate request, not abuse.
- Look at whether the composite backend writes entries serially; batching or bounded concurrency to S3 would move this well clear of the ceiling.
- Short term, raising `HFS_REQUEST_TIMEOUT` for the `s3-elasticsearch` CI legs would stop the bleeding, but it masks both questions above — prefer it only as a stopgap.

## Environment

- Run: [30740427391](https://github.com/HeliosSoftware/hfs/actions/runs/30740427391), job `91477303737`, `main` @ `ba8dac52`
- Backend: `HFS_STORAGE_BACKEND=s3-elasticsearch`, S3 prefix `ci/us_core_v800/`, Elasticsearch `hfs` index prefix
- FHIR R4, auth disabled, audit disabled

aacruzgon:
## Summary

An `s3-elasticsearch` deployment advertised FHIR `transaction` support in its CapabilityStatement, accepted transaction Bundles, and could honour neither. When the 30s request timeout fired mid-bundle, **466 of 473 entries were left durably committed** with no tombstone or compensating delete, while the client was told the transaction had failed (#489).

This stops claiming the guarantee rather than trying to build it. Two defects are fixed:

1. **A capability lie** — three places asserted transaction support without consulting the backend.
2. **An unbounded fan-out** — `join_all` issued one S3 PUT per entry all at once, which is what blew the timeout in the first place. Refusing transactions removes it outright: that fan-out lived in `process_transaction`, which this PR deletes.

Fixes #489.

## Why not implement atomicity instead

The S3 backend *has* a compensation log. It cannot run on the failure mode that matters, for two structural reasons:

- `TimeoutLayer` **drops** the handler future. Async cancellation stops the task at its await point without propagating an error, so every `Err`/`status >= 400` arm holding rollback logic is skipped by construction.
- The compensation list is only populated *after* `join_all` resolves, so at the moment of cancellation it is empty — there is nothing to undo with, even given a cancellation-safe unwind.

More decisively, the architecture already ruled on this. `crates/persistence/README.md` calls S3 "intentionally storage-focused … archive/history storage", and design discussion #28 places ACID on the relational tier. #28 even specifies the guard:

> The trait separation makes this distinction clear: code that requires atomicity takes `&dyn TransactionProvider`, while code that can tolerate partial failures takes `&dyn ResourceStorage`.

`process_transaction` lives on `BundleProvider`, which every backend implements — so S3 served transactions without ever satisfying the trait meant to gate them. Building a transaction log over an object store would engineer a guarantee this tier was deliberately not given.

### Express and enforce the capability

- `BundleProvider::supports_atomic_transactions()` — a **required** method, no default. A new backend must state its position rather than inherit one, because the wrong default is silent corruption in one direction and a needless refusal in the other. SQLite/PostgreSQL → `true` (real `TransactionProvider`); MongoDB → `true` (server-side session, so it unwinds even on a dropped connection); S3 → `false`; composite → delegates to its primary.
- S3 refuses **before the first write** with a new `TransactionError::AtomicityUnsupported` → **501 + `not-supported`**, matching the two sibling capability gaps already mapped that way. A total refusal means a retry finds the server as it left it; the old partial commit would double-write every POST entry, which carries server-assigned ids and so has no idempotency.

### Stop advertising what we can't do

`BackendCapability::Transactions` already encoded the truth per-backend (declared by SQLite/PostgreSQL/MongoDB, absent from S3) — nothing read it. `CompositeStorage::capabilities` inserted `SystemInteraction::Transaction` unconditionally, and the CapabilityStatement carried `{"code": "transaction"}` as a literal. Both now derive it. `batch` stays unconditional.

### The fan-out goes with the path

*Corrected.* An earlier revision of this description claimed `join_all` → `buffered(16)`. That is not what landed, and it is corrected here rather than quietly dropped.

`origin/main`'s `s3/bundle.rs` had two unbounded `join_all` sites: one in `process_transaction` (`:97`) and one in `process_batch` (`:160`). Deleting the transaction machinery removes the first outright — there is nothing left to bound. The second was never reachable (`BundleProvider::process_batch` had no production caller on any backend) and is deleted by #501, which this PR is now stacked on. Neither was ever bounded, and `.buffered(` appears nowhere in this branch.

Batch entry concurrency — the thing that actually governs an S3 client's wall clock, now that this PR routes those clients to `batch` — is #501's subject, not this one's.

### Keep the Inferno coverage

Every fixture is a transaction Bundle, so the s3-elasticsearch legs would now fail at the load step. Posting them as batches is not a straight swap — `uscore_bundle_patient_355.json` alone has 473 entries and **2,058 inter-entry `urn:uuid` references** that only a transaction resolves. A naive batch load would leave all of them dangling, manufacturing #459 for that backend.

`crates/hfs/tests/inferno/to_batch.jq` resolves them client-side, and `install.sh` probes `/metadata` — now truthful — to decide whether it's needed. The transform deliberately mirrors `resolve_bundle_references`, rewriting only an object's `reference` field and only for `urn:uuid:` values. Rewriting any matching string would corrupt an `Identifier.value` that merely looks like a uuid; `pdex_bundle_patient_999.json` contains exactly that case. POST becomes PUT against the derived id, which also makes loads idempotent.

### Remove the now-dead machinery

With the refusal in place, the compensation log, both rollback helpers, the three-phase transaction body and S3's private `resolve_bundle_references` are unreachable — `bundle.rs` drops from 337 lines to 53, with no `#[allow(dead_code)]` left. Deleted rather than kept: if the decision were ever reversed, a durable design would need a persisted intent journal that survives process death, not an in-process `Vec` a dropped future discards. Meanwhile it reads as a live atomicity guarantee, which is the misreading that produced this bug. SQLite/PostgreSQL/MongoDB keep their own copies of `resolve_bundle_references` — still live, since they support transactions.

## Testing

`cargo check --workspace --all-targets` clean across sqlite, postgres, mongodb, elasticsearch and s3. `cargo test -p helios-persistence -p helios-rest --features R4,sqlite,postgres,s3` — 20 test binaries, **0 failures**.

New tests:
- `transaction_bundle_is_refused_without_writing_anything` — the `count == 0` assertion is the point; a refusal that still wrote something would reproduce #489 in miniature.

Batch is deliberately *not* re-asserted at the S3 level. Two such tests were on this branch before the restack; both drove `BundleProvider::process_batch`, which #501 removed as unreachable, so batch has no S3-level entry point left to test. It is covered over HTTP in `helios-rest` (`batch_conformance.rs`, `batch_if_match.rs`) and end-to-end by the `s3-elasticsearch` Inferno leg.
- `atomicity_unsupported_maps_to_501_not_supported` — status, issue code, and that the message names the backend, points at `batch`, and states nothing was applied.

Removed the three `bundle_transaction_*` tests, which covered the now-unreachable path, leaving a note in their place.

### Verified live against a real s3-elasticsearch deployment

MinIO + Elasticsearch, `HFS_STORAGE_BACKEND=s3-elasticsearch`:

1. `/metadata` lists `batch`, `history-system`, `search-system` — **`transaction` is gone**.
2. `POST /` with a transaction Bundle → **501** with the actionable OperationOutcome; `Patient?_summary=count` confirms **nothing was written**.
3. `install.sh` **self-detected** ("Server does not advertise 'transaction'; rewriting bundles to batch") with no flag or configuration.
4. All 15 fixtures loaded — 8 Patients, 246 Observations, 51 Conditions, 194 Encounters, **identical to the transaction-path numbers on every other backend**.
5. References resolved — `Observation?patient=…` → 27, Condition → 2, Encounter → 1. Identical again.

### Inferno: `us_core_v800`

| backend | pass | rate | failures |
|---|---:|---:|---:|
| **s3-elasticsearch (this branch)** | **329/544** | **60.4%** | **9** |
| postgres (with #495 + #496) | 331/544 | 60.8% | 10 |
| sqlite | 326/544 | 59.9% | 10 |

**Zero s3-elasticsearch-only failures** — its 9 are a strict subset of postgres's 10. This backend has never produced a pass rate before; every nightly died at the load step, so 60.4% is the first number for it.

**Now confirmed in CI** by [run 31016157176](https://github.com/HeliosSoftware/hfs/actions/runs/31016157176) on this branch, once #501 landed beneath it: **30/30 jobs green**, s3-elasticsearch **329 pass / 204 skip / 9 fail / 2 error** at v8.0.0 — reproducing the 329 above exactly, and *identical at the test-id level* to sqlite-elasticsearch (0 tests passing only on one side). The two large fixtures that previously 408'd now load in 9.42s (473 entries) and 5.48s (267 entries).

Worth being precise about the earlier evidence: the table above came from a local MinIO deployment, where latency is negligible and the request timeout never fires. The prior CI run on this branch ([30973069238](https://github.com/HeliosSoftware/hfs/actions/runs/30973069238)) never reached Inferno at all — all six s3-elasticsearch jobs failed at "Load Inferno test data into HFS" and skipped every subsequent step. Sequential batch was the reason, which is why #501 had to land first.

The rewrite was also validated independently by loading both ways into sqlite and diffing: identical across 22 resource types and 7 reference searches.

## Notes

- **Stacked on #501 (#505), which must merge first.** This PR tells S3 clients to use `batch`, and that advice only holds once batch can carry the load — sequential batch 408s past ~90 entries on S3. The base branch is `fix/501-batch-entry-concurrency`; this PR's diff shows only its own changes.
- **Merge order is unconstrained relative to #495/#496/#498** — zero file overlap, and no semantic dependency either way.
- The stale `s3/tests.rs` comment claiming entries "now run concurrently (`buffered`)" — behaviour that never existed — is gone: it went with the tests it annotated during the restack onto #501.
- **Original note, for the record:** zero file overlap with #495/#496/#498, and no semantic dependency either way. Worth noting one interaction I checked: #498 prunes omit entries verified against sqlite and postgres, and this PR makes the s3-elasticsearch legs produce results for the first time — all three pruned entries (`data_absent_reason` ×2, `practitioner_address_test`) also **pass** on s3-elasticsearch, so the two are safe together in any order.
- **This does not reproduce the original timeout.** MinIO on localhost finishes the 473-entry bundle in seconds, so the run above validates the new path's correctness, not the old failure. Against real S3 the batch loader is still strictly sequential on this branch — #501 measures it at ~300 ms/entry, with a 408 past ~90 entries.
- **Behaviour change for any S3-primary deployment**: transaction Bundles now return 501 instead of partially succeeding. That is the intended fix, but it is a visible API change for anyone who was relying on the old behaviour — which could not have been giving them the semantics they asked for.
- The composite's `sync_bundle_results` still swallows secondary-sync errors with `warn!` only, so a clean write can leave S3 committed and Elasticsearch unindexed. Separate pre-existing gap, noted in #390's discussion, not addressed here.
- Clippy deferred to CI at the author's request; the CI-style invocation cannot pass on this toolchain because of pre-existing lints in generated and untouched files.


aacruzgon:
## Draft while blocked on #501 — and two corrections to this PR

Code review on this PR flagged two issues; an overnight full-matrix Inferno run against this branch ([30973069238](https://github.com/HeliosSoftware/hfs/actions/runs/30973069238)) confirmed both. Both were valid, and the numbers in the original description are no longer accurate.

### 1. The batch rewrite relocated stable resource ids — **fixed** (`c914754ce`)

`to_batch.jq` branched on `fullUrl` alone, so entries carrying both a `urn:uuid` fullUrl *and* a concrete `PUT Type/<id>` were moved onto uuid ids. That relocated `Patient/85`, `/355`, `/907`, `/908` and `/999` — the ids Inferno is configured to exercise, the ids the Group fixtures reference, and the target of 21 literal `"Patient/999"` references, which are not `urn:uuid` and so never followed.

It also diverged from the server, which was the premise of the whole transform: phase 1 of `process_transaction` skips non-POST entries, so a PUT keeps the id it names.

Chasing that turned up a second divergence the review did not mention: SQLite, PostgreSQL and MongoDB build the reference map **incrementally**, resolving each entry against the map so far, so they never resolve a *forward* reference. The original rewrite built the map up-front and did, giving s3-elasticsearch three references the other backends lack. Now a single incremental pass.

The consistent rule, applied in all three places: **the loader mirrors the server, including where the server does nothing.** Same reason references to PUT entries stay unresolved (#459) — resolving them would hand one backend better data and break comparability.

**Why the original validation missed it, which matters more than the bug.** The equivalence check compared aggregate counts, which cannot see relocation — 8 Patients either way. And the local Inferno run reported 60.4% with "zero backend-specific failures" while four of five test patients did not exist, because tests for a missing patient **skip** and the gate ignores skips (#491). The pass rate was never evidence the data was correct.

Now: the check compares full sorted id lists per type plus hashed reference values (identical across 23 types), and `install.sh` fetches all 136 fixture-addressed ids after loading. That guard was verified to *fail* on the old rewrite — `MISSING: Patient/pat015 (HTTP 404)` and eight more — not merely to pass on the new one.

**The 60.4% figure in the description above is void.** It was measured with the buggy transform.

### 2. `buffered(16)` was unreachable — **commit dropped**, real defect filed as #501

`BundleProvider::process_batch` has no caller reachable from HTTP; the REST handler implements batch itself with a sequential loop. So that commit could not execute, and claiming it fixed the #489 timeout was exactly the kind of unearned capability claim this PR exists to remove. It has been dropped from the branch.

The underlying defect is real and now measured. From the run above, loading fixtures as batch against real S3:

| bundle | entries | wall clock | result |
|---|---:|---:|---|
| `uscore_bundle_patient_355.json` | 473 | **30.1s** | **408** |
| `uscore_bundle_patient_85.json` | 267 | **30.1s** | **408** |
| all others | ≤54 | 0.2–15.8s | 200/201 |

~300ms per entry, consistently, so ~90 entries is the ceiling inside the 30s budget. `patient_85` loaded in 5.5s through the concurrent transaction path — sequential batch is a ~15× regression for the same data.

That is a pre-existing defect, not one this PR introduces; it was invisible because nothing large went through batch until this PR started routing S3 there. Filed as **#501**.

### Status

**Draft until #501 lands.** This PR declines transaction Bundles on S3 and points callers at `batch`; that advice is only sound once batch can carry the load. Merging now would turn the six s3-elasticsearch legs red at the load step.

Unaffected and still verified: the capability derivation, the refusal and its 501, the CapabilityStatement fix, the removal of the unreachable transaction machinery, and the tests. Every other leg of the overnight run — sqlite, sqlite-elasticsearch, postgres, mongodb, all six suites — passed.

Once #501 merges, this rebases and the s3-elasticsearch legs should produce a trustworthy pass rate for that backend for the first time.

smunini (approved):

smunini (approved):

smunini (approved):

smunini (approved):

PR #509, opened 2026-08-05
perf(fhir-codegen): share generated serde/FHIRPath bodies across FHIR types (1.53 GiB → 1.02 GiB)

### LINKED ISSUE #508: fhir codegen: serde deserializer derives on generated choice-type enums bloat dependent binaries — helios-rest --lib links to 1.53 GiB
## Summary

The serde `Deserialize` derives emitted for the generated FHIR models dominate every binary that deserializes typed resources. The `helios-rest --lib` test binary links to **1.53 GiB** with debuginfo already off — 4× `helios-persistence` and 38× the `hfs` binary. **72% of its symbol bytes are the four FHIR versions, and 599.6 MB of that is deserializer glue alone.**

This is a build-time and link-time tax paid on every platform and every build. It is also the suspected trigger of a macOS-only CI failure (see below), though that link is not yet confirmed.

## Measurements

Built at `053b4a374`, `rustc 1.95.0`, `CARGO_PROFILE_DEV_DEBUG=0` (mirroring the workflow-level env in `ci.yml`, so these figures carry **no** debuginfo):

```
cargo test -p helios-rest --lib --all-features --no-run
```

| test binary | size | ratio |
|---|---:|---:|
| `helios_rest` --lib | **1,642,581,296 B (1.53 GiB)** | 1.0× |
| `helios_persistence` --lib | 406,630,512 B (0.38 GiB) | 0.25× |
| `hfs` --bins | 42,694,600 B (0.04 GiB) | 0.03× |

Inside `helios_rest`: `.text` is 1.09 GiB of 1.35 GiB in sections, across **1,706,009 symbols**.

Symbol bytes by origin (1108.2 MB attributed):

| group | size | share |
|---|---:|---:|
| `helios_fhir::r4` | 236.9 MB | 21.4% |
| `helios_fhir::r5` | 211.9 MB | 19.1% |
| `helios_fhir::r6` | 179.9 MB | 16.2% |
| `helios_fhir::r4b` | 173.6 MB | 15.7% |
| third-party + core | 189.0 MB | 17.1% |
| other helios crates | 116.1 MB | 10.5% |

Within those 802.3 MB of FHIR-model symbols:

| path | size |
|---|---:|
| deserialize | **599.6 MB** |
| serialize | 128.9 MB |
| everything else | 73.8 MB |

## Mechanism

The largest individual symbols are all the same shape, at ~0.22 MB **each**:

```
<<helios_fhir::r5::ParametersParameterValue as Deserialize>::deserialize::EnumVisitor as Visitor>::visit_map
<<helios_fhir::r6::ElementDefinitionFixed    as Deserialize>::deserialize::EnumVisitor as Visitor>::visit_map
<<helios_fhir::r6::ExtensionValue            as Deserialize>::deserialize::EnumVisitor as Visitor>::visit_map
```

Choice types (`value[x]`, `ElementDefinitionFixed`, `ParametersParameterValue`, …) carry one variant per permitted FHIR datatype — dozens each. serde's derive expands `visit_map` into a match over every variant, so each enum costs a fraction of a megabyte. Multiply by hundreds of such enums × 4 FHIR versions and you get the 599.6 MB.

It is not one pathological generic; it is a multiplication.

This also explains why `helios-rest` is the outlier while `helios-persistence` links the same models at a quarter the size: persistence largely moves resources as `serde_json::Value` and never instantiates most of the derives, whereas `helios-rest` serializes and deserializes typed resources on every route, forcing the lot.

## Suspected CI impact (unconfirmed)

`Test Rust` on the macOS/ARM64 agent aborts in dyld before `main` when it reaches this binary — run [31023273581](https://github.com/HeliosSoftware/hfs/actions/runs/31023273581/job/92365164525):

```
Running unittests src/lib.rs (target/debug/deps/helios_rest-2bcca06fdfde1c19)
dyld[30141]: dyld cache '(null)' not loaded: syscall to map cache into shared region failed
dyld[30141]: Library not loaded: /System/Library/Frameworks/SystemConfiguration.framework/...
process didn't exit successfully (signal: 6, SIGABRT)
```

On Apple Silicon the main executable loads at `0x100000000` and the dyld shared cache is mapped above it, leaving ~2 GiB for the executable's segments. A Mach-O that runs past that leaves dyld nowhere to map the cache, which produces exactly this signature. The Linux figure (1.53 GiB) is under that ceiling, but Mach-O layout and workspace-wide feature unification both push upward. Every smaller test binary in the same job loaded and ran fine.

Note that `ci.yml` currently attributes this signature to `-undefined dynamic_lookup` in rustflags; that run carried no rustflags at all, so that comment is not the whole story.

A probe measuring the actual Mach-O is in flight on branch `ci/mac-docker-probe`. **If it shows the binary loads fine by hand, the macOS link here is wrong** — but the bloat stands on its own regardless.

## Direction

Emit a shared deserializer for choice-type enums in `helios-fhir-gen` — one implementation keyed on the field-name suffix — instead of relying on serde's per-enum derive expansion. Hundreds of near-identical `visit_map` bodies collapse to one.

Two non-options, for the record:

- **Don't drop FHIR versions from `--all-features` on `Test Rust`.** `HFS_REQUIRE_R5` exists precisely so the R5 conformance suite fails loudly rather than compiling to its `#[cfg(not(R5))]` no-op twin (#307). Trading that away to fit a macOS ceiling is the wrong trade.
- **Don't reach for `-Wl,-pagezero_size`** to move the load address. `ci.yml` already records why `-stack_size` had to be abandoned: cargo applies `RUSTFLAGS` to proc-macro and cdylib links, and ld64 rejects both flags on anything that is not a main executable, breaking the build outright.

## Reproduction

```bash
export CARGO_PROFILE_DEV_DEBUG=0
cargo test -p helios-rest --lib --all-features --no-run
ls -l target/debug/deps/helios_rest-*

BIN=$(ls target/debug/deps/helios_rest-* | grep -v '\.d$' | head -1)
size -A "$BIN" | sort -k2 -nr | head -5
nm -S --size-sort -C "$BIN" | tail -25
```

smunini:
Fixes #508.

## Result

`helios-rest --lib`, same recipe as the issue (`CARGO_PROFILE_DEV_DEBUG=0`, `rustc 1.95.0`, Linux):

| | size | vs baseline |
|---|---:|---:|
| baseline (reproduced exactly) | 1,642,581,296 B (1.53 GiB) | — |
| choice-enum deserializer shared | 1,582,117,064 B | −60 MB |
| + struct constructor hoisted | 1,125,974,616 B | −517 MB |
| + choice-enum FHIRPath shared | **1,095,621,400 B (1.02 GiB)** | **−547 MB (−33.3%)** |

FHIR-model symbol bytes: 802.3 MB → 433.5 MB.

## The issue's stated direction accounts for 60 MB of it

I implemented it first, exactly as written. The per-enum `visit_map` key scan now lives in `helios_serde_support::deserialize_choice_parts`, generic only over the `MapAccess` implementation and keyed on a `&'static [&'static str]`, so it is emitted once per deserializer instead of once per enum. Dispatch became a jump table over the matched variant index rather than a string match, and the per-variant arms delegate to shared helpers — which also means the primitive type no longer has to be spelled out, so a chunk of fragile type-extraction logic (and four `panic!`s) came out of the macro.

Choice-enum `EnumVisitor` symbols went from ~70 MB to 9.4 MB, and those `visit_map` bodies are no longer the largest symbols in the binary.

But that is not where the 599.6 MB was.

## What actually dominated

Measuring the remainder showed a different multiplication: **every `<T as Deserialize>::deserialize` was landing in the binary 13–23 times.**

```
distinct names: 3709   total occurrences: 31177   bytes: 353.1 MB
copies-per-name: 2-4 for ~1850 types, 13-23 for ~1850 types
one copy each would be ~45.9 MB
```

The multiplier is the `D` type parameter — serde_json's reader and `Value`, serde's `MapAccessDeserializer` / `ContentDeserializer` wrappers, the XML deserializer — crossed with each crate in the graph that instantiates them. `Coding` had 23 copies in 11 distinct sizes; `Measure`, only ever deserialized as a top-level resource, had 3.

The fat part of that body is the `field` / `_field` reunification into `Element`, and it never touches `D`. It just happened to live inside `fn deserialize<D>`. Emitting it as a nested non-generic `fn __from_temp` — items in a function body do not inherit the enclosing generics — collapses 353.6 MB to 64.5 MB. That is the 456 MB step.

I then applied the same treatment to the `FhirPath` derive, whose choice-enum `to_evaluation_result` had by then become the largest individual symbol (~0.21 MB each): the identical per-variant repetition #508 describes, in a different derive.

## Changes

- `crates/serde-support/src/lib.rs` — `deserialize_choice_parts`, `deserialize_choice_element_parts`, `deserialize_choice_value`
- `crates/fhirpath-support/src/lib.rs` — `with_choice_type_info`, `wrap_choice_field`
- `crates/fhir-macro/src/lib.rs` — call the above; index-keyed dispatch; hoist the struct constructor

`serde_json::Error` is the intermediate error type in the hoisted constructor. It implements `serde::de::Error`, so error text is unchanged; all message wording is preserved throughout (no test asserted on it, but the strings are identical anyway).

## Verification

All green:

- `helios-fhir` (all features)
- `helios-serde` — including `test_examples` round-trips over the FHIR spec examples, and the `xml` feature paths
- `helios-fhirpath` + `helios-sof` — 989 tests, 98 binaries
- `helios-rest --lib` — 483 tests
- `helios-fhir-macro`, `helios-serde-support`, `helios-fhirpath-support`, `helios-fhir-validator`
- `cargo fmt --all --check` clean
- `cargo clippy --all-targets --all-features` clean with the CI flag set

## Pushing further

Tracked in #510.

The remaining FHIR-model bytes are measured and the next levers are specific, but each is a bigger change than this one, so I stopped here rather than commit the project to them:

| remaining | size | symbols |
|---|---:|---:|
| serde's own `#[derive(Deserialize)]` on the generated `Temp*` structs | 165.5 MB | 72,822 |
| `Serialize` | 135.1 MB | 34,843 |
| hoisted `__from_temp` | 51.3 MB | 26,698 |
| outer generic `deserialize` wrappers | 13.2 MB | 37,797 |
| choice `EnumVisitor` | 9.4 MB | 1,930 |

Both leaders are still multiplied by `D` / `S` the same way the constructor was — but unlike the constructor, neither can simply be hoisted, because both genuinely consume the generic parameter. Erasing them would mean something like `erased-serde` at the type boundary: a real architectural change with runtime dispatch cost and error-type consequences. That is #510, which carries the full breakdown, the benchmarking that would need to happen first, and the decisions it depends on.

Two cheaper observations from the same data, both also captured in #510:

- The hoisted `__from_temp` is non-generic yet still appears ~7× per type. Worth understanding before assuming a single copy is achievable elsewhere — it suggests nested items in generic functions are not deduplicated across crates as cleanly as the rule implies.
- The 13–23× spread correlates with *nesting depth*, not resource size. Types deserialized only at the top level (`Measure`, `Patient`) have 3 copies; types deserialized as fields inside many containers (`Coding`, `Extension`) have 22–23. Reducing the number of distinct `Deserializer` types the workspace routes FHIR JSON through would shrink the multiplier without touching the derive at all.

## Notes

- Not addressed here: #508 observes that `ci.yml` attributes the macOS dyld signature to `-undefined dynamic_lookup` in rustflags, and that the failing run carried no rustflags. That comment is still inaccurate; left alone since it is a separate concern.
- The macOS link remains **unconfirmed**. The binary is now well under the ~2 GiB Apple Silicon ceiling, but I have no Apple Silicon evidence either way. The bloat stands on its own regardless, as the issue says.
- The build script refreshed 1,602 R6 spec fixtures under `crates/fhir/tests/data/json/R6/` during these builds (the 24-hour auto-download). Deliberately **not** included in this branch.

https://claude.ai/code/session_012scJWEd3vbUsJMWBqWAa8n

### DIFF
--- a/crates/fhir-macro/src/lib.rs
+++ b/crates/fhir-macro/src/lib.rs
@@ -1677,7 +1677,14 @@ fn generate_deserialize_impl(data: &Data, name: &Ident) -> proc_macro2::TokenStr
                 let variant_key = rename.unwrap_or_else(|| variant_name_str.clone());
                 variant_names.push(variant_key.clone()); // Keep track of expected keys
 
-                // Generate the specific deserialization logic for this variant
+                // Generate the specific deserialization logic for this variant.
+                //
+                // Every arm below delegates the actual work to a shared helper in
+                // `helios-serde-support` rather than expanding it inline. Choice enums carry
+                // one variant per permitted FHIR datatype (dozens each) and there are hundreds
+                // of such enums per FHIR version, so inlining the extension-reunification and
+                // error-formatting logic multiplies out into hundreds of megabytes of
+                // near-identical code in any binary that deserializes typed resources.
                 let deserialization_logic = match &variant.fields {
                     Fields::Unnamed(fields) if fields.unnamed.len() == 1 => {
                         // Newtype variant (e.g., String(String))
@@ -1687,100 +1694,28 @@ fn generate_deserialize_impl(data: &Data, name: &Ident) -> proc_macro2::TokenStr
 
                         if is_element || is_decimal_element {
                             // --- Element/DecimalElement Variant Construction ---
-                            let underscore_variant_key_str = format!("_{}", variant_key); // For error messages
-
-                            // Determine the primitive type V or PreciseDecimal for the value field
-                            let primitive_type_for_element = if is_decimal_element {
-                                quote! { crate::PreciseDecimal }
-                            } else {
-                                // Extract V from Element<V, E> or the alias's underlying primitive
-                                // Need to re-determine the base type here
-                                let base_type = get_base_type(field_ty);
-                                if let Type::Path(type_path) = base_type {
-                                    if let Some(last_segment) = type_path.path.segments.last() {
-                                        if last_segment.ident == "Element" {
-                                            // Direct Element<V, E>
-                                            if let PathArguments::AngleBracketed(generics) =
-                                                &last_segment.arguments
-                                            {
-                                                if let Some(GenericArgument::Type(inner_v_type)) =
-                                                    generics.args.first()
-                                                {
-                                                    quote! { #inner_v_type }
-                                                } else {
-                                                    panic!("Element missing generic argument V");
-                                                }
-                                            } else {
-                                                panic!("Element missing angle bracketed arguments");
-                                            }
-                                        } else {
-                                            // Alias
-                                            let alias_name = last_segment.ident.to_string();
-                                            let primitive_type_str =
-                                                extract_inner_element_type(&alias_name);
-                                            let primitive_type_parsed: Type = syn::parse_str(
-                                                primitive_type_str,
-                                            )
-                                            .expect("Failed to parse primitive type string");
-                                            quote! { #primitive_type_parsed }
-                                        }
-                                    } else {
-                                        panic!("Could not get last segment of Element type path");
-                                    }
-                                } else {
-                                    panic!("Element type is not a Type::Path");
-                                }
-                            };
-
+                            // `V` and `E` are inferred from the assignments into `element`,
+                            // so the primitive type never has to be spelled out here.
                             quote! {
-                                // Check if parts exist *before* potentially moving them
-                                let has_value_part = value_part.is_some();
-                                let has_extension_part = extension_part.is_some();
-
-                                // Deserialize the extension part if present
-                                let mut ext_helper_opt: Option<IdAndExtensionHelper> = None;
-                                if let Some(ext_value) = extension_part { // Move happens here
-                                    ext_helper_opt = Some(serde::Deserialize::deserialize(ext_value)
-                                        .map_err(|e| serde::de::Error::custom(format!("Error deserializing extension {}: {}", #underscore_variant_key_str, e)))?);
-                                }
-
-                                // Deserialize the value part if present, consuming value_part
-                                let deserialized_value_opt = if let Some(prim_value) = value_part { // Move of value_part happens here
-                                    // Use #primitive_type_for_element determined outside
-                                    Some(<#primitive_type_for_element>::deserialize(prim_value)
-                                         .map_err(|e| serde::de::Error::custom(format!("Error deserializing primitive {}: {}", #variant_key, e)))?)
-                                } else {
-                                    None::<#primitive_type_for_element> // Explicit type needed for None
-                                };
-
-                                // Construct the element using deserialized parts
-                                let mut element: #field_ty = Default::default(); // Start with default
-
-                                // Assign deserialized value
-                                element.value = deserialized_value_opt; // Assign the Option<V> or Option<PreciseDecimal>
-
-                                // Merge the extension data if it exists
-                                if let Some(ext_helper) = ext_helper_opt {
-                                    if ext_helper.id.is_some() {
-                                        element.id = ext_helper.id;
-                                    }
-                                    if ext_helper.extension.is_some() {
-                                        element.extension = ext_helper.extension;
-                                    }
-                                }
-                                // Note: The check `if !has_value_part && has_extension_part { element.value = None; }`
-                                // is now redundant because element.value is already None if !has_value_part.
-
+                                let (__value, __id, __extension) =
+                                    ::helios_serde_support::deserialize_choice_element_parts::<_, _, A::Error>(
+                                        value_part, extension_part, #variant_key,
+                                    )?;
+                                let mut element: #field_ty = Default::default();
+                                element.value = __value;
+                                element.id = __id;
+                                element.extension = __extension;
                                 Ok(#name::#variant_name(element))
                             }
                             // --- End Element/DecimalElement Variant Construction ---
                         } else {
                             // --- Regular Newtype Variant Construction ---
                             quote! {
-                                let value = value_part.ok_or_else(|| serde::de::Error::missing_field(#variant_key))?;
-                                let inner_value = serde::Deserialize::deserialize(value)
-                                    .map_err(|e| serde::de::Error::custom(format!("Error deserializing non-element variant {}: {}", #variant_key, e)))?;
-                                Ok(#name::#variant_name(inner_value)) // Removed .into()
+                                Ok(#name::#variant_name(
+                                    ::helios_serde_support::deserialize_choice_value::<_, A::Error>(
+                                        value_part, #variant_key, "non-element variant",
+                                    )?
+                                ))
                             }
                             // --- End Regular Newtype Variant Construction ---
                         }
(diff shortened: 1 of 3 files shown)


smunini:
## Verified on the macOS agent: this also fixes the `Test Rust` dyld abort

`Test Rust` on `github-agent5` (macOS 26.6 / ARM64) has been aborting before `main` when it reaches `helios-rest --lib` — run [31023273581](https://github.com/HeliosSoftware/hfs/actions/runs/31023273581/job/92365164525):

```
dyld: dyld cache '(null)' not loaded: syscall to map cache into shared region failed
dyld: Library not loaded: .../SystemConfiguration.framework
process didn't exit successfully (signal: 6, SIGABRT)
```

**Cause.** On arm64 the executable loads at `0x100000000` and the dyld shared cache is mapped at `0x180000000`. The failing Mach-O placed `__LINKEDIT` — the symbol table, not the code — from `0x13db50000` to `0x183900000`, overrunning the cache base by 57 MB, so dyld had nowhere to map the cache and every system framework failed to resolve. `__TEXT` ended at `0x13d52c000`, nowhere near the boundary; with ~1.7 M symbols the symbol table was larger than the code.

**Measured across every `helios_rest` build sitting in that runner's target dir:**

| binary | bytes | `__LINKEDIT` top | vs `0x180000000` | loads? |
|---|---:|---|---|---|
| `2bcca06fdfde1c19` (post-#509) | 1,351,231,440 | `0x1508a4000` | under | **LOADED** |
| `307eac708a869c91` (post-#509) | 1,270,049,216 | `0x14bb38000` | under | **LOADED** |
| `6f7e33fda7020137` | 1,400,360,576 | `0x153780000` | under | **LOADED** |
| `4b5a1c7aef68d5e4` | 2,203,686,032 | `0x18359c000` | **OVER** | ABORTED rc=134 |
| `98cc99847ade0006` | 2,204,740,736 | `0x18369c000` | **OVER** | ABORTED rc=134 |
| `d6b3b8133ccc2d89` | 2,207,361,360 | `0x18391c000` | **OVER** | ABORTED rc=134 |
| `fae956f89ebef24a` | 2,204,675,008 | `0x18368c000` | **OVER** | ABORTED rc=134 |

7/7 — over the base aborts, under it loads, no exceptions.

`2bcca06fdfde1c19` is the exact path that aborted in the CI run above. Before this PR it was 2,207,244,816 B topping at `0x183900000`; rebuilt with this PR merged it is 1,351,231,440 B topping at `0x1508a4000`, and it loads. Same artifact, both sides of the boundary.

**Size reduction reproduced independently** on Linux, identical recipe to #508 (`CARGO_PROFILE_DEV_DEBUG=0`, rustc 1.95.0, `crates/fhir/src/` byte-identical between builds): 1,642,581,296 B → 1,095,621,400 B, **−546,959,896 B (−33.3%)**. Byte-for-byte identical to this PR's reported figure.

### Two caveats worth recording

**Headroom is finite.** The pre-fix overrun was only 57 MB — 3% past the line. Post-fix there is ~800 MB of headroom, but nothing prevents this from silently returning as the FHIR models grow, and the failure mode is an opaque dyld abort that reads like anything but a size problem. A cheap guard (assert `__LINKEDIT` top stays under `0x180000000`, or just cap the binary size) would keep the next occurrence from costing another investigation.

**Symbol count is the metric that matters here, not bytes.** Because it is `__LINKEDIT` that overruns, what protects the macOS build is fewer symbols rather than less code. This PR's win comes largely from removing duplicate `<T as Deserialize>::deserialize` instantiations, which cuts symbols directly — so it happens to attack exactly the right quantity. Future size work aimed only at `__TEXT` would not necessarily buy macOS headroom.

Verification workflow is on `ci/mac-docker-probe` (temporary, to be deleted).

PR #520, opened 2026-08-06
test(ui-e2e): wait for search indexing before ES-sensitive asserts

sandhums:
## Summary
- Add `waitForSearchHit` helper for composite backends where create ACKs before Elasticsearch indexes.
- Use it in SearchParameter CRUD and the chained/`_has` query builder test that fail nightly on `sqlite-elasticsearch` / `s3-elasticsearch`.
- Poll detail-action href and result-row count with a longer timeout so registry/UI swaps can finish.

## Test plan
- [ ] Re-run **UI Tests (backend matrix)** workflow (or wait for next nightly)
- [ ] Confirm `sqlite-elasticsearch` and `s3-elasticsearch` pass SearchParameter CRUD + chain/`_has` query
- [ ] Confirm sqlite/postgres/mongodb matrix jobs still green

Made with [Cursor](https://cursor.com)

### DIFF
--- a/crates/ui/e2e/pages/api.ts
+++ b/crates/ui/e2e/pages/api.ts
@@ -46,6 +46,46 @@ export async function readResource(
 }
 
 /**
+ * Wait until a just-created resource is visible to FHIR search.
+ *
+ * On composite backends (sqlite/s3 + Elasticsearch) create returns as soon as
+ * the write store acknowledges; the search index can lag a few hundred ms.
+ * Nightly ui-tests-matrix flakes when the next step searches immediately.
+ */
+export async function waitForSearchHit(
+  request: APIRequestContext,
+  type: string,
+  query: string,
+  opts: { timeoutMs?: number } = {},
+): Promise<void> {
+  const timeoutMs = opts.timeoutMs ?? 15_000;
+  const deadline = Date.now() + timeoutMs;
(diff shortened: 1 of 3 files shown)


smunini:
Closing as superseded by #593, which landed on `main` and fixes the same issue — the nightly `ui-tests-matrix` reds on the Elasticsearch legs, caused by specs that create a resource and immediately search for it while the ES composites are still indexing asynchronously.

#593 takes the same approach (a bounded `waitSearchable` poll in `e2e/pages/api.ts`, applied at the same call sites in `queries.spec.ts` and `search-parameters.spec.ts`) and additionally:

- exports `HFS_COMPOSITE_SYNC_MODE=synchronous` in the workflow so the write is in ES before the API responds, leaving only the index-refresh window for the poll to cover;
- waits on **every** created resource in the chained-search and paging specs, not just the last (indices are per resource type and refresh on their own ticks);
- covers `resources.spec.ts` and `editor-sync.spec.ts` as well.

Thanks @sandhums — this diagnosed the race correctly. This branch also has conflicts with `main` now, so rebasing it would mean resolving against the fix that already landed. If the nightly still flakes on the two assertions relaxed here (the `.detail__actions` href poll and the chained-query row count), those are worth cherry-picking at that point.

Issue #474, opened 2026-07-31
search: _tag, _profile and _security filters are silently dropped — even under Prefer: handling=strict — and the self link claims they were applied


angela-helios:
Found by #448 sweep iteration 3 (SQLite, Synthea data, 27 patients).

```
GET /Patient?_tag=http://example.org/none       → 200, total 27 (all patients)
GET /Patient?_profile=http://example.org/none   → 200, total 27
GET /Patient?_security=http://example.org/none  → 200, total 27
GET /Patient?_tag=… with Prefer: handling=strict → still 200, still 27
```

Two problems compounding:
1. The three meta-level token parameters are not implemented, and the search drops them silently instead of applying them — the response returns **everything**. For `_security` (labels used in access-control workflows) a silently unapplied filter is the worst possible failure mode.
2. The searchset's **self link echoes the parameter back** (`…/Patient?_tag=http%3A%2F%2Fexample.org%2Fnone`), which per the spec is the server's statement of which parameters it actually processed — so the client is actively told the filter worked.

Per FHIR search, unsupported parameters may be ignored **only in lenient handling and only if reported** (dropped from the self link, ideally with an OperationOutcome); under `Prefer: handling=strict` they SHALL be an error. Either implement the three (they are ordinary token indexes over `meta.tag` / `meta.profile` / `meta.security`) or reject/report them honestly.

Part of the #448 sanity pass, iteration 3.

PR #534, opened 2026-08-10
fix(search): index _source instead of evaluating Resource.meta.source

### LINKED ISSUE #523: search: `_source` is never indexed, so `_source` searches match nothing on PG/Mongo/ES (and everything on SQLite)
`_source` is never indexed on any backend, so `Patient?_source=…` silently returns the wrong set: **0 matches on PostgreSQL / MongoDB / Elasticsearch even when a resource does carry that `meta.source`**, and **every resource on SQLite** (SQLite drops the parameter — that part is #474).

Found while checking whether #474 also affects the non-SQLite backends. It does not; this is a separate, all-backends defect.

## Root cause

`_source` is the only `Resource`-level search parameter with no embedded definition, so it keeps the spec-file expression verbatim: `Resource.meta.source`. The extractor's FHIRPath evaluation returns nothing for a `Resource.`-prefixed path evaluated against a concrete resource, so no index row is ever written.

Probe against `crates/persistence/src/search/extractor.rs` (`SearchParameterExtractor::extract_for_param`) with a Patient carrying `meta.source = "http://example.org/src"`:

```
EXPR Resource.meta.source  -> []
EXPR Patient.meta.source   -> [Uri("http://example.org/src")]
EXPR meta.source           -> [Uri("http://example.org/src")]
EXPR Resource.meta.profile -> []            <-- same failure, masked (see below)
```

The registry's active `Resource` params show why the siblings survive — every one of them except `_source` is shadowed by an embedded definition whose expression has no `Resource.` prefix:

```
("_lastUpdated", "meta.lastUpdated", Embedded)
("_id",          "id",              Embedded)
("_profile",     "meta.profile",    Embedded)
("_security",    "meta.security",   Embedded)
("_tag",         "meta.tag",        Embedded)
("_source",      "Resource.meta.source", Embedded)   <-- from the spec file, unshadowed
```

`crates/persistence/src/search/extractor.rs:331` (`filter_expression_for_resource`) keeps only the `|`-alternatives whose prefix matches the concrete resource type, and falls back to the *original* expression when none match — which is exactly the `Resource.meta.source` case.

## Observed behaviour

Seeded one Patient with `meta.source = "http://example.org/src"` and one without:

| Backend | `_source=http://example.org/none` | `_source=http://example.org/src` |
|---|---|---|
| SQLite | 2 (all) | 2 (all) |
| PostgreSQL | 0 | **0** (expected 1) |
| MongoDB / Elasticsearch | untested, same index path as PG | expected 0 |

`_source` is advertised in the CapabilityStatement (`crates/rest/src/handlers/capabilities.rs`), so clients are told it works.

## Suggested fix

Either strip a leading `Resource.` in `filter_expression_for_resource` when the expression's prefix is the abstract `Resource` type (fixes `_source` and any other unshadowed `Resource`-base parameter, present and future), or add an embedded `_source` definition with expression `meta.source` alongside the other four. The first is the real fix; the second only papers over this one parameter.

Add coverage that asserts a *positive* match for each meta-level parameter — the current suite has no `_tag`/`_profile`/`_security`/`_source` search tests at all, which is why both this and #474 went unnoticed.

smunini:
Fixes #523 — the extraction half of the meta-parameter story. #474 fixed the query half on SQLite; this is why that fix's test file covers `_tag`/`_profile`/`_security` but not the `_source` named in its title, and why its commit message's "the extraction side was already correct" holds for three of the four parameters but not for `_source`.

## The defect

`_source` was never indexed on any backend, so `Patient?_source=…` matched **nothing** wherever the filter is genuinely applied — PostgreSQL, MongoDB, Elasticsearch, and SQLite since #474.

It is the only `Resource`-level parameter with no embedded definition, so it kept the spec bundle's `Resource.meta.source`. `filter_expression_for_resource` kept only the `|`-alternatives whose prefix matched the concrete resource type, and fell back to the *original* expression when none matched. FHIRPath resolves a leading identifier against the resource's own type, so `Patient.meta.source` and `meta.source` both resolve, but `Resource.meta.source` yields an empty collection — no index row was ever written. The four siblings survived only because embedded definitions shadow them with resource-relative expressions:

```
("_lastUpdated", "meta.lastUpdated",     Embedded)
("_id",          "id",                   Embedded)
("_profile",     "meta.profile",         Embedded)
("_security",    "meta.security",        Embedded)
("_tag",         "meta.tag",             Embedded)
("_source",      "Resource.meta.source", Embedded)   <-- from the spec file, unshadowed
```

## The fix

1. **`crates/persistence/src/search/extractor.rs`** — abstract base prefixes (`Resource.`, `DomainResource.`) are stripped rather than matched literally, so the remainder evaluates against the resource. This covers every present and future `Resource`-base parameter rather than special-casing `_source`; `_language` on R5/R6 is indexed by the same change. That's the reason for fixing it here rather than adding an embedded `_source` definition, which would paper over this one parameter.
2. **`crates/fhir/src/search/loader.rs`** — `_source` nonetheless joins the embedded fallback set, closing the asymmetry that made this a `_source`-only defect. It also stops a fallback-only registry (no spec bundle on disk) from typing `_source` as `String` by value-shape heuristic, which would then search the wrong index column.
3. **`crates/persistence/src/backends/sqlite/search/query_builder.rs`** — comment only. The `_ => None` arm that caused #474 claimed it "falls through to regular handling"; it does not, and the replacement says what `None` actually costs.

## Coverage

`tests/search/meta_params_suite.rs` is backend-agnostic and runs on **both** SQLite and PostgreSQL (the `#[path]` sharing pattern from `fts_purge_suite.rs`), asserting each of the four parameters **positively and negatively**, with a positive control that both fixtures are reachable before filtering.

Every one of those properties is load-bearing, and their absence is why this survived alongside #474:

- a positive-only test passes against a backend that ignores the filter;
- a negative-only test passes against one that never indexed it;
- and a single-backend test cannot distinguish a dropped filter from a correct filter over a missing index row — which is exactly the pair of bugs in play.

The existing SQLite-only `meta_params_tests.rs` from #474 is left untouched; the two are complementary and the new file's docs say so.

Confirmed to actually catch the defect: with the fix reverted, the suite fails on `_source` with 0 matches on PostgreSQL, while the other three still pass.

## Verification (rebased onto current `main`)

- `helios-persistence` lib tests (sqlite+postgres): 840 passed
- PostgreSQL binary against a real container: 155 passed
- SQLite `search_suite` (103) / `sqlite_tests` (85) / `crud_suite` (82) / `search_param_seeding` (6): all green
- `helios-fhir` search tests: 25 passed
- `cargo fmt --check` clean; clippy clean under the CI flag set with `-D warnings`

One pre-existing assertion needed updating: `search_param_seeding.rs` bounded the embedded fallback count at 9, which the new `_source` fallback exceeds. The loader's own fallback test now asserts the whole meta set by name and that none of them carries a `Resource.` prefix, so dropping one again fails loudly.

## Not included

`_source` still isn't advertised in the CapabilityStatement's common parameter list, and no backend declares `SpecialSearchParam::Source`. Nothing reads that flag outside tests today, so it's cosmetic — but now that the parameter genuinely works everywhere, advertising it would be truthful. Happy to add it here or as a follow-up.

https://claude.ai/code/session_01BT25XtZGQnsvdqVrfNZ8sk

### DIFF
--- a/crates/fhir/src/search/loader.rs
+++ b/crates/fhir/src/search/loader.rs
@@ -655,6 +655,25 @@ impl SearchParameterLoader {
             .with_source(SearchParameterSource::Embedded),
         );
 
+        // `_source` completes the meta-level set. Its absence here is what made
+        // #523 a `_source`-only defect: the other four were shadowed by these
+        // embedded definitions, whose expressions carry no `Resource.` prefix,
+        // so only `_source` kept the spec file's unevaluable
+        // `Resource.meta.source` and went unindexed. Without it, a registry
+        // built from the fallbacks alone also types `_source` by value-shape
+        // heuristic (String), which would look for the value in the wrong
+        // index column.
+        params.push(
+            SearchParameterDefinition::new(
+                "http://hl7.org/fhir/SearchParameter/Resource-source",
+                "_source",
+                SearchParamType::Uri,
+                "meta.source",
+            )
+            .with_base(vec!["Resource"])
+            .with_source(SearchParameterSource::Embedded),
+        );
+
         // SQL-on-FHIR canonical resources: `url` and `version` are not in the
         // base FHIR R4/R4B search-parameters bundle for ViewDefinition (which
         // first appears in R5+). The SoF `$viewdefinition-run`,
@@ -706,13 +725,33 @@ mod tests {
         let params = loader.load_embedded().unwrap();
 
         assert!(!params.is_empty());
-        assert!(params.len() <= 9, "Minimal fallback should have ~9 params");
+        assert!(
+            params.len() <= 10,
+            "Minimal fallback should have ~10 params"
+        );
 
-        let has_id = params.iter().any(|p| p.code == "_id");
-        assert!(has_id, "Should have _id parameter");
-
-        let has_last_updated = params.iter().any(|p| p.code == "_lastUpdated");
-        assert!(has_last_updated, "Should have _lastUpdated parameter");
+        // The whole `Resource`-level meta set, asserted by name. `_source` was
+        // the one omission, and because these definitions shadow the spec
+        // file's `Resource.`-prefixed expressions, its absence is what left
+        // `_source` unindexed (#523). Losing any of them again is the same bug.
+        for code in [
+            "_id",
+            "_lastUpdated",
+            "_tag",
+            "_profile",
+            "_security",
+            "_source",
+        ] {
+            let def = params
+                .iter()
+                .find(|p| p.code == code)
+                .unwrap_or_else(|| panic!("Should have {code} parameter"));
+            assert!(
+                !def.expression.starts_with("Resource."),
+                "{code} fallback expression must be resource-relative, got {:?}",
+                def.expression
+            );
+        }
 
         let has_patient_specific = params
             .iter()
--- a/crates/persistence/src/backends/sqlite/search/query_builder.rs
+++ b/crates/persistence/src/backends/sqlite/search/query_builder.rs
@@ -545,7 +545,13 @@ impl QueryBuilder {
                 self.build_filter_condition(&param.values, param_offset)
             }
             _ => {
-                // Other special parameters - fall through to regular handling
+                // Not "fall through to regular handling", as this comment used
+                // to claim: the caller returns whatever this yields, so `None`
+                // *drops* the parameter and the search answers with every
+                // resource of the type. That is how `_tag`/`_profile`/
+                // `_security`/`_source` came to be unfiltered (#474). Anything
+                // that must actually filter belongs in an arm above, or in the
+                // caller's exclusion list so it never reaches here.
                 None
             }
         }
--- a/crates/persistence/src/backends/sqlite/search_impl.rs
+++ b/crates/persistence/src/backends/sqlite/search_impl.rs
@@ -1301,8 +1301,8 @@ mod tests {
 
     fn create_test_backend() -> SqliteBackend {
         // Point at the workspace's data directory so the search-parameter
-        // registry loads the full FHIR spec (otherwise only the 5 minimal
-        // embedded params are available and chained-search tests fail to
+        // registry loads the full FHIR spec (otherwise only the handful of
+        // minimal embedded params are available and chained-search tests fail to
         // resolve param types like Observation.code → Token).
         let data_dir = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"))
             .join("..")
--- a/crates/persistence/src/search/extractor.rs
+++ b/crates/persistence/src/search/extractor.rs
@@ -328,18 +328,29 @@ impl SearchParameterExtractor {
     ///
     /// This method extracts only the parts that start with the given resource type and
     /// simplifies common patterns that use `resolve()`.
+    ///
+    /// Parts prefixed with an abstract base type (`Resource.`, `DomainResource.`)
+    /// apply to every resource, so the prefix is stripped instead of being matched
+    /// literally — see [`strip_abstract_base_prefix`].
     fn filter_expression_for_resource(&self, expression: &str, resource_type: &str) -> String {
         // Split by | and filter to parts starting with our resource type
         let parts: Vec<String> = expression
             .split('|')
             .map(|p| p.trim())
-            .filter(|p| {
+            .filter_map(|p| {
+                // Abstract-base parts first: when `resource_type` is itself
+                // "Resource", the literal prefix match below would keep the
+                // unevaluable form.
+                if let Some(rest) = strip_abstract_base_prefix(p) {
+                    return Some(rest);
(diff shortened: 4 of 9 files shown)


angela-helios (commented):
Reviewed — the fix itself is sound: I verified the two halves both carry weight (the extractor prefix-stripping covers spec-loaded `Resource.meta.source` expressions, and the embedded `Uri`-typed fallback is still needed because `infer_param_type_from_value` would type a URI as String in a fallback-only registry). Four observations, none blocking:

1. **`search_param_seeding.rs:189` — the `<=` bound passes silently on the exact regression class this PR fixes.** `outcome.created <= 10` duplicates the loader's magic count and had to be bumped in lockstep here; losing a fallback (the #523 failure mode) still passes. Deriving it — `assert_eq!(outcome.created, loader.load_embedded().unwrap().len())` — makes it exact and self-maintaining (`SearchParameterLoader` is already exported by helios-fhir).

2. **`extractor.rs:436` — the bare-abstract-base `$this` branch looks unreachable.** Bare concrete types (`Observation`) keep verbatim via the prefix match above; the branch only fires on a literal bare `Resource`/`DomainResource`, which no shipped parameter has (the test comment says as much). Returning `None` and dropping the evaluator round-trip would remove a branch future readers must puzzle over — or at least the comment's `Observation` example points at the wrong branch.

3. **`extractor.rs:430` — `strip_abstract_base_prefix` could stay zero-copy.** It returns `Option<String>` and the closure immediately re-borrows; both remainders are borrowable from the input and `"$this"` is 'static, so `Option<&str>` + `matches_type.then_some(p)` keeps the pre-existing `&str` pipeline. Runs per resource × parameter × union member.

4. **Nit:** the #523/#474 narrative is restated at ~6 sites (suite header, both suite-binding docs, postgres_tests module + fn docs, loader comment). One authoritative copy in the suite header with one-line pointers elsewhere would keep them from drifting.

Also ran the four meta-param suites locally against sqlite — green.

angela-helios (commented):
Follow-up with deeper findings — **two of these look blocking**, both from the same root: the abstract-prefix strip activates *every* `Resource.`-prefixed spec expression, not just the meta set this PR audited. On R5/R6 the spec also ships `_in` (`Resource.id`, type reference) and `_language` (`Resource.language`, token), both previously inert because their expressions evaluated empty:

1. **`_in` becomes a wrong-results parameter on R5/R6.** Every resource now indexes its *own id* as an `_in` reference row. `GET /Patient?_in=42` goes through the bare-id reference branch and returns `Patient/42` itself — asserting List membership that was never checked. Pre-PR it matched nothing. Also one junk index row per resource per reindex.

2. **`_language` diverges by backend on R5/R6.** Postgres filters it correctly via the token path; SQLite routes it to the special-handler `_ => None` arm (it's not in the exemption list at `query_builder.rs:391-395`) and silently drops the condition — every resource returned with a self link claiming the filter applied. That's the #474 failure mode, reintroduced for a parameter this PR makes live, and it violates the invariant the PR's own rewritten comment establishes ("anything that must actually filter belongs in an arm above, or in the caller's exclusion list").

   Suggested containment for both: strip only when the whole expression set is under audit — e.g. limit the strip to the known meta parameters, or explicitly handle/exclude `_in` and `_language` in the same PR.

3. **Union members containing a string-literal `|` now mangle.** The `'|'`-split runs before stripping; `Resource.extension.where(url='a|b').valueString` splits mid-literal, the stripped half becomes an unbalanced fragment, `evaluate_fhirpath` errors, and extraction of **all** members of that parameter aborts — a regression: pre-PR the concrete members survived because the mangled halves failed the type filter. (Pre-existing splitter weakness, but the strip turns it from harmless to destructive.)

4. **Conditional writes type `_source` wrong.** `CompositeStorage::infer_conditional_param_type` (composite/storage.rs:213) is a name heuristic that types `_source` as String (and never learned about it), while the extractor writes Uri rows — `If-None-Exist: Patient?_source=http://x` queries `value_string`, never matches, and **creates a duplicate on every request**. The registry-miss fallbacks in postgres/sqlite `build_search_parameters` have the same gap (`_tag|_profile|_security` → Token, no `_source` → Uri).

5. **No migration note**: pre-upgrade resources have no `_source` index rows, so `?_source=` silently misses older data until a `$reindex` — worth a line in the PR body/release notes.

6. Smaller, non-blocking: a parenthesized `(Resource.meta.source)` still silently extracts nothing (same #523 symptom, one level up); the `DomainResource` half of `ABSTRACT_BASE_TYPES` is unreachable for `base: ["DomainResource"]` definitions (extract() only consults the concrete-type and `Resource` buckets); and the loader test's `<= 10` bound can't catch a *lost* fallback (see my earlier comment) nor a feature-gated duplicate.

Happy to pair on the containment if useful — the meta-param fix itself is right, it's the blast radius of the generic strip that needs a fence.

PR #558, opened 2026-08-17
fix(ui): the review-pass fixes — 7 issues across chrome, dialogs, and the guided editor

### LINKED ISSUE #542: Investigate: Recent searches dropdown behavior when a user accumulates many queries
## Summary

Investigate how the **Recent searches** dropdown behaves once a user accumulates a lot of queries. A first pass through the code found that the recents list itself is bounded, but the control that displays it is not — the same dropdown also renders *every* saved query, in a panel with no scroll container.

## How it works today

**Storage.** `recentSearches` is an array in the per-user settings document (`/_user/settings`), newest-first, deduped by exact path string, capped at `MAX_RECENT = 10` — **client-side only** (`crates/ui/assets/saved-queries.js:41`).

**Recording.** Only explicit Runs record (`runSearch(path, record)`, `saved-queries.js:1603`); paging deliberately does not. `recordRecent` (`saved-queries.js:1750`) does: GET the whole settings doc → drop any exact-match entry → `unshift` → `PATCH {recentSearches: list.slice(0, 10)}` conditional on the ETag, re-reading and retrying on 412 (`MAX_RETRIES = 2`).

**Rendering.** `renderRecent` (`saved-queries.js:1669`) fills the one dropdown with **two** groups: a "Saved" group listing every entry of `savedQueries.<ResourceType>.<id>`, then the "Recent" group of up to 10.

**Server-side bounds** (`crates/rest/src/handlers/user_settings.rs`): whole document ≤ 256 KiB (`MAX_SETTINGS_DOCUMENT_BYTES:85`), saved queries ≤ 100 **per resource type** (`MAX_SAVED_QUERIES_PER_TYPE:89`).

## What looks wrong at scale

1. **The dropdown is unbounded even though recents are.** Recents cap at 10 — fine. But the Saved group above them has no cap, no scroll, and no filter box. With a 100-per-type server limit across ~145 resource types, only the 256 KiB document cap bounds the row count. Hundreds of rows in one `<details>` panel is reachable.
2. **No scroll container.** `.menu__panel--right` (`crates/ui/assets/app.css:1523`) sets `width: 380px; max-width: 90vw` but **no `max-height` and no `overflow-y`** — unlike `.addbox--modal` (`app.css:955`) and `.nav-panel` (`app.css:3559`), which both do. Needs confirming in a browser whether a long list is reachable by page scroll or clipped outright.
3. **Recents get buried.** The Saved group renders first and never collapses, so the "Recent" group the control is named after drops below the fold as soon as there are more than a handful of saved queries.
4. **Arbitrary saved ordering.** Saved rows sort by resource type, then by raw `id` (`Object.keys(entries)`), not by name, recency, or `accessCount` — which reads as random once the list is long. (The recents themselves are correctly newest-first.)
5. **Exact-string dedupe.** `/Patient?name=smith`, `/Patient?name=Smith`, and the same params in a different order are three separate entries, so the 10 slots fill with near-duplicates of one logical search.
6. **Write cost per Run.** Every recorded run is a full document GET + a PATCH + a full re-render of the saved list. As the document grows toward 256 KiB, each search pays that round trip in both directions, plus a re-read on every 412 from a concurrent tab.
7. **No server-side cap on `recentSearches`.** `validate` bounds document size and saved-queries-per-type, but nothing bounds the recents array (`user_settings.rs:266-330`). A stale client, a different client, or a direct `PATCH /_user/settings` can grow it until the 256 KiB cap — at which point **every** settings write starts failing, including theme, nav, and tenant choice. A delete patch should shrink the document back under the cap, but that recovery path is untested.

## Questions to answer

- [ ] At what row count does the dropdown become unusable, and is the overflow scrollable or clipped? (Seed ~100 and ~500 saved queries and look.)
- [ ] Should the Saved group be capped/filtered/scrolled in this dropdown, or moved out of it entirely? The Saved Queries page already lists them properly.
- [ ] Is `MAX_RECENT = 10` the right number, and should it be enforced server-side as well as client-side?
- [ ] Should dedupe normalize the query (param order, case) instead of comparing raw strings?
- [ ] What happens to the UI when the 256 KiB cap is hit — is the error surfaced usefully, and can the user recover without hand-editing the document?
- [ ] Is the per-Run full-document read/write acceptable as the document grows, or should recents move to a narrower endpoint?

## Acceptance criteria

- [ ] Findings written up on this issue with reproduction steps and row counts.
- [ ] The dropdown stays usable and scrollable at the maximum row count the server permits.
- [ ] Recents remain reachable regardless of how many saved queries exist.
- [ ] A server-side bound on `recentSearches` exists, or a documented decision that the client cap is sufficient.
- [ ] Tests cover the cap, eviction order, and a large saved list — today there is one e2e assertion that a run appears (`crates/ui/e2e/tests/queries.spec.ts:393`) and nothing else.

## Pointers

- `crates/ui/assets/saved-queries.js` — `MAX_RECENT:41`, `recentSearches():121`, `renderRecent():1669`, `recordRecent():1750`, click/delete handlers `:1962`
- `crates/ui/templates/partials/search-builder.html:28-41` — the `<details>` dropdown
- `crates/ui/assets/app.css:1523` (`.menu__panel--right`), `:1530-1590` (`.recent-item*`), `:3647` (`.recent-group`)
- `crates/rest/src/handlers/user_settings.rs:78-90, 266-330` — size and per-type caps
- `crates/ui/e2e/tests/queries.spec.ts:393` — the only current coverage

### LINKED ISSUE #544: UI: hide the tenant picker unless a non-default tenant has been provisioned
## Summary

The tenant picker in the sidebar renders on every page for every install, including single-tenant ones where it offers exactly one choice and does nothing. Hide it unless the user has actually provisioned a tenant beyond the server default.

## Current behavior

`crates/ui/templates/layouts/base.html:32-50` renders the `<details class="menu">` selector unconditionally, on every page. The summary shows `status.tenant_initials()` + `status.tenant_label()`; the panel's options are fetched lazily by htmx from `GET /ui/tenant/options` when the menu first opens.

`tenant_options` (`crates/ui/src/lib.rs:830-862`) lists the registry and, when the effective tenant isn't among the rows, inserts it at the top. So on a stock single-tenant server the menu opens to exactly one entry — the default tenant, already checked — plus the "Tenants" maintenance link. It is a control with nothing to control.

## Wanted

Render the picker only when there is something to pick: more than one provisioned tenant, or a provisioned tenant that isn't the server default (`state.default_tenant`). Otherwise the sidebar shows either a plain non-interactive tenant label or nothing at all — **decide which**, since the current label doubles as the "which tenant am I in" indicator and that information may still be worth showing on a single-tenant install.

## The catch worth knowing before starting

The selector's option list is deliberately **not** loaded at page render time. From `crates/ui/src/lib.rs:808-809`: *"loaded when the menu opens (htmx) so pages do not pay a registry listing per load."* Deciding whether to render the control requires knowing the tenant count *before* the menu opens — i.e. exactly the per-load registry listing that design avoided.

So this is not a template-only change. Suggested approach:

- Cache the count (or the "has a non-default tenant" boolean) in `WebState`, refreshed when a tenant is provisioned or deregistered through `/ui/tenants` (`tenants::create` / `tenants::delete`, `crates/ui/src/lib.rs:704-706`). A short TTL is a reasonable fallback for out-of-band registry changes.
- Expose it on `Status` (`crates/ui/src/lib.rs:262+`), which `base.html` already has, and gate the `<details>` on it.
- **Don't** solve it by rendering the control and hiding it client-side after the options load — it sits in the sidebar above the fold and would flash on every page, and it would break with JS off.

Also note `state.tenants` is `Option<Arc<dyn ResourceStorage>>` and `list_tenants()`'s default impl returns an empty vec (`crates/persistence/src/core/storage.rs:786`), so backends without a registry report zero tenants. Make sure "no registry" collapses to hidden, not to some error state.

## Non-issue, so nobody has to re-check it

Hiding the picker does **not** strand the user: `/ui/tenants` is already a first-class sidebar nav entry (`base.html:96`), independent of the selector panel. Creating a second tenant stays reachable, and the picker appears once one exists.

## Acceptance criteria

- [ ] On a server with only the default tenant provisioned, the sidebar tenant picker is not rendered.
- [ ] As soon as a second (or non-default) tenant is provisioned, the picker appears without a restart.
- [ ] Deregistering back down to one tenant hides it again.
- [ ] Page loads do not gain a registry listing per request.
- [ ] The behavior holds with JavaScript disabled (no flash, no client-side-only hiding) — cover it in `e2e/tests/nojs/`.
- [ ] A decision is recorded on whether the current tenant name still shows as static text when the picker is hidden.
- [ ] Tests updated/added: `crates/ui/tests/tenants_http.rs:300` (`tenant_choice_persists_and_the_selector_follows_it`) asserts the selector's presence today, and `crates/ui/e2e/tests/tenants.spec.ts` covers the maintenance page — neither covers the single-tenant hidden case.

## Pointers

- `crates/ui/templates/layouts/base.html:32-50` — the selector; `:96` — the independent Tenants nav entry
- `crates/ui/src/lib.rs:830-862` — `tenant_options`; `:808-809` — the lazy-load rationale
- `crates/ui/src/lib.rs:214-232` — `resolve_prefs`, where the effective tenant and `state.default_tenant` are resolved
- `crates/ui/templates/partials/tenant-options.html`
- `crates/ui/src/tenants.rs` — provision/deregister, the cache-invalidation points
- `crates/persistence/src/core/storage.rs:786` — `list_tenants` default impl

### LINKED ISSUE #545: UI: Add-tenant panel can't be cancelled — no close (×) control, no Esc, no click-outside
## Summary

On the Tenants page, opening **Add tenant** and typing something leaves you with no way to back out. The panel has a submit button and nothing else — no ×, no Cancel, no Esc, and clicking elsewhere on the page does nothing. Add a close control.

## Repro

1. Go to `/ui/tenants`.
2. Click **Add tenant**.
3. Type something into Tenant ID.
4. Try to cancel. The only thing that works is clicking the **Add tenant** button again — a control labeled as the *open* action, so nobody looks there.

## What's actually there

`crates/ui/templates/pages/tenants.html:51-78` — the panel is a native `<details class="addbox">` disclosure (chosen so it works without JavaScript). Its `.addbox__actions` block holds a single submit button. `.addbox` in `crates/ui/assets/app.css:860-930` is `position: absolute` with no backdrop, so clicks outside the panel land on the page and do nothing.

`<details>` has no Esc behavior of its own, and no script adds one here.

**The same panel pattern is used in two other places, with a different failure:**
`bulk-import.html:12` and `bulk-import-detail.html:14` use `<details class="addbox addbox--modal">`. The `--modal` variant (`app.css:931-960`) stretches the `<summary>` across the viewport as a backdrop while open, so click-outside *does* close those. But there is still no × and no Esc, and an invisible backdrop is not a discoverable affordance. Whatever we do here should cover all three.

## The pattern to copy

The app already solves this correctly one page over. The Resources edit modal has an explicit × — `crates/ui/templates/pages/resources.html:133`:

```html
<button type="button" class="modal__x" data-modal-close aria-label="{{ i18n.t("action-cancel") }}">×</button>
```

— with Esc handled in `crates/ui/assets/resources.js:53-54`. The `action-cancel` Fluent key already exists (`locales/en/main.ftl:66`), so no new string is needed for the label.

## Constraint worth knowing before starting

The whole reason these are `<details>` and not `<dialog>` is that they open without JavaScript (`tenants.html:45-48` says so explicitly). A `<details>` cannot be closed from *inside* its own panel without script, so:

- **Suggested:** add the × as a progressive enhancement — present in the markup, hidden by default, revealed by a small pinned asset that also wires Esc and sets `details.open = false`. With JS off, behavior is exactly what it is today, so nothing regresses.
- Also give the plain `.addbox` a click-outside backdrop like `--modal` already has, so dismissal doesn't depend on JS at all.
- **Alternative:** move to a native `<dialog>`, which gets Esc and `<form method="dialog">` close buttons for free — but `showModal()` is JS-only, so no-JS opening would regress. Flagging the trade-off rather than picking it.

## Also decide

Should the form reset when the panel closes? Today the typed values survive a close/reopen. Either behavior is defensible; pick one deliberately — "cancel" implying the input is discarded is the more common expectation.

## Acceptance criteria

- [ ] The Add-tenant panel has a visible close control that discards the panel.
- [ ] Esc closes it.
- [ ] Clicking outside closes it, on the plain `.addbox` as well as `--modal`.
- [ ] Same treatment applied to the two bulk-import dialogs.
- [ ] Nothing regresses with JavaScript disabled — cover it in `crates/ui/e2e/tests/nojs/`.
- [ ] Close control is labeled for screen readers (reuse `action-cancel`) and reachable by keyboard; `a11y.spec.ts` passes with the panel open.
- [ ] A recorded decision on whether closing resets the form.
- [ ] e2e coverage for open → type → cancel on `/ui/tenants` (`crates/ui/e2e/tests/tenants.spec.ts` has no test for this today).

## Pointers

- `crates/ui/templates/pages/tenants.html:51-78` — the reported panel
- `crates/ui/templates/pages/bulk-import.html:12`, `bulk-import-detail.html:14` — same pattern, `--modal` variant
- `crates/ui/assets/app.css:860-930` (`.addbox`), `:931-960` (`.addbox--modal` backdrop trick), `:927` (`.addbox__actions`)
- `crates/ui/templates/pages/resources.html:120-135` + `crates/ui/assets/resources.js:53-54` — the × and Esc pattern to reuse
- `locales/en/main.ftl:66` — `action-cancel`

### LINKED ISSUE #546: UI editor: creating a non-Patient resource dumps unanchored, duplicated "X: X is required" errors under the Guided form
## Summary

Creating a new resource of any type that has required elements (i.e. anything other than Patient) drops a wall of unstyled red text under the Guided form. See the screenshot: a brand-new `Claim` shows "8 issues" and eight free-floating lines like `priority: priority is required`, with no fields to fix them, the element name printed twice, and the list running off the bottom of the modal.

## Repro

1. `/ui/resources` → pick any type with required elements (Claim, Encounter, Observation…).
2. **Create new**.
3. The Guided form renders `Claim` + a lone "+ Add element" button, then eight red lines below it.

Patient never shows this because every Patient element is optional — which is why it only reproduces "on anything other than Patient".

## Root cause

`crates/ui/src/editor.rs:346-358`. `build_rows` only emits rows for elements **present in the document**. A new resource is `{"resourceType": "Claim"}`, so there are no rows. Meanwhile `validate_sync` reports a required-cardinality error for every mandatory element. Any error whose path doesn't match a row falls through to `orphan_errors`:

```rust
.map(move |message| format!("{path}: {message}"))
```

and the validator's own message is already `"{key} is required"` (`crates/fhir-validator/src/engine/errors.rs:138-140`) — so the concatenation yields **`priority: priority is required`**. Every required element on a blank resource becomes an orphan.

Rendering is `crates/ui/templates/partials/editor-body.html:214-215` — a bare `<p class="editor__orphan">` per message, styled by four lines of CSS (`crates/ui/assets/app.css:3378-3382`: margin, font-size, color) with no container, heading, icon, or grouping. That is the visual mess.

## Defects, separately

1. **Duplicated element name** — `priority: priority is required`, from the path prefix plus the validator message that already names the key.
2. **Non-deterministic order.** `orphan_errors` is built by iterating a `HashMap` (`editor.rs:349`), so the list reshuffles on every keystroke. The screenshot's order (priority, status, patient, use, created, provider, insurance, type) is not ElementDefinition order.
3. **No affordance.** The form says `priority` is required but renders no row for it — the only control is a generic "+ Add element" the user must search through.
4. **Unstyled and uncontained.** Orphans are loose paragraphs after the tree, not a panel, list, or summary.
5. **Overflow.** The last entry is clipped at the modal's bottom edge in the screenshot; confirm whether that region scrolls.
6. **Untranslated.** These strings come straight from the validator, not the Fluent catalog, against the crate's "no user-visible prose outside `locales/`" rule (`crates/ui/README.md`).
7. **Arguably premature.** The errors are *correct* — a Claim really does require those elements — but firing all of them at an untouched blank resource, before the user has typed anything, is noise. Worth deciding whether the live pass should suppress required-cardinality errors until a field is touched or a save is attempted.

## Suggested direction (not prescriptive)

The cleanest fix probably removes the orphan case for this scenario entirely: **seed a row for every required element** when building the form, so the errors anchor to real, fillable rows exactly like every other validation message. Orphans then go back to being what they were meant for — genuinely unmatchable paths — and should still be given a proper container and a stable sort. Whatever the approach, `format!("{path}: {message}")` needs to stop double-printing the key.

## Acceptance criteria

- [ ] A new resource of a type with required elements renders a usable form, not a block of loose red text.
- [ ] Required elements are reachable/fillable without hunting through "+ Add element".
- [ ] No message prints the element name twice.
- [ ] Error order is stable across re-renders.
- [ ] Any remaining orphan errors have a styled container and don't overflow the modal.
- [ ] A recorded decision on whether required-cardinality errors fire before the user has entered anything.
- [ ] Messages shown to users come from the Fluent catalog.
- [ ] Test coverage for creating a non-Patient resource — `crates/ui/e2e/tests/resources-editor.spec.ts` and `editor-controls.spec.ts` don't cover this today.

## Pointers

- `crates/ui/src/editor.rs:300-370` — `validate_sync`, path anchoring, `orphan_errors`
- `crates/ui/templates/partials/editor-body.html:214-215` — orphan rendering; `:86-147` — the row + `editor-row__error` pattern the orphans should join
- `crates/ui/assets/app.css:3378-3382` — `.editor__orphan`
- `crates/fhir-validator/src/engine/errors.rs:138-140` — `msg_required`

---

## Also on this panel: the "Checked as you type / Checked on save" legend is crammed

The legend directly under the Guided form heading squeezes two full sentences onto a single line and is barely readable.

`.editor-legend` (`crates/ui/assets/app.css:4143-4149`) is:

```css
.editor-legend {
  display: flex;
  gap: 14px;
  margin: 4px 0 8px;
  color: var(--muted);
  font-size: 11px;
}
```

- **No `flex-wrap`**, so the two items are forced onto one row and compress against each other however narrow the container gets. This is worst inside the Resources edit modal, which is narrower than the standalone `/ui/editor` page.
- **11px `--muted` text** for two complete sentences — under the minimum that reads comfortably, and the low-contrast muted token compounds it.
- The strings are long and get longer when translated: EN `editor-legend-live` is 63 characters; the German equivalent is 67 (`locales/de/main.ftl:442`). Any fix has to hold up in `es` and `de`, not just `en`.

Worth reconsidering the shape rather than only the wrapping — e.g. stack the two items, or reduce each to a short dot-plus-label with the detail moved to a tooltip or the panel's help text. The two `::before` dots (`app.css:4151-4167`) are decorative; the meaning is carried by the text, so restructuring is safe for screen readers.

**Additional acceptance criteria**

- [ ] The legend is legible at the modal's width, not just on the full editor page.
- [ ] It wraps or restacks rather than compressing, and holds up in `en`, `es`, and `de`.
- [ ] Type size and contrast are appropriate for body prose (`a11y.spec.ts` should keep passing).

**Pointer:** `crates/ui/assets/app.css:4143-4167`, `crates/ui/templates/partials/editor-body.html:79-82`, `locales/{en,es,de}/main.ftl:442-447`

### LINKED ISSUE #547: UI editor: Guided form is not usable for authoring a new resource (empty tree, picker and focus lost on every edit)
## Summary

Creating a new Patient from `/ui/resources` gives you a Guided form you can't practically author in: it opens with no fields at all, and every interaction with it destroys and re-renders the whole panel, losing the add-element picker, focus, and caret each time. Investigate and fix the guided authoring path for new resources.

## Repro

1. `/ui/resources` → **Create new** with Patient selected.
2. The Guided form shows `Patient` and a single "+ Add element" button. No fields.
3. Open "+ Add element", pick an element — the picker slams shut and the panel re-renders.
4. Type a value, then move to the next field — the field you're leaving triggers a re-render that replaces the element you were moving into.

Net effect: the guided form reads as broken. Every existing test that populates a new resource goes through the raw JSON textarea instead (see coverage note below), so nothing catches this.

## What the code shows

**1. A new resource has no rows to show.** `build_rows` (`crates/ui/src/editor.rs:333-344`) walks the elements *present in the document*. A new Patient is `{"resourceType": "Patient"}`, so there is nothing to walk and the tree renders empty. Patient has no required elements either, so unlike the Claim case in #546 there aren't even errors — just a blank panel with one button. Whether the form should pre-seed common or required elements is a design decision worth making here.

**2. Every mutation replaces the entire panel.** Both implementations do a full `innerHTML` swap with no state restoration:

- `crates/ui/assets/resources.js:79-90` — `editorBody.innerHTML = html`
- `crates/ui/assets/editor.js:52-70` — `body.innerHTML = html` (`applyView()` afterwards is a documented no-op, `editor.js:193-195`)

Nothing preserves focus, caret position, scroll offset, or `<details>` open state across the swap. Two consequences follow directly:

- **The add picker closes on every add.** The element list is a `<details class="editor-add">` (`crates/ui/templates/partials/editor-body.html:151-157`). Adding an element re-renders the panel, so the disclosure collapses and the filter box clears. Adding five elements means reopening and re-filtering five times.
- **Field-to-field editing fights itself.** Primitive edits commit on `blur` (`resources.js:176-180`, `editor.js:307-315`) — so leaving a field fires a round trip that rebuilds the DOM, discarding the input you were tabbing into along with the focus.

**3. The same interactions are implemented twice.** `resources.js:75-78` says it plainly: *"The editor inside the modal is the same fragment the Editor page uses, but editor.js is bound to that page's ids — so the interactions live here, scoped to the modal body."* Any fix has to land in both files, and the two can drift independently. Worth considering whether they should be unified as part of this.

## Please confirm in a browser first

This is a code reading of a report that the form "appears not to be working" — pin down which step actually fails before fixing. In particular, confirm whether focus loss makes Tab-through editing impossible or merely awkward, and whether `/ui/editor` (the standalone page) behaves the same as the modal.

## Suggested direction

The root problem is the wholesale swap. Options, roughly in order of appetite:

- Preserve state across the swap: restore focus by `data-path`, restore caret, keep `<details open>` and scroll. Smallest change, keeps the round-trip model.
- Swap only the subtree that changed rather than the whole body — htmx already does targeted swaps everywhere else in this app.
- Reconsider committing primitive edits on `blur` at all, given a commit re-renders the thing you just left.

Plus a decision on what a brand-new resource should show: an empty tree, the required elements pre-seeded, or a small set of common ones.

## Acceptance criteria

- [ ] A new Patient can be authored end to end in the Guided form, without touching raw JSON.
- [ ] Adding several elements in a row does not require reopening the picker each time.
- [ ] Focus and caret survive a value commit; Tab moves to the next field and stays there.
- [ ] Scroll position and open/closed panel state survive a re-render.
- [ ] The same behavior holds on `/ui/editor` and in the Resources modal.
- [ ] A recorded decision on what the guided form shows for a brand-new resource.
- [ ] e2e coverage through the guided path — today **every** create test in `crates/ui/e2e/tests/resources-editor.spec.ts` populates the resource via `fillRaw` / `applyJson` (the raw JSON textarea, `e2e/pages/editor.ts:50-58`), so the guided authoring path is entirely untested.

## Pointers

- `crates/ui/src/editor.rs:333-344` — `build_rows`, which only walks elements present in the document
- `crates/ui/assets/resources.js:79-90` — modal round trip; `:176-180` — blur→set; `:93-121` — add/remove
- `crates/ui/assets/editor.js:52-70` — standalone round trip; `:307-315` — blur→set; `:365-373` — add filter
- `crates/ui/templates/partials/editor-body.html:151-205` — the `<details>` add picker
- `crates/ui/e2e/tests/resources-editor.spec.ts`, `crates/ui/e2e/pages/editor.ts`
- Related: #546 (the same panel's error rendering for non-Patient types)

### LINKED ISSUE #548: UI editor: move Collapse all / Expand all next to the JSON heading, leave Edit raw alone on the right
## Summary

In the resource editor's JSON pane, all three header buttons are currently bunched together on the right. Split them: **Collapse all** and **Expand all** belong next to the "JSON" heading on the left, and **Edit raw** should sit alone on the right.

## Current markup

`crates/ui/templates/partials/editor-body.html:31-38` puts all three in one `.card-head__tools` container:

```html
<div class="card-head">
  <h3><span class="icon">…</span> {{ i18n.t("editor-view-json") }}</h3>
  <div class="card-head__tools">
    <button … data-json-fold="all">{{ i18n.t("editor-collapse-all") }}</button>
    <button … data-json-fold="none">{{ i18n.t("editor-expand-all") }}</button>
    <button … id="editor-json-edit">{{ i18n.t("editor-edit-raw") }}</button>
  </div>
</div>
```

`.card-head` is `display: flex; justify-content: space-between` (`crates/ui/assets/app.css:2948-2955`), so the heading pins left and the single tools group pins right.

## Wanted

`[ <> JSON ] [Collapse all] [Expand all] ……………………… [Edit raw]`

The two fold controls read as operations *on the JSON view* and belong with it; "Edit raw" is a mode switch and deserves the visual separation.

## Notes for whoever picks this up

- **No JS changes needed.** Both click handlers are delegated from the editor body container — `crates/ui/assets/resources.js:93-97` (modal) and `crates/ui/assets/editor.js:233-262` (standalone page) — and match on `[data-json-fold]` / `#editor-json-edit`, not on DOM position. Moving the buttons inside the same container is safe.
- **`.card-head` is shared.** It's also used by the Guided form header in the same fragment (`editor-body.html:66`), by `pages/batch.html:33,77`, and by `pages/editor.html:56`. Prefer a modifier or an extra wrapper over changing `.card-head`'s own flex rules, so the other four headers don't shift.
- The fragment renders in two places — the standalone `/ui/editor` page and the Resources edit modal — so check both; the modal pane is narrower and is where crowding shows first.
- Related: #543 tracks unifying the stylesheet approach. If a shared header primitive comes out of that work, this layout should be expressed through it rather than as a one-off.

## Acceptance criteria

- [ ] Collapse all / Expand all sit immediately after the "JSON" heading on the left.
- [ ] Edit raw sits alone at the right edge of the pane header.
- [ ] Collapse all, Expand all, and Edit raw all still work, on `/ui/editor` and in the Resources modal.
- [ ] The Guided form header, `batch.html`, and `editor.html` headers are visually unchanged.
- [ ] Layout holds at the modal's narrower width without the buttons colliding or wrapping awkwardly.

## Pointers

- `crates/ui/templates/partials/editor-body.html:31-38`
- `crates/ui/assets/app.css:2948-2971` (`.card-head`, `.card-head__tools`), `:2972-2984` (`.editor-json__act`)
- `crates/ui/assets/resources.js:93-97`, `crates/ui/assets/editor.js:233-262`

### LINKED ISSUE #549: UI editor: assist entry of FHIR temporal primitives (date, dateTime, time, instant) in the Guided form
## Summary

The Guided form gives no help entering FHIR temporal primitives — `date`, `dateTime`, `time`, `instant`. Every one is a bare text box, so the user has to know the spec's format rules from memory and only finds out they got it wrong after a server round trip. Add authoring assistance for these types.

Spec reference: <https://hl7.org/fhir/datatypes.html#primitive>

## Current behavior

`crates/ui/templates/partials/editor-body.html:133-138` renders **every** primitive identically:

```html
<input class="editor-row__input" type="text" value="{{ row.value }}" data-set="{{ row.path }}" …>
```

No format hint, no placeholder, no picker, no client-side check. A malformed `birthDate` is caught only on blur, when the value round-trips to `/ui/editor/render` and comes back as a red row error.

## Two things already in place that this can build on

**1. The row already knows its FHIR type.** `Row.type_label` is populated from the schema's `type_` (`crates/ui/src/editor.rs:446-449`), so at render time the server knows a field is `dateTime` rather than just "a string". The template currently only prints it as a hint chip (`editor-body.html:98-99`). No new plumbing is needed to branch on type.

**2. The authoritative format regex is already loaded.** The validator carries the FHIR spec pattern per primitive on `schema.regex` and matches against it (`crates/fhir-validator/src/engine/primitives.rs:41-49`), compiled anchored as `^(?:…)$`. That same pattern can drive a live client-side check and a generated format hint, instead of hand-copying regexes into the UI.

## The design constraint — don't just drop in native pickers

Native HTML date inputs cannot represent what FHIR permits, and would silently destroy valid data:

| Type | FHIR permits | Why the native control doesn't fit |
|---|---|---|
| `date` | `YYYY`, `YYYY-MM`, or `YYYY-MM-DD` — partial precision is legal | `<input type="date">` can only express a full day |
| `dateTime` | the three `date` forms, **or** a full timestamp; if a time is present a timezone offset is **required** | `<input type="datetime-local">` has no timezone, and can't do partials |
| `instant` | always full precision with an offset (`YYYY-MM-DDThh:mm:ss.sss+zz:zz`) | no native control produces this |
| `time` | `hh:mm:ss[.sss]`, no date, no timezone | `<input type="time">` commonly omits the seconds FHIR requires |

So the input must stay free-text and authoritative, with assistance layered on top. There's a precedent in this same editor: the live `$expand` picker enriches a plain text input with a datalist via `data-vs-url` (`editor-body.html:136`, `resources.js:126-172`) without taking over the field. The same shape fits here.

## Suggested direction

- **Per-type placeholder and format hint** driven off `type_label`, stating what's allowed including partial precision.
- **Live format feedback** using the schema regex already available server-side — flag a malformed value as you type rather than after a blur round trip.
- **An optional assist affordance** that writes a conformant string: a picker, plus shortcuts like "today" / "now". Precision must be an explicit choice for `date`/`dateTime`, never silently upgraded from `2024` to `2024-01-01`.
- **Timezone handling** for `dateTime` and `instant` — default to the browser offset, allow `Z` or an explicit offset, and never emit a time without one.
- Keep the typed value the source of truth so an existing partial date survives a visit to the field untouched.

## Scope

In scope: `date`, `dateTime`, `time`, `instant`.

Out of scope but worth noting: the complex types built from these — `Period` (`start`/`end` are `dateTime`), `Timing` (`event`, `repeat.bounds[x]`), `Age`/`Duration` — pick up the benefit for free once the primitive rows improve, since their children render as ordinary rows. Also, the same "no assistance" gap applies to `boolean`, `integer`, `unsignedInt`, and `positiveInt`, which are likewise plain text boxes; if the mechanism generalizes cheaply, say so, but the ask here is the temporal types.

## Acceptance criteria

- [ ] `date`, `dateTime`, `time`, and `instant` rows show what format is expected before the user types.
- [ ] Malformed values are flagged without waiting for a server round trip.
- [ ] Partial-precision `date` and `dateTime` values can be entered and are preserved unchanged.
- [ ] A `dateTime`/`instant` with a time always carries a timezone offset.
- [ ] `time` values keep the required seconds component.
- [ ] Existing values load, display, and survive an untouched visit to the field.
- [ ] Works on `/ui/editor` and in the Resources edit modal (two separate JS implementations — see #547).
- [ ] Keyboard-accessible and labeled; `a11y.spec.ts` still passes.
- [ ] Any user-facing text comes from the Fluent catalog (`locales/{en,es,de}/main.ftl`).
- [ ] e2e coverage for entering each temporal type through the guided form.

## Pointers

- `crates/ui/templates/partials/editor-body.html:133-138` — the single untyped input
- `crates/ui/src/editor.rs:442-455` — `Row` construction, `type_label` from `schema.type_`
- `crates/fhir-validator/src/engine/primitives.rs:1-50` — the spec regex per primitive
- `crates/ui/assets/resources.js:126-172`, `crates/ui/assets/editor.js:318-360` — the `$expand` datalist pattern to model this on
- Related: #547 (guided-form authoring is unusable for new resources), #546 (error rendering on the same panel)

angela-helios:
Closes #542. Closes #544. Closes #545. Closes #546. Closes #547. Closes #548. Closes #549.

Three of the quick items from the UI review pass, batched because they share the chrome:

**#544 — tenant picker gating.** The sidebar picker rendered on every install, offering exactly one choice on single-tenant servers. It now renders only when the effective tenant differs from the default or the registry knows a second tenant. The check rides the existing `resolve_prefs` middleware (one indexed `list_tenants` read per page load — the same cost class as the settings read it already performs), flows through `RequestTenant.multi` → `Status::show_tenant_picker()`, and the layout gates on it. Tested both directions: hidden on a fresh single-tenant install, appears the moment a second tenant is provisioned through the page's own form.

**#545 — Add-tenant panel escape hatches.** The panel had a submit button and no way out. A small `addbox.js` asset (defer-loaded, ~30 lines, no framework) adds three: **Esc**, **click outside**, and explicit **× / Cancel** controls wired via `data-addbox-close`. The same affordances land on the bulk-import dialogs, which had the same gap. The `<details>` disclosures remain fully usable with JavaScript disabled — the script only adds ways out, per the crate's no-JS-first rule.

**#548 — JSON pane header layout.** Collapse all / Expand all now sit next to the "JSON" heading (a `card-head__tools--start` group with `margin-right: auto`); Edit raw stays alone on the right.

Full helios-ui suite green (13 tenants tests incl. the two new behavior tests, i18n parity ×3 with the new `ui-cancel`/`ui-close` keys), clippy clean, `--features ui` builds.

**Added after opening:** the #542 investigation landed here too — the Recent/Saved dropdown now scrolls at min(420px, 60vh) and the recents cap is re-asserted on read; full findings on the issue.

**Second batch (same review pass):**

**#546 — orphan issues panel.** Creating a resource with required elements dumped bare, duplicated lines ("priority: priority is required") under the Guided form. Unanchored issues now dedupe, drop the doubled element name (the validator's message already opens with it), and render in a bounded, titled alert panel. The deeper "give each missing required element an actual field" belongs with the #547 authoring-UX work.

**#549 — temporal input assistance.** `date`/`dateTime`/`time`/`instant` inputs carry a format placeholder, a spec-shaped `pattern` driving live native `:invalid` styling (red before any server round trip), and a hint title. Patterns rather than native date pickers on purpose: FHIR's partial dates (`2024`, `2024-05`) are legal and native pickers can't express them.

**Third batch — #547, guided authoring.** The editor's full-body re-render is by design (the document is the state, validation server-side); what it destroyed was the user's place. Now: a blur with an unchanged value doesn't round-trip at all (tabbing stops re-rendering); the swap captures UI state at response time and restores focus + caret, open add-pickers with their filter text, and tree scroll; a mutation that creates a node reports its path (`data-focus` on the fragment) and the caret lands in the new field selected and ready to type, with the picker left open so multi-add flows; and an empty document opens the root picker by itself. Applied to both editor.js and the resources modal's copy of the cycle. This also delivers the deeper half of #546: required-element issues now have an immediate path to a field (auto-open picker → click → focused input).


angela-helios:
Brought this branch up to date with main (which now carries #539) and cherry-picked the #560 fix onto it, so the merge ref is green regardless of merge order: the rsa crate is out of the lock here too, and the keys test asserts the redirect contract. If #560 merges first, the shared changes collapse to a no-op on this side.

smunini (approved):

PR #574, opened 2026-08-19
feat(ui,rest,observability): make the dashboard chart functional — real type picker, multi-series, tooltip, a11y

### LINKED ISSUE #555: Home dashboard: make the "FHIR resources over time" chart functional
## Summary

The "FHIR resources over time" chart on the Home page renders, but most of what surrounds it is inert: two controls that look interactive do nothing, there is no hover or tooltip, only one series is ever plotted, the type list is a fixed hardcoded set, and the whole SVG is hidden from assistive technology. Make it functional.

**Design reference:** [Helios FHIR Server Hi-Fi mockup](https://claude.ai/design/p/8f15d8ea-6935-489c-adf8-82eaeec2358e?file=Helios+FHIR+Server+Hi-Fi.html&via=share) — use this as the guide for the target behavior and appearance. *(Note: this link is not machine-readable from CI/tooling — it returns 403 to unauthenticated fetches — so the findings below come from reading the code. Reconcile them against the mockup, which wins on anything visual.)*

## What works today

The chart is genuinely server-rendered from live data: `build_chart` (`crates/ui/src/lib.rs:1395+`) computes SVG geometry from per-type cumulative counts supplied by `helios_observability::dashboard`, and both the time-window selector and the legend are plain links that re-render the page — so they work without JavaScript. That progressive-enhancement baseline should survive whatever is added.

## What is not functional

**1. The resource-type pill is not a control.** `crates/ui/templates/pages/index.html:52-56` renders it as a `<span class="pill">` with a chevron-down icon — it looks exactly like a dropdown and does nothing. The actual type selector is the legend at the bottom of the card. Either make the pill a real picker or stop styling it as one.

**2. The expand button is dead.** `index.html:57-59` is a `<button class="pill pill--square" aria-label="{{ i18n.t("chart-expand") }}">` with no handler — there is **no chart JavaScript anywhere** in `crates/ui/assets/`. Nothing binds it.

**3. No hover, tooltip, or data points.** The series is a bare `<polyline>` (`index.html:78`) with no point markers and no readout. You cannot tell what value any position represents.

**4. Only one series at a time.** `ChartView` carries a single `polyline` for `selected_type` (`crates/ui/src/lib.rs:352-368`). Comparing types means re-rendering the page one type at a time. Whether the target is multi-series overlay is a mockup question.

**5. The charted types are a hardcoded list.** `DEFAULT_DASHBOARD_TYPES` (`crates/rest/src/dashboard.rs:35-44`) is a fixed set of eight clinical types — Patient, Observation, Encounter, Condition, MedicationRequest, DiagnosticReport, Procedure, AllergyIntolerance. A tenant whose data is entirely outside that list sees an empty chart and a legend of zeros, with no way to chart what it actually has.

**6. The chart is invisible to screen readers.** `index.html:73` is `<svg class="chart" role="img" aria-hidden="true">` — `role="img"` with **no accessible name**, and `aria-hidden="true"` removing it from the accessibility tree entirely. There is no table or text alternative. Any interactive version makes this worse, so it needs solving as part of the work.

**7. No empty state.** When `has_data` is false the template draws axes and nothing else (`index.html:77-79`) — no message explaining why.

**8. Silent fallback to fabricated data.** `build_index_page` (`crates/ui/src/lib.rs:1294-1296`) falls back to `sample_snapshot(window)` when no provider answers, which generates plausible invented curves. The provider is registered in `crates/rest/src/lib.rs:595`, so the normal `hfs` path is fine — but when it isn't, the page presents made-up numbers indistinguishable from real ones. Confirm this can never be silently shown, or label it.

## Scope questions for the mockup

- Multi-series overlay, or one series at a time as now?
- Hover tooltip contents — value, bucket timestamp, delta vs cumulative?
- Is the type pill a dropdown that replaces the legend, or do both stay?
- What does "expand" do — a larger modal view, fullscreen, something else?
- Area fill, point markers, line styling, and how the palette handles multiple series in light and dark.

## Constraints

- **No SPA framework, no bundler, no npm** for browser code, and **no off-origin requests** — so no charting library from a CDN. Any interactivity is hand-written vanilla JS in a pinned asset, per `crates/ui/README.md`.
- **Keep it working without JavaScript.** The window selector and legend are links today; enhancement must layer on top rather than replace them.
- All user-visible strings go in `locales/{en,es,de}/main.ftl`.
- Theme-aware: colors come from the CSS custom properties, correct in light and dark.

## Acceptance criteria

- [ ] Every control on the card either does something or is removed — no inert pill, no dead expand button.
- [ ] Hovering the series gives a readable value and time.
- [ ] The chart can plot the resource types a tenant actually has, not only a fixed list of eight.
- [ ] An explicit empty state when there is no data.
- [ ] The chart is accessible: an accessible name, and a text or tabular alternative to the visual series.
- [ ] No fabricated sample data is ever presented as real.
- [ ] Still functional with JavaScript disabled (cover in `crates/ui/e2e/tests/nojs/`).
- [ ] Correct in light and dark themes.
- [ ] e2e coverage beyond today's two assertions in `crates/ui/e2e/tests/dashboard.spec.ts` (window selector, legend).
- [ ] `a11y.spec.ts` passes.

## Pointers

- `crates/ui/templates/pages/index.html:28-100` — the chart card
- `crates/ui/src/lib.rs:1395+` — `build_chart`; `:352-368` — `ChartView`; `:1294-1296` — the sample fallback; `:1554+` — `sample_snapshot`
- `crates/observability/src/dashboard.rs` — `DashboardSnapshot`, `DashboardSeries`, `DashboardWindow`
- `crates/rest/src/dashboard.rs:35-44` — `DEFAULT_DASHBOARD_TYPES`
- `crates/ui/assets/app.css` — `.chart`, `.chart-card`, `.chart-legend`, `.window-picker`, `.grid-line`, `.axis-label`, `.series`
- Related: #553 (dashboard tenant/version coverage), #554 (dashboard job cards), #540 (Uptime)

angela-helios:
Closes #555. The "FHIR resources over time" card goes from decorative to functional, following the hi-fi mockup's resource picker design.

**Stacked on #572** (base `fix/543-stylesheet-unification`; GitHub retargets to `main` when it merges).

## The provider learns what a tenant actually stores

`DashboardSnapshot` now carries every type with at least one stored resource (via the existing `count_all_types`, largest first), and the caller passes the set to chart. Empty selection = the tenant's three largest types, skipping definitional/infrastructure resources (a server that seeds its own SearchParameters would otherwise chart those instead of clinical data — they stay available in the picker). Selections are capped at three (the mockup's choice; also bounds the per-type delta queries) and join the snapshot cache key. The hardcoded eight-type list is gone.

## Every control does something

- **The type pill is the picker** (mockup `resPick`): a `details` menu of checkbox rows — every stored type with its count, a typeahead filter pinned above a scrolling list, each option a plain link that toggles the type. Selecting past the cap swaps the oldest series out. The pill label stays one name wide (`Encounter +2`).
- **Expand works**: a link that re-renders the taller plot server-side (520 viewBox) and collapses back.
- **Multi-series**: up to three series on one shared y scale, each with its own `--series-N` palette slot; the legend entries carry the color dot and the total, and double as remove links while more than one is plotted.
- **Hover tooltip**: a guide line plus a per-series readout of the nearest bucket (date, name, value). The data rides in an inert `application/json` carrier with server-computed SVG coordinates — the script does no chart math.
- **Windows** (1h/24h/30d) keep the charted set and expansion across switches.

## Honesty and access

- The SVG has an accessible name (title + plotted types) instead of `aria-hidden`, and the same numbers render as text: a details-folded table of the sampled buckets, one column per series.
- An explicit empty state replaces the silent bare axes.
- The placeholder snapshot (no provider registered) renders under a visible "sample data" notice instead of posing as real.
- Everything except the tooltip works with JavaScript disabled — picker, windows, expand, legend, and the table are links and native disclosure.
- All strings in en/es/de.

## Tests

- `dashboard.spec.ts`: 2 assertions → the whole card (accessible name, window switching, picker toggling with the cap, legend removal, filter, tooltip in/out, expand round-trip, tabular alternative), plus a `nojs` spec walking the same controls with JavaScript off.
- Provider unit tests updated (empty store → empty state; the observability cache tests catch the new trait shape).
- Full hermetic suite locally: **128 passed**, including axe (both themes) over the new card and the design-system guard over the new classes.

## Demo seeding

Verified against a seeded store: ~1.2k resources across 8 types created through ordinary batch bundles, with their history rows backdated across 30 days so every window shows real growth curves. The two scripts (batch seeder + SQLite backdater) are local dev tooling — happy to commit them under a dev/ path if we want them in-repo.


angela-helios:
Heads-up on the parallel merges: your main catch-up and mine landed on this branch at the same time — reconciled in `83685f669`+. One deliberate call: your `status-chip` family for the export job chips got folded into the existing `.tag` status-pill vocabulary instead (`tag--in-progress/complete/failed/cancelled`, defined next to the other status tags) — #572 exists to keep one pill family, and the design-system guard now enforces exactly that. Everything else from your merge is kept as-is. Full suite locally: 133 passed, including the guards over the export pages.

smunini (approved):

Issue #357, opened 2026-07-23
Subscription status transitions are never persisted — runtime state is lost on restart and the stored resource contradicts $status


mauripunzueta:
## Summary

Subscription runtime state — status, event counter, consecutive-failure counter — lives **only** in the engine's in-memory `DashMap`. Nothing is ever written back to the stored `Subscription` resource. Every runtime decision the server makes about a subscription is therefore lost on restart, and the stored resource permanently disagrees with the engine's live view.

## Evidence

`SubscriptionManager::update_status` (`crates/subscriptions/src/manager/mod.rs:299-328`) mutates the `DashMap` entry and returns; there is no storage write, and the manager holds no storage handle. The same is true of `increment_event_count` (`:331`) and `record_failure` (`:339`).

Every runtime transition goes through `update_status`, so every one of them is lost:

| Transition | Where | Trigger |
|---|---|---|
| `Requested → Active` | `engine/mod.rs:646` | handshake succeeded |
| `→ Error` | `engine/mod.rs:549, 599, 616, 663, 681, 714` | handshake build failure, permanent error, retries exhausted, unsupported/unconfigured channel |
| `→ Error` | `engine/mod.rs:872` | `consecutive_failures >= error_threshold` |
| `→ Off` | `engine/mod.rs:866` | `consecutive_failures >= off_threshold` |

A grep for a storage write from the engine or manager returns nothing.

## Impact

**1. The REST API contradicts itself.** `GET /Subscription/{id}/$status` reads the engine (`handlers/subscriptions.rs:41-47`, `engine.manager().get_subscription(...)`) and reports the live status. `GET /Subscription/{id}` reads storage and reports the stored one. For any subscription created as `requested` and since activated, these disagree indefinitely — `$status` says `active`, the resource read says `requested`.

**2. The delivery circuit-breaker is amnesiac.** A subscription the server turned `off` after `off_threshold` consecutive delivery failures — the mechanism that exists to stop hammering a broken endpoint — reverts to its stored status on restart and resumes delivering. Same for `error`. A crash-looping server never stops retrying a dead subscriber.

[...]

PR #585, opened 2026-08-19
feat(ui,rest,observability): Subscriptions operator page — read-only dashboard over the live engine

### LINKED ISSUE #580: ui: Subscriptions page — read-only operator dashboard (status cards, delivery table)
## Summary

Implement the Subscriptions page in the web UI, following Brett's new design (walked through in the review recording). **Read-only at this stage**: an operator dashboard over the topic-based subscriptions engine — no create/edit/delete, no channel configuration. The sidebar's Subscriptions entry currently renders as a "coming soon" placeholder.

## The design

One screen, same shell as the other workspaces (sidebar + content pane):

**Four highlight cards** across the top, each a status count with a one-line qualifier:

| Card | Figure | Qualifier |
|---|---|---|
| FAILING | count of subscriptions in `error` | "Needs attention" |
| IDLE | count with no connected clients (websocket topics nobody is listening to) | "No clients" |
| ACTIVE | count in `active`, delivering | "delivering" |
| DELIVERED IN 24 H | total notifications delivered in the last 24 hours (e.g. 6,303) | first-try success rate (e.g. "96.4% first try") |

**A results table** underneath, with a sort control ("SORT: Most recent") in its header:

| SUBSCRIPTION | CHANNEL | STATUS | LAST 24 HRS — SENT | FAILS STREAK |
|---|---|---|---|---|
| topic-ish name (e.g. `encounter-start`) | `rest-hook` / `websocket` / … | chip: `active` / `error` / `0 clients` | notifications sent (e.g. 4,182) | consecutive failures (e.g. 7; `-` when n/a) |

Status renders as a colored chip; the FAILING/IDLE/ACTIVE cards and the STATUS column must agree on vocabulary and color. Rows for websocket topics with no listeners show the `0 clients` state rather than pretending to deliver.

## Data sources — what exists, what doesn't

- **Subscription resources** are searchable through the ordinary FHIR API, and `GET /Subscription/{id}/$status` returns the standard `SubscriptionStatus` (`crates/rest/src/handlers/subscriptions.rs`); `$events` exists too.
- The engine's manager tracks the **consecutive delivery-failure streak** and the event counter per subscription, and drives active→error transitions (`crates/subscriptions/src/manager/mod.rs`, `manager/status.rs`) — the FAILS STREAK column and the FAILING card map to this.
- **Gap:** nothing currently persists a per-window delivery count or a first-try success rate, so the "DELIVERED IN 24 H" card and the "LAST 24 HRS — SENT" column need a read path that does not exist yet — a rolling delivery counter or a queryable notification log in the engine. Scoping that read path (and keeping it cheap) is part of this issue; if it turns out to be substantial, split it out and land the page with the columns it can honestly fill.
- Websocket listener counts ("0 clients" / IDLE) need the channel to expose its current connection count per topic.

Whatever the sources, the page should follow the UI crate's conformance pattern: fetch through the server's own HTTP surface (loopback self-call) or a read path handed to `mount()` — the UI crate does not reach into the engine directly.

## Constraints

- Server-rendered Askama + htmx; no SPA, no CDN, no inline scripts (crate rules of the road).
- Progressive enhancement: the page and its sort control must work without JavaScript; any polling refresh (the Bulk Import pattern) layers on top.
- Compose the component vocabulary (`crates/ui/README.md` § Component vocabulary): `.page-head__title`, stat cards, `.table-wrap` > `.data-table`, `.tag` status chips. No new class for an existing primitive — `design-system.spec.ts` enforces it.
- All strings in `locales/{en,es,de}/main.ftl` (the `nav-subscriptions` key already exists).
- Multi-tenant: figures are for the active tenant only.

## Acceptance criteria

- [ ] `/ui/subscriptions` renders the four cards and the table from live engine/storage data; the sidebar entry becomes a real nav link.
- [ ] Status vocabulary is consistent between cards, chips, and the underlying `Subscription.status` values; error streaks match the manager's counters.
- [ ] Honest empty and unavailable states: no subscriptions ≠ engine disabled — say which one it is.
- [ ] Delivery-window figures are real or absent — no placeholder numbers presented as live (same rule as #555).
- [ ] Read-only: no mutating controls anywhere on the page.
- [ ] Sort control works without JavaScript (links or a plain form).
- [ ] Route added to `e2e/pages/routes.ts` (covers a11y both themes, no-cdn, design-system) plus a dashboard-style spec for the page's own behavior and a nojs check.
- [ ] `cargo test -p helios-ui` covers the render with seeded subscriptions.

## Pointers

- `crates/subscriptions/src/manager/` — status transitions, failure streaks, event counters
- `crates/subscriptions/src/channels/{rest_hook,websocket}.rs` — channel delivery, connection state
- `crates/rest/src/handlers/subscriptions.rs` — `$status` / `$events` operations
- `crates/ui/src/bulk_import.rs` — the closest existing workspace pattern (cards + table + polling)
- `crates/ui/templates/pages/_scaffold.html` — starting point
- Related: #555 (dashboard chart, same honesty rules), #554 (dashboard job cards)

angela-helios:
Closes #580. The sidebar's Subscriptions placeholder becomes the read-only operator dashboard from Brett's design.

**Stacked on #572** — based on `fix/543-stylesheet-unification` for the shared vocabulary and guards, targeting `main` so the Development link on #580 is live from day one; the diff shrinks to just this feature when #572 merges.

## What renders

- **Four status cards**: FAILING (needs attention), IDLE (websocket subscriptions with no connected clients), ACTIVE (delivering), and NOTIFICATIONS since registration.
- **The table**: every registered subscription — id and topic (full canonical URL in the tooltip), channel + endpoint, a status chip, the event counter, and the consecutive-failure streak (`—` where there is no delivery to fail).
- **Sort as plain links**: status (default, worst first), most sent, fail streak. Read-only throughout; works without JavaScript.

## The read path

Mirrors the dashboard chart's pattern: a plain-data `SubscriptionsProvider` in `helios-observability`, implemented in `helios-rest` over the live engine (the manager's in-memory inventory plus the WebSocket manager's connection counts) and registered exactly where the engine is built. Registration doubles as the **"subscriptions advertised"** signal from the design note: the nav entry and the page only exist when the engine runs; otherwise the entry hides and the route renders an explained unavailable state.

## Honesty rules (same as #555)

- A websocket subscription nobody is connected to shows **0 clients** even while its resource says `active` — its notifications go nowhere, and that outranks the stored status.
- The design's 24-hour delivery figures are deliberately absent: the engine does not persist delivery history yet, and this page shows real numbers or none. That instrumentation (a rolling delivery counter or queryable notification log, plus first-try success rate) is the follow-up half of #580.

## Verified against the live engine

Manual end-to-end on a running server: an R4 backport topic (`Basic`), a rest-hook whose endpoint rejects the handshake (405) lands in `error` — consistently across the UI chip, the persisted resource (status write-back, #357), and `$status` — while a rest-hook against a real sink handshakes `requested → active` and delivers notification Bundles that the sink actually receives, with the row's SENT counter matching. The idle websocket accumulates SENT with 0 clients, exactly the design's third example row.

## Tests

- **Rust** (`subscriptions_http.rs`): the full state machine over the mounted router with an injected provider — unavailable (nav hidden) → live dashboard (cards summed, worst-first order, chips, streaks, the 0-clients state) → link-based re-sort. Plus a provider unit test in observability.
- **Playwright**: the advertised-but-empty engine the e2e server now runs — cards at zero, empty table, sort links, the nav entry — skipping gracefully on a local binary built without the `subscriptions` feature, which the ui-test workflows now enable. The route joins the shared guard list (a11y both themes, no-cdn, design-system).
- Full hermetic suite locally: **128 passed**.


angela-helios:
Coordination note: we were both pushing main catch-ups to this branch at the same time — everything is reconciled as of `edb2d65a6` (your rustfmt pass included). The tree is the one the full local suite passed at 140: subscriptions + the merged editor/export/stylesheet work together. Backing off the branch now — it is all yours.

smunini (approved):

PR #616, opened 2026-08-20
feat(validator): warn when a known extension sits outside its declared context

### LINKED ISSUE #615: validator: extension context (placement) is not checked at validation
From Steve and Angela's extension-authoring review session. Two related questions came up; this issue tracks the one that is a real gap.

## What the spec says

Per FHIR R4 [Defining Extensions](https://hl7.org/fhir/R4/defining-extensions.html), `StructureDefinition.context` declares **where** an extension may be used — `context.type` is `element` (a resource or element path, e.g. `Patient` for the resource root or `Patient.name` for one element), `fhirpath`, or `extension`. Placement is therefore *not* merely resource-scoped: an element-type context can pin an extension to one spot in the data model, and using it elsewhere is invalid. (Separately, [ext-1](https://hl7.org/fhir/R4/extensibility.html) requires every extension to carry either a `value[x]` or nested extensions — a bare `{url}` is structurally invalid; that half is covered: the invariant rides the Extension type schema, and #614 makes the Guided form author values correctly.)

## Where we stand

- The **editor picker** already constrains placement: `extensions_applicable` filters by the resource type, the dotted element path, the element's type name, and `Element`/`Resource`/`DomainResource` (`crates/ui/src/editor.rs:563`). So the guided path answers Steve's "do we constrain these properly" — yes, at authoring time.
- The **ad-hoc add-by-URL** path and, more importantly, the **validation engine** do not check context: a resource carrying a known extension outside its declared context passes `$validate` today. Nothing in `engine/` consults `FhirSchema.context` (the converter carries it — the picker reads it).

## Proposal

When the walk meets an `extension`/`modifierExtension` item whose url resolves to a schema with a declared `context`, check the current element path against the element-type contexts (fhirpath contexts can start warn-only) and report a warning-severity issue on mismatch — warning, not error, because context checking on ad-hoc/unknown extensions must stay permissive.

angela-helios:
Closes #615. The gap surfaced in Steve and Angela's extension-authoring session: the picker constrains placement, `$validate` never did.

## What it does

The walk checks each `extension`/`modifierExtension` item whose url resolves to a registry schema with declared contexts against the index-free parent element path (`Patient`, `Patient.name`), and reports a **warning-severity** `ExtensionContext` issue on mismatch. Element-type contexts name one exact spot — `Patient` is the resource root, not "anywhere in Patient".

## Deliberately conservative

Judged only when every declared context reads as a resource-rooted element path. Datatype-named contexts (`HumanName`), the `Element`/`Resource`/`DomainResource` catch-alls, and FHIRPath expressions make the extension unjudgeable here (the parent's *type* is not in hand during the walk) and pass silently — as do unknown urls and relative sub-extension urls. No false positives by construction; the common miss (a Patient extension pasted into an Observation) is caught.

## Severity plumbing

- Warnings map to warning-severity OperationOutcome issues at `$validate` (existing `rest/validation.rs` mapping) and never trip enforce mode (which rejects on error-severity only).
- The Guided editor's issue count blocks saving, so it now retains only error-severity issues — context guidance belongs to `$validate`, not between the user and the save button.

## Tests

- `extension_context_mismatch_warns`: birthPlace at the Patient root → silent; on an Observation → exactly one warning at `Observation.extension.0`; unknown url → silent. Runs against the real R4 core registry.
- Full `cargo test -p helios-fhir-validator` (conformance fixtures included) and `-p helios-ui` green; clippy clean.

### DIFF
--- a/crates/fhir-validator/src/bin/validator_cli.rs
+++ b/crates/fhir-validator/src/bin/validator_cli.rs
@@ -134,6 +134,7 @@ fn main() -> ExitCode {
         profiles: args.profiles.clone(),
         use_meta_profiles: !args.no_meta_profiles,
         unknown_profile: UnknownProfilePolicy::Warn,
+        ..Default::default()
     };
     let outcome = validator.validate_sync(&resource, &opts);
 
--- a/crates/fhir-validator/src/engine/errors.rs
+++ b/crates/fhir-validator/src/engine/errors.rs
@@ -60,6 +60,10 @@ pub enum ErrorKind {
     /// A profile named in `meta.profile` or the caller's profile list could
     /// not be resolved.
     UnknownProfile,
+    /// A known extension used outside its declared context (#615). Always
+    /// warning severity: context matching is conservative and ad-hoc
+    /// extensions must stay permissive.
+    ExtensionContext,
 }
 
 /// Issue severity. Internal only — deliberately not serialized, so the
@@ -158,6 +162,13 @@ pub(crate) fn msg_expected_object_comma(json_type: &str) -> String {
 }
 
 /// [reference] `expected {min} > {actual_len}`
+pub(crate) fn msg_extension_context(url: &str, at: &str, contexts: &[String]) -> String {
+    format!(
+        "extension {url} is not declared for use at {at}; its contexts are [{}]",
+        contexts.join(", ")
+    )
+}
+
 pub(crate) fn msg_min(min: u64, actual_len: usize) -> String {
     format!("expected {min} > {actual_len}")
 }
--- a/crates/fhir-validator/src/engine/mod.rs
+++ b/crates/fhir-validator/src/engine/mod.rs
@@ -56,6 +56,13 @@ pub struct ValidationOptions {
     pub use_meta_profiles: bool,
     /// What to do when a profile reference cannot be resolved.
     pub unknown_profile: UnknownProfilePolicy,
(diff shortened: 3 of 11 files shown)


angela-helios:
The spec-examples sweep rejected the always-on design, and it was right to — updating the PR's claims:

1. **Element-level judging is unsound without SD-path mapping.** An instance path `Questionnaire.item.item` never equals the declared context `Questionnaire.item`; matching needs the StructureDefinition path (recursion collapsed via contentReference), which the walk does not have. Hundreds of false positives across the questionnaire examples.
2. **Even root-only judging flags canonical HL7 content wholesale.** The publisher places `structuredefinition-wg`/`-fmm`/`-standards-status`/`-normative-version` (declared context: `StructureDefinition`) on CodeSystem/ValueSet roots throughout the published spec — e.g. every `codesystem-*.json` example.

So the check landed as: **resource-root extensions only, behind `ValidationOptions::check_extension_context`, default off.** A profile author can opt in; a default `$validate` is unchanged, and the sweep pins that (all three baselines intact with the option off, and the option-on path is unit-tested). Wiring the option to something reachable — an `HFS_VALIDATION_*` knob or a `$validate` parameter — can ride a follow-up if we want it exposed.

Also for the record: an earlier push accidentally committed the re-downloaded R6 example corpus (1,375 files of line-ending churn from `crates/fhir/build.rs`); it was force-replaced with the clean history before review.

smunini (approved):

Issue #456, opened 2026-07-30
search: day-precision date comparisons are off at the boundaries (eq matches nothing; ge excludes the day itself; lt/le include the next midnight)


angela-helios:
Found during the #448 manual sanity pass over Synthea data (SQLite backend, main @ 2026-07-30). Month- and year-precision date searches behave correctly; **day-precision** comparisons are wrong at every boundary.

### Repro
Load a Patient with `birthDate: "1995-10-02"` (cohort below has 10 patients; dates listed at the end).

| Query | Expected | Actual | Broken semantics |
|---|---|---|---|
| `/Patient?birthdate=1995-10-02` | 1 | **0** | eq range comes out empty |
| `/Patient?birthdate=eq1995-10-02` | 1 | **0** | same |
| `/Patient?birthdate=ge1995-10-02` | 7 | **6** | ge excludes the named day (acts like gt) |
| `/Patient?birthdate=le1995-12-31` | 4 | **5** | le includes a patient born **1996-01-01** (closed bound at next midnight) |
| `/Patient?birthdate=lt1996-01-01` | 4 | **5** | lt includes the boundary day (acts like le) |
| `/Patient?birthdate=1995-10` | 1 | 1 | month precision OK |
| `/Patient?birthdate=1995` | 2 | 2 | year precision OK |
| `/Patient?birthdate=ge1995-10-02&birthdate=le1995-10-02` | 1 | **0** | consistent with broken ge |

Reading the pattern: the comparisons use the wrong end of the day-precision parameter interval — `ge`/`lt` compare against the parameter's **end** where the spec needs its **start**, and `le` closes its upper bound exactly at the next day's midnight (so `le1995-12-31` captures `1996-01-01T00:00`). `eq`, built from both bounds, ends up empty. Per FHIR date search semantics, a day-precision parameter is the interval `[00:00, 24:00)` of that day and eq/ge/le/gt/lt are defined against the resource value's own interval.

Chained variants inherit the bug (`/Condition?patient.birthdate=1995-10-02` → 0 with matching data present).

Cohort birthdates: 1975-03-21, 1991-09-26, 1994-10-03, **1995-10-02**, 1996-01-01, 1998-09-28, 2003-06-30, 2005-10-25, 2011-11-28, 2022-05-07.

Found on SQLite; other backends not yet checked (the #448 pass will). Worth checking the shared date-range translation rather than the backend, given chained search reproduces it.

Part of the #448 sanity pass.

[...]

PR #621, opened 2026-08-20
test(persistence): the #456 date-boundary table runs on every search backend

### LINKED ISSUE #519: search: verify day-precision date comparison semantics on all non-SQLite backends (PG, ES, Mongo, composite)
Follow-up to #456 / #463.

PR #463 fixed day-precision date search boundaries **for SQLite only**. Its scope note is explicit:

> SQLite only. Postgres parses dates into real timestamps on a separate path (`parse_date_value`) and needs no change; the Elasticsearch and Mongo date handlers are unverified.

We need to confirm — with tests, not just reasoning — that the same date-comparison semantics hold on every other backend.

## Scope

For each backend other than SQLite:

- **PostgreSQL** — believed correct via `parse_date_value` (real `timestamptz` comparisons), but that belief is currently untested at the boundaries.
- **Elasticsearch** — date handler unverified.
- **MongoDB** — date handler unverified.
- **Composite storage** — whichever backend serves search in a composite configuration must give the same answers.

## What to verify

Reproduce the exact boundary table from #456 against each backend, using the same 10-patient cohort (birthdates: 1975-03-21, 1991-09-26, 1994-10-03, **1995-10-02**, 1996-01-01, 1998-09-28, 2003-06-30, 2005-10-25, 2011-11-28, 2022-05-07):

| Query | Expected |
|---|---|
| `/Patient?birthdate=1995-10-02` | 1 |
| `/Patient?birthdate=eq1995-10-02` | 1 |
| `/Patient?birthdate=ge1995-10-02` | 7 |
| `/Patient?birthdate=le1995-12-31` | 4 |
| `/Patient?birthdate=lt1996-01-01` | 4 |
| `/Patient?birthdate=1995-10` | 1 |
| `/Patient?birthdate=1995` | 1 |
| `/Patient?birthdate=ge1995-10-02&birthdate=le1995-10-02` | 1 |

Note: `birthdate=1995` is **1**, not 2 — #463 established that the old answer of 2 wrongly included the 1996-01-01 birth via a text-compare leak. Any backend returning 2 there is also wrong.

Also cover the cases #463 found were broken beyond the plain parameter handler on SQLite, since they are separate code paths per backend:

1. **Chained search** — `/Condition?patient.birthdate=1995-10-02` must agree with the direct query (chained search compared dates with a raw text `=` on SQLite).
2. **`_filter`** — date comparisons inside `_filter` must agree with the equivalent plain parameter.
3. **Full-precision instants** — a resource stored as `2016-01-23T13:07:42-04:00` must match the same instant expressed in UTC, and full-precision `eq` must not build an unsatisfiable range.
4. **Multi-value `eq`** — repeated date params on one search (the SQLite fix also cleared a bind-parameter numbering collision there).
5. **Mixed stored precisions** — resources whose own `birthDate` precision differs (day vs. full datetime) must both be compared correctly against the same query.

## Deliverable

- Integration tests pinning the boundary table per backend (mirroring the 4 tests added to `crates/rest/tests/search_integration.rs` in #463), so this can't regress silently.
- Fixes for any backend that fails, or a separate issue per broken backend if the fixes are non-trivial.
- A note in the issue confirming Postgres is genuinely correct (or not) rather than assumed.

Related: #448 (Synthea sanity pass that surfaced this), #456, #463.

angela-helios:
Closes #519.

A backend-agnostic suite (`tests/search/date_boundary_suite.rs`, the same `#[path]` arrangement as the if-match and tenant-fidelity suites) seeds the ten-patient #456 cohort and drives the eight-query boundary table through each backend's real `SearchProvider::search` path — including the subtle `birthdate=1995` = **1** row (#463's text-comparison catch) and the AND-ed `ge…&le…` pair, which is two parameters, not one two-valued parameter. Eventually-consistent backends get a visibility poll before judgment.

## Verdicts

| backend | result |
|---|---|
| PostgreSQL | **whole table correct** (verified locally — the `parse_date_value` belief is now a test) |
| MongoDB | **whole table correct** (verified locally, full registry so `birthdate` indexes) |
| Elasticsearch | wired identically; **CI is the arbiter** — the local ES image's bundled JDK crashes on this machine's WSL2 cgroup layout (`CgroupInfo.getMountPoint` NPE) before ES even boots, an environment incompatibility unrelated to dates |
| SQLite | already pinned at the SQL-condition level by #463's own tests |
| Composite | serves search through one of the four engines above, each of which the table now pins |

No product changes — pure verification, and (pleasant surprise) no new date bugs: the class of defect #463 fixed in SQLite does not recur elsewhere. The chained-date sweep for ES/Mongo noted in #619 remains follow-up work — their chain paths need a live-container chained harness that does not exist yet.

### DIFF
--- a/crates/persistence/src/backends/mongodb/search_impl.rs
+++ b/crates/persistence/src/backends/mongodb/search_impl.rs
@@ -3,7 +3,7 @@
 use std::collections::HashSet;
 
 use async_trait::async_trait;
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, Datelike, Utc};
 use helios_fhir::FhirVersion;
 use mongodb::{
     Cursor,
@@ -45,6 +45,22 @@ fn bson_to_chrono(dt: &BsonDateTime) -> DateTime<Utc> {
 
 fn chrono_to_bson(dt: DateTime<Utc>) -> BsonDateTime {
     BsonDateTime::from_millis(dt.timestamp_millis())
+}
+
+/// The exclusive end of the period a partial date names: `1995` → 1996-01-01,
+/// `1995-10` → 1995-11-01, `1995-10-02` → 1995-10-03. `None` for values that
+/// carry a time component — those are instants, not periods.
+fn implied_period_end(raw: &str, start: DateTime<Utc>) -> Option<DateTime<Utc>> {
+    match raw.len() {
+        4 => Some(
+            start
+                .with_year(start.year() + 1)
+                .expect("year+1 stays in range"),
+        ),
+        7 => Some(start + chrono::Months::new(1)),
+        10 => Some(start + chrono::Duration::days(1)),
+        _ => None,
+    }
 }
 
 fn parse_date_for_query(value: &str) -> Option<DateTime<Utc>> {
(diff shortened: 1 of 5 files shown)


angela-helios:
The first CI run caught a real one — updating the verdict table:

MongoDB's `birthdate=1995-10` = 0 was **not** flake: my local pass was a UTC-4 accident, and under `TZ=UTC` it reproduces deterministically. The cause is semantic: `build_date_filter` compared `eq` against the exact start instant, so month/year queries matched nothing born after midnight on the first. Partial dates name *periods* — the filter now mirrors the #463 SQLite mapping arm for arm (`eq` → `[start, end)`, `ne` → outside, `gt`/`sa` → from the period's end, `le` → to it). Full Mongo suite (74) + lib (805) green under UTC.

So the PR is no longer verification-only: **the suite found and fixed exactly the class of bug #519 was hunting.** PostgreSQL remains clean; ES still rides this CI run.

smunini (approved):

PR #643, opened 2026-08-24
fix(search): correct SQLite :missing behavior and backend parity

### LINKED ISSUE #632: SQLite :missing=true searches always return an empty result set
## Summary

SQLite search returns an empty Bundle for valid `:missing=true` queries, even when resources without the requested value exist.

The request succeeds with `200 OK`, making the failure indistinguishable from a legitimate empty result. This creates silent false negatives in data-quality checks, cohort construction, follow-up workflows, and any client that depends on presence or absence searches.

FHIR R4 defines `:missing=true` as selecting resources without a value for the parameter and `:missing=false` as selecting resources with a value. The modifier applies to ordinary search parameters except combination parameters. [FHIR R4 search modifiers](https://hl7.org/fhir/R4/search.html#modifiers)

The defect is in SQLite query generation. The REST layer parses the request correctly, and the UI only displays the empty response returned by the server.

## Environment

- HFS version: `0.2.1`
- Revision: [`3b46edc07b343ff9ea97a694e80147439e89e0ad`](https://github.com/HeliosSoftware/hfs/commit/3b46edc07b343ff9ea97a694e80147439e89e0ad)
- FHIR version: R4
- Confirmed backend: SQLite

## Minimal reproduction

Start HFS with native SQLite search:

```bash
export HFS_STORAGE_BACKEND=sqlite
export HFS_DATABASE_URL=:memory:
export HFS_BASE_URL=http://127.0.0.1:8080

hfs
```

Create one Patient with a birth date:

```bash
curl -fsS -X PUT "$HFS_BASE_URL/Patient/with-date" \
  -H 'Content-Type: application/fhir+json' \
  --data '{
    "resourceType": "Patient",
    "id": "with-date",
    "birthDate": "1980-01-15"
  }'
```

Create another Patient without a birth date:

```bash
curl -fsS -X PUT "$HFS_BASE_URL/Patient/without-date" \
  -H 'Content-Type: application/fhir+json' \
  --data '{
    "resourceType": "Patient",
    "id": "without-date",
    "active": true
  }'
```

Run the absence search:

```bash
curl -fsS \
  "$HFS_BASE_URL/Patient?birthdate:missing=true&_total=accurate&_count=100" |
  jq '{total, ids: [.entry[]?.resource.id]}'
```

### Current result

```json
{
  "total": 0,
  "ids": []
}
```

### Expected result

```json
{
  "total": 1,
  "ids": [
    "without-date"
  ]
}
```

Run the presence control:

```bash
curl -fsS \
  "$HFS_BASE_URL/Patient?birthdate:missing=false&_total=accurate&_count=100" |
  jq '{total, ids: [.entry[]?.resource.id]}'
```

The control currently returns the expected resource:

```json
{
  "total": 1,
  "ids": [
    "with-date"
  ]
}
```

## Runtime evidence

A larger SQLite dataset contained 74 active Patients. Index inspection showed that 73 had a `birthdate` index entry and one did not.

| Parameter | Present resources | Missing resources | `missing=true` result |
|---|---:|---:|---:|
| `birthdate` | 73 | 1 | 0 |
| `identifier` | 73 | 1 | 0 |
| `telecom` | 34 | 40 | 0 |
| `address` | 33 | 41 | 0 |
| `general-practitioner` | 3 | 71 | 0 |

Direct retrieval confirmed that the missing Patient existed and did not contain `birthDate`. `birthdate:missing=false` returned the other 73 Patients and excluded that resource.

Saved queries showing birthdate missing true returning zero results

## Root cause

REST correctly parses `birthdate:missing=true` into `SearchModifier::Missing` with the value `true`:

- [`search_query_builder.rs`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/rest/src/extractors/search_query_builder.rs#L324-L435)

SQLite then constructs a valid resource-level absence predicate:

```sql
resource_id NOT IN (
  SELECT resource_id
  FROM search_index
  WHERE tenant_id = ?1
    AND resource_type = ?2
    AND param_name = 'birthdate'
)
```

That predicate is subsequently placed inside the generic parameter wrapper:

```sql
resource_id IN (
  SELECT resource_id
  FROM search_index
  WHERE tenant_id = ?1
    AND resource_type = ?2
    AND param_name = 'birthdate'
    AND resource_id NOT IN (
      SELECT resource_id
      FROM search_index
      WHERE tenant_id = ?1
        AND resource_type = ?2
        AND param_name = 'birthdate'
    )
)
```

The resulting condition requires a resource ID to belong to the set of resources with a `birthdate` index row while simultaneously excluding every ID in that set. No resource can satisfy both conditions.

Relevant code:

- Generic SQLite parameter wrapper: [`query_builder.rs`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/persistence/src/backends/sqlite/search/query_builder.rs#L379-L445)
- SQLite missing predicate: [`query_builder.rs`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/persistence/src/backends/sqlite/search/query_builder.rs#L720-L795)
- Result retrieval and totals: [`search_impl.rs`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/persistence/src/backends/sqlite/search_impl.rs#L120-L215)

`missing=false` works for ordinary parameters because both the outer and inner conditions select resources with the index entry.

## Affected SQLite scope

The contradiction applies to every ordinary parameter type routed through the generic SQLite value handler:

- String
- Token
- Date
- Number
- Quantity
- Reference
- URI

SQLite currently advertises `missing` support for these types:

- [`backend.rs`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/persistence/src/backends/sqlite/backend.rs#L736-L768)

Runtime reproduction confirmed Date, String, Token, and Reference parameters. Number, Quantity, URI, and Resource metadata parameters are affected through the same code path but still need runtime coverage.

The same path can affect metadata parameters such as `_tag`, `_profile`, `_security`, and `_source`.

## Composite parameter behavior

SQLite also advertises `missing` for composite parameters, although FHIR R4 excludes combination parameters from the modifier.

Composite dispatch happens before the missing handler. The Boolean value is interpreted as a composite search value and fails closed.

This was reproduced with two Observations, one containing `code-value-quantity` and one without it:

| Query | HTTP status | Total |
|---|---:|---:|
| Unfiltered Observation search | 200 | 2 |
| `code-value-quantity:missing=true` | 200 | 0 |
| `code-value-quantity:missing=false` | 200 | 0 |

Capability advertisement and request validation should be aligned with the intended composite behavior.

## Backend scope

| HFS storage mode | Search implementation | Status |
|---|---|---|
| `sqlite` | SQLite | Affected and runtime-confirmed |
| `postgres` | PostgreSQL | Not affected; integration test passed |
| `mongodb` | MongoDB | `:missing` is unsupported and returns `400` |
| `s3` | None | Not applicable; standalone S3 does not support search |
| `sqlite-elasticsearch` | Elasticsearch | Expected unaffected; shared code path confirmed, runtime pending |
| `postgres-elasticsearch` | Elasticsearch | Expected unaffected; shared code path confirmed, runtime pending |
| `mongodb-elasticsearch` | Elasticsearch | Expected unaffected; shared code path confirmed, runtime pending |
| `s3-elasticsearch` | Elasticsearch | Expected unaffected; shared code path confirmed, runtime pending |

There is no standalone Elasticsearch HFS mode.

Composite storage selects the backend registered with the Search role. A custom composite without a dedicated search backend can inherit the behavior of its primary provider, including the SQLite defect.

Relevant routing code:

- [`composite/storage.rs`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/persistence/src/composite/storage.rs#L471-L551)

### PostgreSQL

PostgreSQL handles `SearchModifier::Missing` before ordinary type-specific filtering and produces a top-level `IN` or `NOT IN` condition.

- [`postgres/query_builder.rs`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/persistence/src/backends/postgres/search/query_builder.rs#L509-L522)
- [`postgres/query_builder.rs`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/persistence/src/backends/postgres/search/query_builder.rs#L665-L687)

The existing Docker integration test passed for both Boolean values and asserted exact IDs.

### MongoDB

MongoDB does not advertise `missing` for any parameter type:

- [`mongodb/backend.rs`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/persistence/src/backends/mongodb/backend.rs#L681-L699)

Requests reach backend-specific validation and produce `400`, although some types report a typed-value parsing error instead of a clear unsupported-modifier error.

### Elasticsearch-backed modes

All supported Elasticsearch composites route search to Elasticsearch, independently of the CRUD backend.

Elasticsearch handles `SearchModifier::Missing` before normal type dispatch and creates a resource-level positive or negative nested query:

- [`elasticsearch/query_builder.rs`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/persistence/src/backends/elasticsearch/search/query_builder.rs#L186-L200)
- [`elasticsearch/modifier_handlers.rs`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/persistence/src/backends/elasticsearch/search/modifier_handlers.rs#L9-L57)

The query-generation unit tests pass. Live `:missing` integration coverage is still absent.

### Invalid Boolean values

The REST layer does not require the value to be exactly `true` or `false`.

SQLite treats any case-insensitive value other than `true` as `false`. Elasticsearch treats any value other than exact lowercase `true` as `false`.

Requests such as these can silently behave like `missing=false`:

```text
birthdate:missing=invalid
birthdate:missing=
birthdate:missing=true,false
```

Invalid and multiple values should return `400` with an `OperationOutcome`.

### Contained resources

SQLite index rows for contained resources use the container resource ID and set `is_contained = 1`. The current missing predicate does not restrict this flag.

A contained element with the same parameter name could therefore make a top-level resource appear to have that parameter. Presence checks should distinguish top-level and contained search contexts.

## Existing test gap

The current SQLite `missing=true` test passes with an empty result because it only checks properties of returned resources. It never asserts cardinality or the expected resource ID:

- [`modifier_tests.rs`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/persistence/tests/search/modifier_tests.rs#L35-L75)

The `missing=false` test has the same weakness:

- [`modifier_tests.rs`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/persistence/tests/search/modifier_tests.rs#L77-L113)

The unchanged `test_missing_true` test passes despite the confirmed defect.

PostgreSQL already has an exact-membership test that can serve as the model:

- [`postgres_tests.rs`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/persistence/tests/postgres_tests.rs#L2361-L2421)

## Recommended fix

Handle `SearchModifier::Missing` once at the parameter level, before composite and ordinary value handling. The resulting resource-level condition must bypass the generic wrapper that requires an index row for the same parameter.

The implementation should:

1. Preserve tenant and resource-type isolation.
2. Distinguish top-level and contained index rows.
3. Centralize the duplicated SQLite missing-query logic.
4. Validate `true` and `false` centrally in REST.
5. Align advertised support with actual behavior, particularly for composite parameters.

## Regression coverage

Add a shared exact-membership contract using one resource with a value and one without it.

The suite should cover:

- String, Token, Date, Number, Quantity, Reference, and URI
- `_tag`, `_profile`, `_security`, and `_source`
- Exact IDs and cardinality for both Boolean values
- Complement and disjointness of the `true` and `false` result sets
- Empty arrays, absent elements, and primitives with extensions but no value
- Updates from absent to present and present to absent
- Deletion and stale index cleanup
- Repeated parameters and an additional search criterion
- `_total=accurate`, `_summary=count`, `_count`, and cursor pagination
- Tenant isolation
- Identical logical IDs across resource types
- Contained-resource isolation
- Invalid, blank, and comma-separated Boolean values
- Capability advertisement for unsupported modifier and parameter combinations

Run the shared contract as follows:

- SQLite must pass all supported cases.
- PostgreSQL must retain its current correct behavior.
- Elasticsearch must pass through a live integration test.
- Each Elasticsearch composite should be smoke-tested with `HFS_COMPOSITE_SYNC_MODE=synchronous`; wait until both seeded resources are searchable before asserting a negative query.
- MongoDB must have an explicit unsupported-modifier contract with `400`.
- Standalone S3 should be skipped as unsupported.

## Acceptance criteria

- `missing=true` returns exactly the resources without a value for every supported ordinary parameter type.
- `missing=false` returns exactly the resources with a value.
- Both result sets are disjoint and their union equals the eligible unfiltered resource set.
- Totals, count-only responses, filtering, and pagination remain consistent.
- Updates and deletes change presence membership without leaving stale results.
- Tenant and contained-resource boundaries remain intact.
- Invalid values return `400` with a clear `OperationOutcome`.
- Capability metadata advertises only the modifier combinations the active search backend supports.
- Native PostgreSQL behavior remains unchanged.
- Elasticsearch-backed modes pass live regression coverage.
- MongoDB continues to reject the unsupported modifier explicitly.
- No UI change is required for the backend correction.

Gordex2014:
## Summary

- Resolve SQLite `:missing` at resource level before ordinary, composite, and type dispatch.
- Ignore contained index rows when checking top-level presence in SQLite and PostgreSQL.
- Use authoritative `_id` and `_lastUpdated` fields across SQLite, PostgreSQL, and Elasticsearch.
- Require exact `true` or `false` values and reject unsupported parameter combinations.
- Align capability behavior across SQLite, PostgreSQL, Elasticsearch, and MongoDB.

## Root cause

SQLite built the correct absence predicate, then wrapped it in a query that required an index row for the same parameter. `birthdate:missing=true` therefore required `birthdate` to be both present and absent.

The existing tests iterated over returned resources without asserting cardinality or IDs, so an empty result passed.

## Backend behavior

| Backend | Result |
| --- | --- |
| SQLite | Fixed |
| PostgreSQL | Existing behavior retained; contained-row isolation and metadata parity added |
| Elasticsearch | Existing ordinary behavior retained; metadata dispatch corrected |
| MongoDB | Explicit unsupported modifier |
| S3 | Search requires Elasticsearch |

## Verification

- `cargo test -p helios-persistence --features R4B,R5,R6,sqlite,postgres,elasticsearch,mongodb missing --lib`
- `cargo test -p helios-persistence --no-default-features --features R4,R4B,R5,R6,sqlite --test search_suite test_missing`
- `cargo test -p helios-rest --test search_integration missing_modifier`
- `cargo test -p helios-persistence --features postgres --test postgres_tests postgres_integration_search_missing_modifier`
- Computer Use against the SQLite UI with the existing dashboard seed

Elasticsearch unit coverage passed. The live test binary compiled, but Docker Desktop blocked before container creation in its credential helper.

### Before

SQLite missing search returning zero results

### After

The captures below show `:missing=true` returning only the patient without `birthDate`, and `:missing=false` returning the 140 seeded patients with the field present.

Closes #632
missing-true
missing-false

### DIFF
--- a/crates/persistence/src/backends/elasticsearch/backend.rs
+++ b/crates/persistence/src/backends/elasticsearch/backend.rs
@@ -637,8 +637,7 @@ impl ElasticsearchBackend {
             SearchParamType::Number => vec!["missing"],
             SearchParamType::Quantity => vec!["missing"],
             SearchParamType::Uri => vec!["contains", "below", "above", "missing"],
-            SearchParamType::Composite => vec!["missing"],
-            SearchParamType::Special => vec![],
+            SearchParamType::Composite | SearchParamType::Special => vec![],
         }
     }
 }
@@ -647,6 +646,15 @@ impl ElasticsearchBackend {
 mod tests {
     use super::*;
     use serde_json::json;
+
+    #[test]
+    fn missing_is_not_advertised_for_composite_or_special_parameters() {
+        assert!(ElasticsearchBackend::modifiers_for_type(SearchParamType::Composite).is_empty());
+        assert!(ElasticsearchBackend::modifiers_for_type(SearchParamType::Special).is_empty());
+        assert!(
+            ElasticsearchBackend::modifiers_for_type(SearchParamType::String).contains(&"missing")
+        );
+    }
 
     #[test]
     fn test_config_defaults() {
--- a/crates/persistence/src/backends/elasticsearch/search/modifier_handlers.rs
+++ b/crates/persistence/src/backends/elasticsearch/search/modifier_handlers.rs
@@ -17,17 +17,22 @@ pub fn build_missing_clause(param: &SearchParameter) -> Option<Value> {
         .map(|v| v.value == "true")
         .unwrap_or(true);
 
-    let path = nested_path_for_type(param.param_type);
-    let name_field = format!("{}.name", path);
-
-    let exists_query = json!({
-        "nested": {
-            "path": path,
-            "query": {
-                "term": { name_field: &param.name }
-            }
+    let exists_query = match param.name.as_str() {
+        "_id" => json!({ "exists": { "field": "resource_id" } }),
+        "_lastUpdated" => json!({ "exists": { "field": "last_updated" } }),
+        _ => {
+            let path = nested_path_for_type(param.param_type);
+            let name_field = format!("{}.name", path);
+            json!({
+                "nested": {
+                    "path": path,
+                    "query": {
+                        "term": { name_field: &param.name }
+                    }
+                }
+            })
         }
-    });
+    };
 
     if is_missing {
         // Resource does NOT have this parameter
--- a/crates/persistence/src/backends/elasticsearch/search/query_builder.rs
+++ b/crates/persistence/src/backends/elasticsearch/search/query_builder.rs
@@ -185,6 +185,13 @@ impl<'a> EsQueryBuilder<'a> {
 
     /// Builds a clause for a single search parameter.
     fn build_parameter_clause(&self, param: &SearchParameter) -> Option<Value> {
+        // Presence is independent of the parameter's ordinary value syntax.
+        // Resolve it before `_id` and `_lastUpdated`, which would otherwise
+        // interpret the boolean literal as an ID or date value.
+        if param.modifier == Some(SearchModifier::Missing) {
+            return modifier_handlers::build_missing_clause(param);
+        }
+
         // Handle special parameters
         match param.name.as_str() {
             "_id" => return self.build_id_clause(param),
@@ -192,11 +199,6 @@ impl<'a> EsQueryBuilder<'a> {
             "_text" => return fts::build_text_clause(param),
             "_content" => return fts::build_content_clause(param),
             _ => {}
-        }
-
-        // Handle :missing modifier
-        if param.modifier == Some(SearchModifier::Missing) {
-            return modifier_handlers::build_missing_clause(param);
         }
 
         // Dispatch based on parameter type
@@ -424,6 +426,29 @@ mod tests {
         let es_query = builder.build(&query);
         let body_str = serde_json::to_string(&es_query.body).unwrap();
         assert!(body_str.contains("resource_id"));
+    }
+
+    #[test]
+    fn missing_precedes_id_and_last_updated_dispatch() {
+        let builder = EsQueryBuilder::new("acme", "Patient", "hfs_acme_patient".to_string());
+
+        for (name, param_type, field) in [
+            ("_id", SearchParamType::Token, "resource_id"),
+            ("_lastUpdated", SearchParamType::Date, "last_updated"),
+        ] {
+            let query = SearchQuery::new("Patient").with_parameter(SearchParameter {
+                name: name.to_string(),
+                param_type,
+                modifier: Some(SearchModifier::Missing),
+                values: vec![SearchValue::eq("false")],
+                chain: vec![],
+                components: vec![],
+            });
+            let es_query = builder.build(&query);
+            let clause = &es_query.body["query"]["bool"]["must"][0];
+
+            assert_eq!(clause["exists"]["field"], field);
+        }
     }
 
     fn not_param(values: Vec<SearchValue>) -> SearchQuery {
--- a/crates/persistence/src/backends/mongodb/search_impl.rs
+++ b/crates/persistence/src/backends/mongodb/search_impl.rs
@@ -677,7 +677,8 @@ impl MongoBackend {
         for param in &query.parameters {
             if matches!(
                 param.modifier,
-                Some(SearchModifier::Above)
+                Some(SearchModifier::Missing)
+                    | Some(SearchModifier::Above)
                     | Some(SearchModifier::Below)
                     | Some(SearchModifier::In)
                     | Some(SearchModifier::NotIn)
@@ -1784,5 +1785,33 @@ mod date_filter_tests {
     #[test]
     fn invalid_dates_error() {
         assert!(build_date_filter_doc(&SearchValue::parse("not-a-date"), "value_date").is_err());
+    }
+}
+
+#[cfg(test)]
+mod query_support_tests {
+    use super::*;
+    use crate::backends::mongodb::MongoBackendConfig;
+
+    #[test]
+    fn missing_is_rejected_before_type_specific_value_parsing() {
+        let backend = MongoBackend::new(MongoBackendConfig::default()).unwrap();
+        let query = SearchQuery::new("Patient").with_parameter(SearchParameter {
+            name: "birthdate".to_string(),
+            param_type: SearchParamType::Date,
+            modifier: Some(SearchModifier::Missing),
+            values: vec![SearchValue::eq("true")],
+            chain: vec![],
+            components: vec![],
+        });
+
+        let error = backend.validate_query_support(&query).unwrap_err();
+        assert!(matches!(
+            error,
+            StorageError::Search(SearchError::UnsupportedModifier {
+                ref modifier,
+                ..
+            }) if modifier == "missing"
+        ));
     }
 }
--- a/crates/persistence/src/backends/postgres/backend.rs
+++ b/crates/persistence/src/backends/postgres/backend.rs
@@ -958,8 +958,7 @@ impl PostgresBackend {
             SearchParamType::Number => vec!["missing"],
             SearchParamType::Quantity => vec!["missing"],
             SearchParamType::Uri => vec!["contains", "below", "above", "missing"],
-            SearchParamType::Composite => vec!["missing"],
-            SearchParamType::Special => vec![],
+            SearchParamType::Composite | SearchParamType::Special => vec![],
         }
     }
 }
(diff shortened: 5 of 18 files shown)


Gordex2014:
Closed the Codecov patch coverage gap in commit 3631ff3dc.

The uncovered line was the empty-values guard in the SQLite query builder. I added the unit test missing_parameter_without_values_is_ignored, which verifies that a :missing parameter with no values returns no backend condition without panicking.

Validation:
- Targeted unit test passed
- helios-persistence library tests: 800 passed, 0 failed
- cargo fmt --all -- --check passed
- git diff --check passed

smunini (approved):

PR #659, opened 2026-08-24
fix(ui): the resource-type rail unions the compartment set with the generated Resource enum

### LINKED ISSUE #648: ViewDefinition is missing from the UI resource-type rail, so the Resource Editor cannot reach it
## Summary

`ViewDefinition` never appears in the resource-type rail on the Resources page (nor on Search, Queries, or Bulk Export), so there is no way to reach the Resource Editor for a ViewDefinition through the UI. The type is a first-class resource everywhere else in HFS: it is a variant of the generated `Resource` enum, it is advertised in `/metadata`, the REST write path accepts it, and the validator's schema packs contain its StructureDefinition.

## Current behavior

1. Open `/ui/resources`.
2. Scan the resource-type rail (or use its filter) for `ViewDefinition`.
3. It is absent. The same rail is used on `/ui/search`, `/ui/queries`, and `/ui/bulk-export`, so ViewDefinition is missing from all of them.

The editor itself is reachable only by hand-writing the URL: `/ui/editor?resource_type=ViewDefinition` renders (the schema pack has the type), but nothing in the UI links there.

## Expected behavior

`ViewDefinition` appears in the type rail alongside the other supported types, and clicking through to the Resource Editor lets a user create, edit, and save one.

## Evidence

The UI type list does **not** come from `/metadata`. It comes from the first `CompartmentDefinition`'s `resource[]` array:

- `crates/ui/src/compartments.rs:134` — `CompartmentCatalog::resource_type_names()` returns `definitions().first().map(|def| def.resource.iter().map(|r| r.code.clone()))`. Its doc comment states this is "Used by the resource pickers on the Search, Queries, Resources, and Bulk Export pages."
- `crates/ui/src/lib.rs:1332` (Resources), `:1252` (Search), `:1291` (Queries), `crates/ui/src/bulk_export.rs:265` — every caller feeds that list into `build_rail_entries`.
- `data/compartment-definitions-r4.json` — the `Device` compartment (the first entry) enumerates 145 resource types. `ViewDefinition` is not one of them, and the string does not occur anywhere in the file. Same for the R4B/R5/R6 files.

CompartmentDefinition is a spec artifact, so ViewDefinition — which is defined by the SQL-on-FHIR IG, not the FHIR core spec — will never be listed there.

Meanwhile, everywhere the rest of the server enumerates resource types, ViewDefinition *is* present:

- `crates/fhir/src/r4.rs:167761` — `ViewDefinition(Box<ViewDefinition>)` is a `Resource` enum variant (fhir-gen builds this from the IG's StructureDefinition; see `crates/fhir-gen/src/lib.rs:3278`).
- `crates/fhir-macro/src/lib.rs:3281` — the derived `get_resource_type_names()` emits a literal for every enum variant, so ViewDefinition is included.
- `crates/rest/src/fhir_types.rs:165` — `get_resource_type_names_for_version()` returns that generated list.
- `crates/rest/src/handlers/capabilities.rs:137-181` — `build_capability_statement` maps that list straight into `rest.resource[]` with no filtering, so `/metadata` **does** advertise ViewDefinition. `is_valid_resource_type()` (`crates/rest/src/fhir_types.rs:110`) uses the same list, so the REST create/update path accepts it.
- The validator's schema pack has the type: `zcat crates/fhir-validator/packs/fhir_schemas_r4.json.gz | grep -c ViewDefinition` is non-zero, and `crates/ui/src/editor.rs:214` resolves the editor tree from `packs::core_registry(version)`.

So the gap is isolated to the UI's choice of source for the rail: CompartmentDefinition rather than CapabilityStatement.

## Suggested approach

The rail's source is the thing to change. `CompartmentDefinition.resource[]` is the wrong list to enumerate "types this server supports" — it happens to be a near-complete list of core resource types, which is why it has worked until now, but it is definitionally limited to spec types and excludes anything HFS adds (today ViewDefinition; tomorrow any other IG-defined or R6 "additional" resource).

Switch `CompartmentCatalog::resource_type_names()` — or introduce a sibling method and repoint the four rail call sites at it — to read `CapabilityStatement.rest[0].resource[].type` from the tenant's `/metadata`, which already returns the correct set. The catalog already fetches conformance over HTTP and caches per (tenant, version) with the same invalidation semantics, so the plumbing is in place; keep the "empty response is not cached" guard at `crates/ui/src/compartments.rs:100-116`.

Worth confirming while in there:

- Whether `/ui/resources?type=ViewDefinition` → **Create new** → **Save** round-trips through the REST API once the type is selectable, and whether the Guided form renders a usable tree for it.
- Whether ViewDefinition has search parameters registered; the rail entries carry per-type counts sourced from the observability dashboard snapshot, which should degrade gracefully to zero.
- Whether `CompartmentCatalog::resource_type_names()` still has non-rail callers that genuinely want the compartment membership list.

## Scope / out of scope

In scope: the Resources / Search / Queries / Bulk Export type rail, and reaching the Resource Editor for ViewDefinition.

Out of scope: SQL-on-FHIR ViewDefinition *execution* (`$run`, sof-server, runners) — this is only about authoring the resource in the UI. Also out of scope: the compartment pages themselves, which legitimately enumerate compartment membership.

Related: #636 covers the opposite direction — an *unrecognized* type in `?type=` being accepted as a create target. A fix here that grounds the rail in CapabilityStatement likely gives #636 its validation source too.

angela-helios:
Closes #648.

The rail enumerated the first CompartmentDefinition's resource list — the spec's set — so types HFS itself serves but the spec's compartments never mention (`ViewDefinition`) were unreachable from Resources, Search, Queries, and Bulk Export alike.

`resource_type_names` now unions that enumeration with the generated `Resource` enum's names (`FhirResourceTypeProvider`, dispatched on the sidebar's FHIR version, same cfg pattern as `rest/fhir_types.rs`). Three consequences:

- `ViewDefinition` appears in the rail everywhere the rail appears, and the editor is reachable for it.
- The list is alphabetical now (it was rough spec order).
- A degraded compartment fetch leaves the full enum list instead of an empty rail — strictly better than before.

Pinned in `router_http` (`ViewDefinition` and `EffectEvidenceSynthesis` both present — the union keeps both sources) and the e2e rail spec. Full `cargo test -p helios-ui` + the rail Playwright spec green.


angela-helios:
`Test Rust` here is failing on #671, not on this PR: both attempts landed on the Windows agent `github-agent4`, where the CRLF checkout breaks every multi-line assertion of rendered HTML (`tests::queries_page_renders_shell_and_marks_nav_current` is a main test this branch does not touch). #672 pins the embedded UI content to LF and fixes it deterministically — verified locally on Windows with the full `helios-ui` suite green. Once #672 merges, a re-run of this job goes green on any agent. Everything else on this PR is already green, including the multi-version job.

smunini (approved):

PR #664, opened 2026-08-24
fix(ui): dashboard card names the effective FHIR version, worded "used for"

### LINKED ISSUE #553: Home dashboard: reword "enabled for" to "used for", and cover multiple tenants and FHIR versions
## 1. Copy change

The Home page "Resource types" card reads **"enabled for R4"**. Change it to **"used for R4"**.

`locales/en/main.ftl:137`:

```
card-resource-types-sub = enabled for { $version }
```

The number above it is `snapshot.distinct_types` — how many distinct resource types actually have data stored — so "used" is the accurate word; "enabled" suggests a configuration state the figure has nothing to do with.

The other two catalogs need the equivalent change:

- `locales/es/main.ftl:133` — `habilitados para { $version }`
- `locales/de/main.ftl:133` — `aktiviert für { $version }`

## 2. Test the dashboard across tenants and FHIR versions

Neither dimension is covered today — `crates/ui/e2e/tests/dashboard.spec.ts` has three tests (stat cards render, window selector, legend) and none touch tenants or versions. While scoping the tests, two things turned up that should be checked as part of this work:

### a. The version on this card does not follow the version selector — likely a real bug

`crates/rest/src/dashboard.rs:194` fixes the snapshot's version at provider construction:

```rust
fhir_version: config.default_fhir_version.to_string(),
```

That's the **server's configured default**, not the user's selected FHIR version. Meanwhile the same page's status/footer is built from `RequestVersion` (`build_index_page` → `current_status(version, rv.0, tenant)`, `crates/ui/src/lib.rs:1284-1296`), which *does* follow the user's stored choice.

So switching FHIR version in the UI selector should leave this card's `{ $version }` unchanged while the footer changes — two different versions displayed on one page. Confirm in a browser, and decide which one the card should reflect.

Compounding it: `helios_observability::dashboard::snapshot` keys its cache on `(window, tenant)` only (`crates/observability/src/dashboard.rs:228`) — there is no version component, so a per-version snapshot could not be cached correctly even if the provider produced one.

Also worth settling as part of the copy change: `distinct_types` counts stored resource types irrespective of FHIR version, so pinning any version label to it may be misleading regardless of which version is shown.

### b. Multi-tenant looks correct but is unverified, and the docs are stale

The snapshot **is** tenant-scoped — `dashboard::snapshot(window, &tenant.id)` (`crates/ui/src/lib.rs:1294`), cache keyed on `(window, tenant)`. So the behavior appears right; it just has no test.

The doc comment on `DashboardMetrics` (`crates/ui/src/lib.rs:325-330`) still says these figures are "derived from the live `DashboardSnapshot` (**default tenant**)". That is no longer true and should be corrected so the next reader doesn't trust it.

## Acceptance criteria

- [ ] The card reads "used for { $version }" in `en`, with matching updates in `es` and `de`.
- [ ] e2e coverage: the dashboard's counts and chart reflect the **selected tenant**, and switching tenants changes them.
- [ ] e2e coverage: the dashboard behaves correctly across FHIR versions, including that the version shown on the card is the intended one.
- [ ] A decision recorded on whether the card should show the server's default version or the user's selected version — and the code matching it (including the snapshot cache key, if the answer is per-version).
- [ ] Tenant isolation verified: one tenant's counts never leak into another's.
- [ ] The stale "(default tenant)" doc comment at `crates/ui/src/lib.rs:325-330` corrected.

## Pointers

- `locales/{en,es,de}/main.ftl:136-137 / 132-133` — the strings
- `crates/ui/templates/pages/index.html` — the stat card
- `crates/ui/src/lib.rs:1284-1307` — `build_index_page`; `:325-340` — `DashboardMetrics` and the stale comment
- `crates/observability/src/dashboard.rs:205-250` — `snapshot` and its `(window, tenant)` cache key
- `crates/rest/src/dashboard.rs:182-260` — the provider that fixes `fhir_version`
- `crates/ui/e2e/tests/dashboard.spec.ts` — where the new tests go

dougc95:
Closes #553.

## What changed

**The bug (issue section 2a) was real, and is fixed at the source.** The "Resource types" card rendered `DashboardSnapshot.fhir_version`, which `StorageDashboardProvider::new` pins to `config.default_fhir_version` once at server startup — while the sidebar selector two labels away renders the per-request `RequestVersion`. After a sidebar switch the two disagreed on one page. The card now renders `status.fhir_version_label()` — the exact same per-request `Status` source the sidebar uses — so card/sidebar disagreement is structurally impossible, and the redundant `DashboardMetrics.fhir_version` field is deleted (Askama's compile-time template checking guarantees no consumer was missed). This also fixes the sample-data path, which hardcoded "R4".

**Copy change:** `card-resource-types-sub` reads "used for { $version }" (en), "usados para" (es), "verwendet für" (de) — the figure counts stored resource types, not a configuration state.

**Decision recorded (acceptance criterion 4):** the card shows the **user's selected version**, not the server default. Rationale: the type catalog, sidebar, and the existing "picker follows the requested version" e2e contract already treat the selection as authoritative. The counts remain cross-version (`count_all_types` has no version filter in any backend); version-scoping them would need a version parameter through `DashboardProvider::snapshot`, the snapshot cache key, and version-filtered count queries in all four backends — noted as a possible follow-up, deliberately not done here. The snapshot cache key needs **no** change: nothing version-dependent is rendered from the cached payload anymore.

**Stale docs:** the "(default tenant)" comment on `DashboardMetrics` plus five more stale sites found while verifying (module docs in `helios-rest`/`helios-observability` dashboard modules, the sidebar-selector comment in `base.html`, `fhir_version_label`'s doc, the chart comment in `index.html`).

## Tests

- **Inner ring (new):** `dashboard_multiversion_http.rs` — the card reads "used for R4" on the default build, and follows a `POST /ui/version` switch to R4B (R4B-gated). Both watched failing before the fix.
- **e2e version (new):** `dashboard-multiversion.spec.ts` asserts the card follows a sidebar switch and not the boot-time default.
- **e2e tenants (new):** `dashboard-tenants.spec.ts` — counts, picker, and chart follow the sidebar-selected tenant; a tenant's markers never appear in the other's dashboard, in either direction. Uses suite-unique marker types (`Flag`/`Basic`), a per-run unique tenant id, and new `AppChrome` tenant-picker helpers mirroring the version ones; `api.ts` seeding gains an `X-Tenant-ID` parameter.
- **Pre-existing test repaired:** the multiversion type-set test read its "R4 ground truth" from whatever version the server seeds — under CI's `HFS_DEFAULT_FHIR_VERSION=R6` pin its final R4 assertion demands an R6-only type of the R4 picker, exposed when the moving R6 spec build gained `DeviceAlert`. It now pins R4 explicitly before reading the baseline.
- `addTenant`'s settle ceiling raised 150s → 300s (NTFS dev machines exceed 150s; the wait is event-driven, fast disks pay nothing).

Verified locally: `helios-ui` suite green on the default R4 build (179 tests) and the R4B leg; all 10 existing dashboard browser tests green; the tenant spec green; the multiversion leg 4/4 against a CI-identical `R4,R4B,R5,R6` build pinned to R6.

### DIFF
--- a/crates/observability/src/dashboard.rs
+++ b/crates/observability/src/dashboard.rs
@@ -13,11 +13,11 @@
 //!
 //! ## Scope and trust
 //!
-//! The registered provider reports counts for a **single tenant** (the server's
-//! default tenant) and is consumed only by the operator dashboard. Per-tenant
-//! counts are deliberately never exported to the public Prometheus `/metrics`
-//! endpoint (see [`crate::metrics`]); this snapshot is a separate, operator-facing
-//! surface.
+//! Each snapshot reports counts for the single tenant passed to [`snapshot`]
+//! (#344) and is consumed only by the operator dashboard. Per-tenant counts
+//! are deliberately never exported to the public Prometheus `/metrics`
+//! endpoint (see [`crate::metrics`]); this snapshot is a separate,
+//! operator-facing surface.
 //!
 //! ## Time resolution
 //!
@@ -154,7 +154,9 @@ pub struct ExportJobCounts {
 /// FHIR types — so this crate stays dependency-light.
 #[derive(Clone, Debug, Default)]
 pub struct DashboardSnapshot {
-    /// Default FHIR version the server serves (e.g. `"R4"`).
+    /// Default FHIR version the server serves (e.g. `"R4"`), fixed when the
+    /// provider is constructed. The UI's dashboard card does not render this:
+    /// it shows the request's effective version instead (#553).
     pub fhir_version: String,
     /// Total non-deleted resources across all types for the default tenant.
     pub total_resources: u64,
--- a/crates/rest/src/dashboard.rs
+++ b/crates/rest/src/dashboard.rs
@@ -7,9 +7,9 @@
 //! and registers it in [`crate::build_app`]; the UI reads the resulting snapshot
 //! through the storage-agnostic `helios-observability` registry.
 //!
-//! The single-tenant scope matches the operator dashboard: figures reflect the
-//! server's **default tenant** only, and are never exported to the public
-//! Prometheus `/metrics` endpoint (per the design in
+//! Figures are scoped per snapshot call to the requesting tenant (#344; the
+//! server's default tenant is only the empty-id fallback), and are never
+//! exported to the public Prometheus `/metrics` endpoint (per the design in
 //! [`helios_observability::metrics`]). The same [`resource_count_series`] helper
 //! also backs the authenticated `/console/metrics/resource-counts` JSON handler,
 //! so the cumulative-bucketing semantics live in exactly one place.
--- a/crates/ui/e2e/pages/api.ts
+++ b/crates/ui/e2e/pages/api.ts
@@ -6,14 +6,20 @@ import type { APIRequestContext } from "@playwright/test";
 
 const FHIR_JSON = "application/fhir+json";
 
-/** POST a resource; returns the server-assigned id. */
+/** POST a resource; returns the server-assigned id. `tenant` scopes the write
+ * via `X-Tenant-ID` (#553 — the default-tenant routes need no header). */
 export async function createResource(
   request: APIRequestContext,
   type: string,
   body: Record<string, unknown>,
+  tenant?: string,
 ): Promise<string> {
   const res = await request.post(`/${type}`, {
-    headers: { "Content-Type": FHIR_JSON, Accept: FHIR_JSON },
+    headers: {
+      "Content-Type": FHIR_JSON,
+      Accept: FHIR_JSON,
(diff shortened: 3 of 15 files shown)


dougc95:
Screenshots from a local multi-version build (`R4,R4B,R5,R6`, booted with `HFS_DEFAULT_FHIR_VERSION=R6`), regions of interest highlighted — red is the fixed card sub-line, blue is the sidebar version selector it now shares a source with.

**Before the sidebar switch** — the card reads "used for R6" and the selector agrees ("FHIR R6"):

553-card-R6
553-sidebar-R6

**After switching the sidebar to R4** — the card follows the selection. Before this PR it stayed pinned at the boot-time default:

553-card-R4
553-sidebar-R4

smunini (approved):

Issue #652, opened 2026-08-24
UI copy: no capitalization convention — headings and buttons mix Title Case and sentence case


smunini:
## Summary

UI labels have no capitalization convention. Headings, buttons, and card titles are Title Case in some places and sentence case in others, sometimes for the same kind of control on adjacent pages. Pick a convention, write it down, and sweep the English catalog.

## Current behavior

Named examples, all in `locales/en/main.ftl`:

| Line | Key | Current | Should be |
|---|---|---|---|
| 285 | `cmp-new` | New compartment definition | New Compartment Definition |
| 554 | `bulk-import-new` | New submission | New Submission |
| 121 | `tenants-add` | Add tenant | Add Tenant |
| 281 | `sp-new` | New search parameter | New Search Parameter |
| 58 | `terminology-heading` | Terminology server | Terminology Server |

These are not isolated. Across the catalog's headings, titles, labels, and action buttons, roughly 18 strings are already Title Case and roughly 62 are sentence case:

**Already Title Case** — `tenants-title = Tenant Maintenance` (111), `sp-heading = Search Parameters` (234), `bulk-import-title = Bulk Import` (553), `bulk-import-create-title = Create Bulk Submission` (555), `bulk-import-add-manifest-title = Add Manifest` (597), `bulk-export-active-title = Active Exports` (669), `nav-resource-editor = Resource Editor` (96), `nav-saved-queries = Saved Queries` (212).

**Sentence case** — `dashboard-heading = Server dashboard` (44), `editor-heading = Resource editor` (458), `queries-heading = Saved queries` (214), `resources-types-heading = Resource types` (517), `card-export-jobs = Export jobs` (155), `card-import-jobs = Import jobs` (157), `tenants-delete = Delete tenant` (134), `chart-pick-heading = Charted resource types` (165).

The same concept is spelled both ways depending on where it appears: `nav-resource-editor = Resource Editor` (96) against `editor-heading = Resource editor` (458), and `nav-saved-queries = Saved Queries` (212) against `queries-heading = Saved queries` (214).

## Expected behavior

A stated convention, applied consistently. The examples above point at Title Case for headings, page titles, card titles, nav entries, and action buttons.

## Evidence

[...]

Issue #649, opened 2026-08-24
Promote SQL on FHIR to a top-level nav section with View Definitions, SQL Queries, SQL Views, Export, and Files


smunini:
## Motivation

`SQL-on-FHIR` is a single dead placeholder in the sidebar — a `nav-item--soon` span with no href — even though the SoF operations are fully routed and serving in `helios-rest`. There is no UI surface for authoring a ViewDefinition, running a query, or retrieving an export.

Brett has designed the first ViewDefinition Editor frame, and the menu needs to grow from one entry into a section with its own children to host it.

Design source: [Figma `CcLtq79cH2aHv4Ii9aNQTP`, node `420-2`](https://www.figma.com/design/CcLtq79cH2aHv4Ii9aNQTP/Untitled?node-id=420-2&t=ug0KKdxWzP4f6aoj-1). Same file the rest of the UI is inspected from — see `crates/ui/README.md:198-215` for the export conventions (icons via the Figma REST API, fills normalized to `currentColor`).

## Proposed behavior

Promote **SQL on FHIR** to its own top-level menu item — a sidebar section of its own, at the same level as Work, Batch & Data, and Server — rather than a leaf inside Batch & Data where it sits today. It gets five children:

```
SQL on FHIR
├── View Definitions
├── SQL Queries
├── SQL Views
├── Export
└── Files
```

Each entry gets a real `/ui/...` route. The ViewDefinition Editor behind **View Definitions** follows the Figma frame.

## Current state

- `crates/ui/templates/layouts/base.html:95-98` — the entire SQL-on-FHIR presence today:

  ```html
  <span class="nav-item nav-item--soon" title="{{ i18n.t("nav-sql-on-fhir") }}">
    <span class="icon">{% include "icons/grid.svg" %}</span>
    <span class="nav-item__label">{{ i18n.t("nav-sql-on-fhir") }}</span>
  </span>
  ```

  It sits *inside* the `nav-section-batch-data` group, between Export and the Server section. It needs to move out and become a peer of that group, not a child of it.

- `locales/en/main.ftl:102` — `nav-sql-on-fhir = SQL-on-FHIR`. Note the current label hyphenates; the requested structure spells it "SQL on FHIR". Pick one and apply it consistently across `locales/{en,es,de}/main.ftl`.

- `crates/ui/src/lib.rs:833-937` — the UI router. No `/ui/sql-*`, `/ui/view-definitions`, or equivalent route exists.

[...]

Issue #567, opened 2026-08-18
SoF 3.0.0-ballot: implement SQLView and transitive dependency-graph resolution


smunini:
SQL on FHIR 3.0.0-ballot introduces `SQLView`, a second profile on `Library` alongside `SQLQuery`. A SQLView is a reusable, named SQL query whose *result serves as a table for other queries*, so transformations compose rather than being duplicated.

We only handle `sql-query`. `crates/rest/src/handlers/sof/subject.rs` already classifies a Library as `SubjectKind::SqlView` when `Library.type` carries `LibraryTypesCodes#sql-view`, but nothing downstream treats it differently: it falls through to the SQLQuery execution path.

## What the profile requires

`SQLView` is `SQLQuery` minus parameters, plus one consequence that is not cosmetic:

- `parameter 0..0` — views cannot be parameterized. A SQLView subject supplying `parameters` must be rejected.
- `relatedArtifact.resource only Canonical(ViewDefinition or SQLView)` — **the dependency graph becomes transitive**. A SQLQuery may depend on a SQLView, which depends on another SQLView, which finally depends on a ViewDefinition. The spec is explicit that "the graph is therefore transitive, and its leaves are always ViewDefinitions."
- Same `sql-must-be-sql-expressions` invariant, same `content` / `sql-text` shape, same `relatedArtifact.label` naming rule.

## Where the work lands

`crates/rest/src/handlers/sof/sqlquery.rs::run_library_subject` resolves `library.depends_on` one level deep and materializes each dependency as a ViewDefinition:

```rust
for dep in &library.depends_on {
    let vd_json = /* inline context, else resolve_canonical_view_definition */;
    let schema = TableSchema::from_view_definition(&vd_json);
    engine.create_table(&label, &schema)?;
    // ... runner.run_view(...) -> insert_rows
}
```

That needs to become a recursive walk: for each dependency, resolve it, and if it is a SQLView, resolve *its* dependencies first, materialize those, run its SQL, and register the result as a table under this level's label. `crates/rest/src/handlers/sof/export.rs::prepare_named_sqlquery` has the same one-level assumption for the export path.

Two things fall out of the recursion that need deciding:

- **Cycle detection.** A SQLView graph can be cyclic; today's flat loop cannot loop forever, a recursive one can. Needs a visited-set keyed by canonical URL, and a clear error when a cycle is found.
[...]

PR #673, opened 2026-08-25
feat(ui): SQL on FHIR as a top-level nav section with five routed children

angela-helios:
Part 1 of #649 — the nav structure, as its own reviewable change per the issue's suggested approach. The ViewDefinition Editor (Figma `420-2`) and the pages over `$sql-run` / `$sql-export` follow in separate PRs.

## What

- **SQL on FHIR is a top-level sidebar section** — a peer of Work, Batch & Data, and Server — replacing the dead `nav-item--soon` placeholder that sat inside Batch & Data. Children: View Definitions, SQL Queries, SQL Views, SQL Export, Files.
- **Every child navigates**: five `/ui/sql/*` routes render a shared stub page (`sql-stub.html`) stating what the page will host and which already-routed server operation serves that data (`$sql-run`, `$sql-export`, the export manifest/file routes). Nothing in the menu is dead.
- **Labels** exist in en/es/de; en follows the Title Case convention (#652), es/de keep their norms. The spelling is "SQL on FHIR" (unhyphenated), per the issue.
- Existing flat `nav__section` pattern — no new CSS, and the section collapses to a divider in the icon-only rail like the other three.

## Decisions for review

- **"SQL Export", not "Export"** — the sidebar already carries Batch & Data's FHIR Bulk Data Export, a different operation; the issue flagged the ambiguity. Also matches the sibling names SQL Queries / SQL Views.
- **SQL Views is a stub route, not `--soon`** — the stub states the surface honestly, and the backend does serve SQLView Library subjects today (transitive resolution pending #567). Easy to flip to `--soon` if preferred.
- **Icons** are reused from the vendored set (grid, code, layers-platforms, export, copy) — the Figma frame shows the sidebar collapsed, so no SQL-specific icons exist yet; the editor PR can bring Brett's if they appear.
- The nojs guard's "coming-soon entries are inert" test lost its premise (no placeholder remains) — inverted to assert **no** dead placeholder exists.

## Tests

- `router_http`: new test — section present with all five links, no `nav-item--soon` anywhere, each route 200s and marks its nav entry `aria-current`.
- The five routes are in `e2e/pages/routes.ts`, so every cross-page guard covers them. Run locally: **a11y (axe, light+dark), design-system, no-cdn: 51 passed; nojs: 11 passed**; `cargo test -p helios-ui` fully green.

### DIFF
--- a/crates/ui/e2e/pages/chrome.ts
+++ b/crates/ui/e2e/pages/chrome.ts
@@ -18,11 +18,6 @@ export class AppChrome {
     return this.page.locator(`a.nav-item[href='${href}']`);
   }
 
-  /** A disabled "coming soon" nav entry by its visible label. */
-  soonItem(label: string): Locator {
-    return this.page.locator("span.nav-item--soon", { hasText: label });
-  }
-
   themeButton(theme: Theme): Locator {
     return this.page.locator(`[data-set-theme='${theme}']`);
   }
--- a/crates/ui/e2e/pages/routes.ts
+++ b/crates/ui/e2e/pages/routes.ts
@@ -22,6 +22,11 @@ export const ROUTES = [
   "/ui/bulk-export",
   "/ui/bulk-export/active",
   "/ui/subscriptions",
+  "/ui/sql/view-definitions",
+  "/ui/sql/queries",
+  "/ui/sql/views",
+  "/ui/sql/export",
+  "/ui/sql/files",
 ];
 
 // The bulk-import detail page only exists with a submission behind it. Seed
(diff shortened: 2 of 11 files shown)


angela-helios:
The browser-suite red was real — `chrome.spec.ts` still asserted the old coming-soon placeholder; updated in 2aab722 (new test covers the section's five links) and verified locally.

The `Test Rust` red is #671, not this PR: the job landed on the Windows agent again and failed `resources_deep_links_focus_the_selected_type`, a main test this branch does not touch. It goes green once #659 (which carries the `.gitattributes` fix) merges, or when the job lands on a Linux agent.

angela-helios:
`Test Rust` red again on the Windows agent, *after* #659 merged: the job checks out with `clean: false` and checkout's `--force` never rewrites files whose blob didn't change, so the agent's workspace still holds the pre-`.gitattributes` CRLF templates. #693 adds a `git checkout-index --all --force` renormalize step to the five ci.yml jobs; once it merges, a re-run here goes green on any agent. Re-running now anyway in case the scheduler picks a Linux agent.

smunini (approved):

PR #706, opened 2026-08-25
fix(ui): reset incompatible query comparators

### LINKED ISSUE #628: Query builder keeps date comparator after changing birthdate to name
## Summary

Changing a Patient condition from the date parameter `birthdate` to the string parameter `name` leaves the `ge` comparator active.

The builder serializes the updated condition without reporting an error, while the narration still describes a date comparison.

Related QA campaign: #556

## Environment

- HFS `0.2.1`
- Revision `3b46edc07b343ff9ea97a694e80147439e89e0ad`
- FHIR R4
- SQLite
- Chrome

## Steps to reproduce

1. Open the query builder.
2. Select Patient.
3. Enter:

   ```text
   GET /Patient?birthdate=ge1980-01-01
   ```

4. Confirm that the builder displays `birthdate`, the `ge` comparator, and `1980-01-01`.
5. Change only the parameter from `birthdate` to `name`.
6. Inspect the comparator, serialized query, narration, and validation state.

## Actual result

The `ge` comparator remains selected after the parameter changes to `name`.

The builder serializes:

```text
GET /Patient?name=ge1980-01-01
```

The narration states that the Patient name is on or after `1980-01-01`. No warning marks the comparator as incompatible, and the Run action remains available.

Changing the parameter does not execute the query automatically.

Query builder showing the retained ge comparator after changing birthdate to name

## Expected result

When the selected parameter does not support the current comparator, the builder must either:

- Clear the comparator and serialize the value as a string condition.
- Mark the condition invalid and block execution.
- Require the user to choose a compatible option.

The builder must not retain date-specific state under `name` or describe a string condition as a date comparison.

## Impact

The serialized query and narration no longer describe the same search semantics. A user can run a query that appears to compare a date but is submitted under a string parameter.

The lack of validation makes the stale comparator look intentional.

## Acceptance criteria

- Changing from `birthdate` to `name` does not leave `ge` active.
- Comparator controls only offer options supported by the selected SearchParameter type.
- The serialized query, narration, and validation state agree.
- The Run action cannot submit an incompatible parameter and comparator combination.
- Changing the parameter does not execute the query automatically.
- A browser regression test covers changing a date parameter with a comparator to a string parameter.

Gordex2014:
## Summary

- reset incompatible comparator controls when a query condition changes parameter type
- preserve prefix-looking values as literals when hydrating known non-ordered SearchParameters
- keep comparator normalization transactional while SearchParameter metadata is loading
- apply the same behavior to direct conditions, forward chains, `_has`, and OR alternatives
- preserve comparator support for date, number, and quantity parameters
- update SQL ViewDefinition router tests for the `public_base_url` argument introduced on `main`
- treat extension-only FHIR primitives as empty for scalar operations, restoring the inherited R5 FHIRPath conformance gate

## Testing

- `node --check crates/ui/assets/saved-queries.js`
- `cargo test -p helios-ui --locked --test router_http` (59 passed)
- `cargo test -p helios-ui --locked`
- `cargo clippy -p helios-fhirpath --all-targets --all-features` with the CI warning policy
- `cargo test -p helios-fhirpath --locked --all-features`
- `cargo build -p helios-hfs --features ui,subscriptions --locked`
- full Chromium `queries.spec.ts` suite: 62 passed
- manual browser QA for the reported birthdate-to-name transition, direct string hydration, and explicit Run

Closes #628


Gordex2014:
## Manual QA evidence

### 1. Before: changing birthdate to name left the stale date comparator visible

01-before-stale-comparator

### 2. After: the incompatible comparator is removed and the value is preserved

02-after-normalized-comparator

### 3. Literal safety: a direct string value beginning with `ge` stays literal

03-after-string-literal

### 4. Explicit Run: the normalized query executes successfully and renders 0 results

04-after-explicit-run

smunini (approved):

PR #710, opened 2026-08-26
helios-sof: Arrow IPC stream output for live SQL-on-FHIR query results

### LINKED ISSUE #703: helios-sof: Arrow IPC output for live SQL-on-FHIR query results
> **Rescoped 2026-08-25** after review discussion of the draft HFS Analytics requirements. Originally filed as a pysof-scoped Arrow feature; the surface is now native-first in helios-sof, with the Python-side piece demoted to an optional follow-up. Original motivation thread: #650.

## What Arrow is here — and what it is not

This is **not** a new file format competing with Parquet. For exported artifacts (object storage, bulk files), Parquet remains the interchange — DuckDB, Postgres, and every analysis engine read it well, and helios-sof already emits it.

Arrow IPC is the **no-reparse handoff for results that never become files**: a streamable wire encoding of the Arrow in-memory columnar format. The draft HFS Analytics requirements route *live* query results through the HFS API (analysis kernels never touch storage directly for live queries). Today those results arrive as CSV/JSON that the kernel must parse row by row. As an Arrow IPC stream (`application/vnd.apache.arrow.stream`), the same results are ingested directly — zero parsing — by every consumer under consideration: DuckDB (`FROM read_arrow(...)` / its IPC reader), pandas, polars, and pyarrow. It deliberately serves both sides of the open "DuckDB-in-kernel vs pandas" question.

The decisive implementation fact: **helios-sof already builds Arrow RecordBatches internally** — `crates/sof/src/parquet_schema.rs` exposes the schema inference and array conversion as public, tested functions, and the Parquet path (`crates/sof/src/lib.rs:4098`) assembles RecordBatches and then immediately encodes them away. This feature exposes the batches one step earlier; it is a door, not an engine.

## Scope

1. **Batch-builder refactor (S).** Public `run_view_definition_record_batches()` (+ chunked sibling) wrapping the existing pipeline and `parquet_schema` functions; the Parquet writer is refactored to consume it. No behavior change — proven by existing Parquet tests.
2. **`ContentType::ArrowIpc` (S-M).** New content type emitting the Arrow IPC stream format, wired through the existing surfaces: `sof-cli`, `sof-server`, and the REST `$sql-run` / `$sqlquery-run` handlers (MIME: `application/vnd.apache.arrow.stream`, following the existing Accept-header patterns). The `arrow` crate already in the tree provides the IPC writer.
3. **Schema stance:** inherit today's Parquet type mapping exactly (bool→Boolean, integer→Int32, decimal→Float64, else Utf8, multi-valued→List) so the columnar outputs always agree. Known caveats carried over deliberately (values-inferred schema; decimal precision) — fixing those is a separate issue affecting Parquet and Arrow together.

## Out of scope

- **pysof zero-copy surface** (Arrow PyCapsule / `__arrow_c_stream__`): still useful for pysof's own users (NDJSON-on-disk transformation, local analysts, agent workflows) but not on the critical path of the analytics architecture — split into a follow-up issue if/when wanted. pysof can trivially consume the IPC bytes meanwhile.
- Declared-type-driven schemas (Decimal128, Date32/Timestamp)
- Streaming Parquet

## Test plan

- Red-first: IPC stream round-trips via `pyarrow.ipc.open_stream` / DuckDB; value equivalence vs JSON output for the same input; nulls and List columns; chunked emission for large results (bounded memory); content negotiation on the REST handlers; existing error types on invalid input.
- Invariance: existing Parquet tests pass unchanged after the refactor.
- Benchmark: end-to-end *CSV response + Python parse* vs *Arrow IPC response + `open_stream`* on a large live-query result.

Refs: #650 (research lineage), #688 (perf work that preceded this), draft HFS Analytics requirements (req 2c/3: live results flow through the HFS API).

dougc95:
Closes #703 (as rescoped after the HFS Analytics requirements review).

## What this is — and is not

**Not a new file format.** Parquet keeps the exported-files job (object storage → DuckDB/Postgres), untouched here. This adds the **no-reparse wire encoding for results that never become files**: live `$sql-run` / `$sqlquery-run` responses that analysis clients consume over the HFS API. Today those arrive as CSV/JSON the client parses row by row, re-deriving types the engine already resolved; as an Arrow IPC stream (`application/vnd.apache.arrow.stream`), DuckDB, pyarrow, polars, and pandas ingest the same results directly — schema carried in-band, zero client-side parsing. It deliberately serves both sides of the analytics draft's "DuckDB-in-kernel vs pandas" open question.

## Design

The engine already built Arrow RecordBatches internally to write Parquet, then threw them away. This PR turns that into a shared seam:

- **Shared batch builder** — `estimate_rows_per_batch` + `for_each_record_batch` + public `run_view_definition_record_batches()`; the Parquet writer is refactored onto it, so schema inference, type mapping, and batching are identical across the columnar outputs *by construction* (proven by the untouched Parquet test suite staying green).
- **`ContentType::ArrowIpc`** — one new arm in `format_output`, reachable as `_format=arrow` or `Accept: application/vnd.apache.arrow.stream` through the existing negotiation on `sof-cli`, `sof-server`, and the REST `$sql-run`/`$sqlquery-run` handlers. Batches are written incrementally to the stream.
- **Chunked NDJSON streaming** rejects ArrowIpc like Parquet — now via a shared `reject_columnar_for_streaming` guard returning a proper error instead of `unreachable!()`.
- `arrow` is a server extension beyond the spec's `OutputFormatCodes` value set; the canonical bindings are unchanged.

Type mapping is inherited from Parquet exactly (bool→Boolean, integer→Int32, decimal→Float64, else Utf8, multi-valued→List) including its known caveats (values-inferred schema, decimal precision) — fixing those is a separate issue affecting both outputs together.

## Consumer view

```python
resp = client.post(f"{hfs}/ViewDefinition/$sql-run",
                   headers={"Accept": "application/vnd.apache.arrow.stream"}, ...)
table = pa.ipc.open_stream(resp.content).read_all()   # pyarrow
df    = pl.from_arrow(table)                          # polars
duckdb.sql("select gender, count(*) from table group by 1")
```

## Testing

Written red-first: 5 lib-level tests (MIME parsing round-trip, IPC stream round-trip incl. null survival, value equivalence vs JSON output, List/multi-valued columns across the boundary, streaming rejection with a real error) plus 2 server tests (Accept-header and `_format=arrow` negotiation end-to-end, response body read back with `StreamReader`). Full `helios-sof` suite: **195 tests across 37 binaries, all green** — including the untouched Parquet tests proving the refactor changed nothing. Clippy clean under CI's exact flags; `cargo check --workspace` clean (the one dependent match, in pysof, is updated here).

A client-parse benchmark (pandas `read_csv` vs `pa.ipc.open_stream` on a 50k-row result) is running and will be posted as a comment.

https://claude.ai/code/session_01XTNGEdcmYyKQoxfZ6KZpGp

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -456,8 +456,10 @@ jobs:
         working-directory: crates/pysof
         run: uv run maturin develop
 
-      # The workspace test loops skip pysof (it is not a default member), so
-      # its Rust suites run here. maturin develop above already compiled the
+      # Runs pysof's Rust suites in the job that owns the crate. (test-rust's
+      # `cargo test --workspace` does also select pysof — `--workspace`
+      # overrides default-members — but this keeps a pysof-scoped signal next
+      # to the Python tests.) maturin develop above already compiled the
       # dependency tree into the shared target dir with the same dev profile.
       - name: Run Rust tests
         working-directory: crates/pysof
--- a/crates/fhirpath-support/src/lib.rs
+++ b/crates/fhirpath-support/src/lib.rs
@@ -1005,6 +1005,27 @@ impl EvaluationResult {
         }
     }
 
+    /// Returns true when this is the Object representation of a FHIR primitive
+    /// element that carries id/extension metadata but no value (e.g. a
+    /// `_field` sibling holding only a data-absent-reason extension).
+    ///
+    /// Such an element is present in the tree (`exists()` is true) but has no
+    /// system value, so value-consuming operations must treat it as empty
+    /// input (see `hasValue()`/`getValue()` in the FHIR spec). The producers
+    /// tag this case with type `FHIR.Element` (`Element<V, E>`) or
+    /// `FHIR.decimal` (`DecimalElement<E>`); valued primitives are scalar
+    /// variants, never Objects, and genuine complex types carry their own
+    /// struct name, so those tags are unambiguous.
+    pub fn is_valueless_primitive_element(&self) -> bool {
+        match self {
+            EvaluationResult::Object {
+                type_info: Some(ti),
+                ..
+            } => ti.namespace == "FHIR" && (ti.name == "Element" || ti.name == "decimal"),
+            _ => false,
+        }
+    }
+
     // === Value Extraction Methods ===
 
     /// Extracts the boolean value if this is a Boolean variant.
--- a/crates/fhirpath/src/evaluator.rs
+++ b/crates/fhirpath/src/evaluator.rs
@@ -3306,6 +3306,80 @@ fn apply_regex_flags(builder: &mut RegexBuilder, flags: &str) {
     }
 }
 
+/// Functions that consume the *value* of their input (and of their
+/// arguments), as opposed to treating it as a node: string manipulation,
+/// type conversion, and math. A FHIR primitive element that carries
+/// extensions but no value is present as a node (`exists()` is true), but
+/// has no system value, so these functions must receive it as empty input
+/// and return empty (see `EvaluationResult::is_valueless_primitive_element`
+/// and the fhir-test-cases `primitivesWithoutValue` group). Existence and
+/// navigation functions (`exists`, `count`, `extension`, `hasValue`, ...)
+/// are deliberately not listed: for those the node itself is the input.
+const VALUE_CONSUMING_FUNCTIONS: &[&str] = &[
+    // String functions
+    "length",
+    "toChars",
+    "substring",
+    "upper",
+    "lower",
+    "indexOf",
+    "lastIndexOf",
+    "startsWith",
+    "endsWith",
+    "contains",
+    "replace",
+    "matches",
+    "matchesFull",
+    "replaceMatches",
+    "join",
+    "split",
+    "trim",
+    "escape",
+    "unescape",
+    "encode",
+    "decode",
+    // Conversions
+    "toString",
+    "toBoolean",
+    "toInteger",
+    "toLong",
+    "toDecimal",
+    "toDate",
+    "toDateTime",
+    "toTime",
+    "toQuantity",
+    "convertsToBoolean",
+    "convertsToInteger",
+    "convertsToLong",
+    "convertsToDecimal",
+    "convertsToString",
+    "convertsToDate",
+    "convertsToDateTime",
+    "convertsToTime",
+    "convertsToQuantity",
+    // Math
+    "abs",
+    "ceiling",
+    "exp",
+    "floor",
+    "ln",
+    "log",
+    "power",
+    "round",
+    "sqrt",
+    "truncate",
+];
+
+/// Maps a value-less primitive element (extensions but no value) to Empty and
+/// leaves every other result untouched. Used at value-consuming sites only.
+fn valueless_element_as_empty(value: &EvaluationResult) -> &EvaluationResult {
+    if value.is_valueless_primitive_element() {
+        &EvaluationResult::Empty
+    } else {
+        value
+    }
+}
+
 /// Calls a standard FHIRPath function (that doesn't take a lambda).
 fn call_function(
     name: &str,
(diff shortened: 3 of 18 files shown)


dougc95:
## Benchmark results: client-side cost of consuming a live query result

50,000-row Patient result (7 columns), rendered by release `sof-cli` in both wire formats, then timed on the client side — what a Python analysis process must do to obtain a usable table. Best of 5, Python 3.11, Windows 11, 12 logical CPUs:

| step | CSV | Arrow IPC |
|---|---|---|
| wire size (uncompressed) | 3.2 MB | 3.9 MB |
| to an Arrow table | `pandas.read_csv`: **58.9 ms** | `pa.ipc.open_stream(...).read_all()`: **0.5 ms** |
| on to a pandas DataFrame | (already there) | +0.8 ms (`to_pandas()`) |

**Net: ~114x faster to a columnar table, ~46x if the destination is a pandas DataFrame.** The Arrow number is sub-millisecond because "reading" an IPC stream is mapping the buffers, not parsing — this is the zero-parse claim made concrete. DuckDB and polars consume the Arrow table with no further conversion, so the 114x figure is the relevant one for the analytics-kernel path.

Honest caveats: this measures client-side parse cost only (server-side encode and network transfer excluded — encode cost is comparable between the formats since both walk the same RecordBatches). Arrow IPC is ~20% larger uncompressed; both formats compress normally under HTTP content-encoding. The margin grows with row count and column width, and shrinks for tiny results where fixed overhead dominates.

Reproduce with `crates/sof/scripts/bench_arrow_client.py` (committed on this branch): build `sof-cli --release`, then `uv run --with pyarrow --with pandas python crates/sof/scripts/bench_arrow_client.py`.

dougc95:
## Review-panel verification (3 independent reviewers) + fixes

Before trusting the CI retrigger, a three-lens review panel (exhaustiveness / semantics / CI-parity) audited whether the previous fix was complete. Unanimous verdict: **it was not** — and everything found is now fixed in `cfd578115`:

**Would have failed CI again (all fixed):**
- Three E0004s remained in **pysof's Rust test targets** (`tests/integration.rs`, `tests/lib_coverage_tests.rs` ×2 — two `ContentType` matches and one `SofError` match missing the new variants). Nobody saw them because cargo aborts on the first failing crate, and the Test Python job (`needs: [test-rust]`) was skipped on the red runs. All three now covered and exercising `arrow`; `cargo test` in `crates/pysof` green (6 suites).

**Behavior gaps beyond compilation (all fixed):**
- `sqlquery.rs` has its **own** `resolve_format` — `Accept: application/vnd.apache.arrow.stream` on a SQLQuery/SQLView Library subject silently fell through to NDJSON while the identical request on a ViewDefinition subject returned Arrow. Now symmetric.
- `sof-cli --format arrow` on an `.ndjson` bundle entered the streaming path and hard-errored, while `--format parquet` fell back to batch. Now both columnar formats take the batch path — verified end-to-end (`.ndjson` → arrow bytes → `pyarrow` round-trip).
- `sof-server` sent `Content-Disposition: attachment` only for Parquet; Arrow responses would render as binary noise in a browser. Now both columnar formats download.

**Confirmed sound by the panel (no changes):** the helios-rest handler fix itself; `$sql-export` exclusion (its kick-off allowlist 400s on `arrow` rather than silently downgrading); no other exhaustive match anywhere in the workspace; MIME strings byte-identical across all seven sites; the Binary-envelope path; Linting and fmt pass workspace-wide.

**One deliberate side effect worth review attention:** RecordBatch-construction failures on the *Parquet* path now surface as `SofError::ArrowConversionError` instead of `ParquetConversionError` (shared batch builder). No caller matches on the specific variant (verified: catch-alls everywhere), but it is an error-taxonomy change.

Also corrected: the ci.yml comment (and prior commit message) claiming the workspace test loops skip pysof — `--workspace` overrides `default-members`, so Test Rust does build pysof's tests; the real earlier miss was cargo aborting before reaching them.

dougc95:
**Note for reviewers:** commit afedde242 (`fix(fhirpath): treat extension-only primitives as empty in value-consuming operations`) is unrelated to the Arrow IPC feature — context:

The **Test FHIRPath** job on this PR started failing through no change here: CI checks out `FHIR/fhir-test-cases` at `master`, and upstream added a `primitivesWithoutValue` test group on 2026-08-21 that exposed a pre-existing evaluator bug (#716). String functions (`length()`, `substring()`, `upper()`, `toChars()`) and comparisons on a FHIR primitive element that carries only an extension and no value (e.g. `"given": [null, ...]` + `_given`) raised type errors instead of treating the absent value as empty input → empty result.

The fix is cherry-picked onto this branch so the pipeline can go green without waiting on a separate PR; full root-cause analysis is in #716. Verified locally: the full R4 + R5 conformance suites pass (R5 including the new upstream group, spliced into the vendored manifest) plus 11 new regression tests; `exists()`/`hasValue()`/navigation semantics are unchanged.

smunini (approved):

PR #711, opened 2026-08-26
fix(bulk-submit): scope pooled connections; the status poll names a stall

### LINKED ISSUE #646: bulk-submit: ingestion freezes silently mid-run with the default worker concurrency on SQLite
Found in #448's Synthea sanity pass. $bulk-submit of 40,538 NDJSON resources (19 types) against stock SQLite:

- **`HFS_BULK_SUBMIT_WORKER_CONCURRENCY=2` (the default): ingestion froze mid-run** — every in-flight file stopped at the same moment (Observation 1,224/21,255, Claim 325/4,231, Condition 48/698 …), **zero errors logged** even at debug, and the status poll URL answers `202` in-progress forever. No lease/stall recovery kicked in.
- **`WORKER_CONCURRENCY=1`: the same submission ingests completely and correctly** (all 19 types reconcile line-for-line against the NDJSON, dedup by id included), at ~23 resources/s.

Smells like a deadlock between the two workers over the SQLite connection. Two asks in one:
1. The deadlock itself (or whatever stops the workers).
2. **The observability gap**: a stopped worker leaves the submission permanently in-progress with nothing in the log — the lease/heartbeat machinery (`HFS_BULK_SUBMIT_LEASE_DURATION`=60s default) never surfaced the stall. A silent freeze should become a failed/stalled status a poller can see.

Repro: Synthea 100-patient NDJSON via a local static provider, stock config, SQLite.

angela-helios:
Closes #646 — both halves.

## The deadlock, reproduced and fixed

`process_entries` (SQLite) took a pooled **sync** connection and held it across the entry loop's awaits, while every per-entry call (`read`, `create_or_update`, `record_change`, `store_entry_result`) pooled its own. Two concurrent workers froze mid-run with nothing logged — exactly the #448 field report.

**Reproduced deterministically**: a test driving two `DefaultSubmitWorker`s over multi-file manifests hits the freeze at 2×20k lines in release (**300 s timeout**, one worker never finishing). With the connection scoped to its statements, the identical run **completes in 207 s**. Postgres carried the same held-client shape in its `process_entries` and gets the same scoping preventively.

## The observability gap

A `processing` manifest whose worker lease expired without renewal or reclaim polled as a quiet `202 processing` forever — a dead worker pool was indistinguishable from progress. Now:

- `SubmissionManifest` carries `lease_expiry` (wired in sqlite/postgres/mongodb).
- The status poll answers `X-Progress: stalled at N% - a worker stopped without handoff; see server logs` plus a warning log once a lease is **three durations** past expiry. State is not mutated — reclaim stays the authority, and a remote-worker deployment may still legitimately pick the manifest up.

## Tests

- `sqlite_bulk_submit_concurrency.rs`: two concurrent raw streams, and the full two-worker shape over multi-file manifests, both under a hard timeout at CI-friendly scale — a returning freeze fails with a name instead of hanging a runner.
- `test_poll_reports_a_stalled_ingestion`: a zero-duration lease makes the stall immediate; the poller sees "stalled".
- Full `helios-persistence` suite and `bulk_submit` rest suite green; all-backend feature build clean.

(The `router_http` mounts gain the eleventh argument main still lacks, as on the sibling branches — resolves on merge.)


angela-helios:
The Test FHIRPath failure here is not from this PR: the conformance job checks out `FHIR/fhir-test-cases@master`, and upstream added the `primitivesWithoutValue` group on Aug 21, which our evaluator failed. #717 fixes the evaluator (and vendors the new group into the local R5 suite); once it merges, a rerun here goes green.

smunini (approved):

Issue #674, opened 2026-08-25
ui: Batch/Transaction preflight page spacing, icons, and REQUEST strip font do not match the design


smunini:
## Summary

The Batch / Transaction preflight screen (`/ui/batch`) does not match the original design. Content inside the preflight card sits flush against the card edges, the card headers and entry rows have no icons, the footer buttons are unevenly spaced, the `REQUEST` strip renders in a font/size that does not belong to the app's type scale, and the footer carries an "Upload another" button that should not be there.

## Current behavior

Load `/ui/batch`, drop in a batch bundle with several entries, and the preflight stage renders:

- The tab row (`Actions` / `Bundle JSON`) and the entry list are direct children of `.card` with no horizontal padding, so the tabs and every row run edge-to-edge against the card border. `.card` itself carries no padding (`crates/ui/assets/app.css:642`); only `.card-head` pads its own content (`:3388`). `.batch-tabs` (`:4432`) and `.batch-rows` (`:4454`) add none.
- The preflight card has no `card-head` at all — the tab row is the first thing in the card, with `margin-bottom: 8px` and nothing above it.
- No icons anywhere on this stage. `icons/export.svg` is used only on the Stage 1 upload card head (`crates/ui/templates/pages/batch.html:34`); the preflight card, the response card head (`:77`), and the per-entry rows have none. The row disclosure control is a literal `▾` text glyph (`crates/ui/assets/batch.js:135`) rather than `icons/chevron-down.svg`.
- The footer (`crates/ui/templates/pages/batch.html:64`) is `justify-content: space-between` with `margin-top: 14px` and no side padding (`crates/ui/assets/app.css:4527`), so `Cancel` and the right-hand group hug the outer edges and do not line up with the card above them.
- `Upload another` sits between `Cancel` and `Execute` (`crates/ui/templates/pages/batch.html:67`), duplicating what `Cancel` already does — `batch-cancel` calls `reset()` and `batch-upload-another` re-opens the file picker (`crates/ui/assets/batch.js:178-181`).
[...]

PR #712, opened 2026-08-26
feat(ui): SQL workspaces match the Figma frame — spacing, fold, title row

### LINKED ISSUE #649: Promote SQL on FHIR to a top-level nav section with View Definitions, SQL Queries, SQL Views, Export, and Files

angela-helios:
Part of #649 — the design-fidelity pass the issue's acceptance criteria still owed ("the ViewDefinition Editor matches Figma node 420-2"), plus the spacing/alignment debts across the SQL pages.

## Against the frame

- **Layout**: View Definitions and SQL Queries/Views move to the two-column `filter-layout--two` — the main panel fills the width like the frame, no phantom third column.
- **Title row**: the view's name left; **Run / Duplicate / Delete / + Create New** right, iconed and spaced (Create New primary, per the frame).
- **JSON fold**: the definition lives in a collapsible `<details>` with a chevroned `card-head` — the frame's expandable panel — and the editor is a proper mono inset pane (`.json-editor`: 13px mono, inner box, vertical resize) instead of a bare 40px-styled textarea. Save sits at the fold's foot.
- **Rail**: name left, resource type (or library status) right as quiet muted text (`.filter-rail__meta`) — no more number-pill hack.

## Structural spacing fixes

- `.card__body` (padding for form panes and field lists — `.card` itself is deliberately padding-free for tables, which left bare content flush against the border: the same defect #674 fixed on batch).
- `.form-actions` for right-aligned form feet; `.btn` gains the icon-to-label gap.
- Export and Files pick up the same insets.
- The inner selected-name heading is an `<h2>` — one `<h1>` per page, as the design-system guard enforces (it caught the double-h1 during this pass).

## Verified

Full `helios-ui` Rust suite; Playwright design-system, a11y (light+dark), and both SQL workspace specs green; visually checked against the Figma geometry with headless screenshots.

Left as *Part of* rather than *Closes* so @angela-helios can close #649 when the finish level satisfies her review of the preview.


angela-helios:
The Test FHIRPath failure here is not from this PR: the conformance job checks out `FHIR/fhir-test-cases@master`, and upstream added the `primitivesWithoutValue` group on Aug 21, which our evaluator failed. #717 fixes the evaluator (and vendors the new group into the local R5 suite); once it merges, a rerun here goes green.

Issue #675, opened 2026-08-25
ui: Batch/Transaction — top-of-page Cancel/Execute, replace Back to Bundle and Execute again with a single Done


smunini:
## Summary

Rework the action buttons on the Batch / Transaction page (`/ui/batch`). The preflight stage should expose `Cancel` and `Execute` at the top of the page as well as the bottom, and the response stage should end with a single `Done` button — replacing today's `Back to Bundle` and `Execute again`.

## Current behavior

Preflight stage (`crates/ui/templates/pages/batch.html:64-71`) has one action row, at the bottom of the page:

```
[ Cancel ]                     [ Upload another ]  [ Execute ]
```

With a large bundle, the entry list pushes `Execute` below the fold; there is no way to run or abandon the bundle without scrolling to the end of the list.

Response stage (`crates/ui/templates/pages/batch.html:86-89`) has:

```
[ Back to Bundle ]                              [ Execute again ]
```

- `batch-back` returns to the preflight stage with the parsed bundle intact (`crates/ui/assets/batch.js:205-207`).
- `batch-execute-again` re-POSTs the same bundle (`crates/ui/assets/batch.js:204`), which silently re-runs a mutation that already succeeded.

There is no `Done` affordance anywhere — after a successful run the user is left on the outcomes screen with only "go back" and "do it again".

## Expected behavior

Preflight stage:

- `Cancel` and `Execute` appear at the top of the page, above the request strip / entry list, in addition to the existing bottom row. Both copies drive the same handlers.
- No `Done` button is visible at this stage.

Response stage:

- `Back to Bundle` is removed and replaced by a single `Done` button.
- `Execute again` is removed.
- `Done` is only visible after a successful execute — it is not present on the upload or preflight stages, and not shown when the run failed outright.

## Evidence

- `crates/ui/templates/pages/batch.html:64-71` — preflight footer, the only action row on that stage
- `crates/ui/templates/pages/batch.html:86-89` — response footer with `batch-back` and `batch-execute-again`
- `crates/ui/assets/batch.js:178` — `batch-cancel` → `reset()` (clears the bundle, returns to upload)
- `crates/ui/assets/batch.js:204-207` — `batch-execute-again` and `batch-back` handlers to delete
- `crates/ui/assets/batch.js:218-227` — whole-bundle failure path: `renderResponse` returns early without calling `show("response")`, so a rolled-back transaction stays on preflight with an inline error and never reaches the response stage
[...]

Issue #581, opened 2026-08-19
UI: no progress indication while adding a tenant


smunini:
## Problem

Adding a tenant in the web UI is a lengthy operation (tenant provisioning does a fair amount of work: storage setup, seeding conformance/search-parameter resources, etc.), but the UI gives the user no indication that anything is happening. After clicking **Add tenant**, the form just sits there until the request finally returns and the table swaps in.

From the user's point of view the click appears to have done nothing, which invites double-submits and makes the app feel broken.

## Where

`crates/ui/templates/pages/tenants.html` — the add-tenant form posts via htmx:

```html
<form
  hx-post="/ui/tenants"
  hx-target="#tenant-rows"
  hx-swap="innerHTML"
>
  ...
  <button type="submit" class="button button--primary">{{ i18n.t("tenants-add-submit") }}</button>
</form>
```

There is no `hx-indicator`, no `hx-disabled-elt`, and no busy/pending state anywhere in the flow.

## Expected behavior

While the create request is in flight, the user should get clear feedback that the operation is running and may take a while:

- A visible progress/spinner indicator (htmx `hx-indicator`) tied to the form.
- The submit button disabled (`hx-disabled-elt="find button[type=submit]"` or similar) so the tenant can't be created twice by impatient clicking.
- Ideally a short "this may take a moment" style message, since provisioning is genuinely slow rather than instantaneous.
- Accessible announcement of the pending/complete state (e.g. `aria-busy` / live region) so it isn't purely visual.

## Notes

- Worth checking whether other long-running UI actions have the same gap, so the fix can introduce a reusable pending-state pattern rather than a one-off.
- New user-facing strings should go through the existing i18n locale files.

PR #714, opened 2026-08-26
feat(ui): shared busy indicator, applied to Batch/Transaction (#679)

### LINKED ISSUE #679: ui: add a reusable busy/working indicator, and apply it to Batch/Transaction upload and execute
## Summary

The UI has no shared "working" affordance. On the Batch / Transaction page, uploading a large bundle and executing it both freeze or stall with no feedback at all. Build one reusable busy indicator and apply it to the batch page first, with the other long-running operations in the app as follow-on adopters.

## Motivation

`/ui/batch` has two operations that can take a long time and neither says anything:

**Upload / parse.** `readFile()` (`crates/ui/assets/batch.js:74-91`) reads the file, `JSON.parse`s it, and calls `renderPreflight()`, which builds a DOM row per entry in one synchronous loop (`crates/ui/assets/batch.js:110-157`). A Synthea bundle with several hundred entries blocks the main thread through the whole parse-and-render. The drop zone stays in its idle state; the page appears hung.

**Execute.** `execute()` (`crates/ui/assets/batch.js:185-202`) fires the POST and returns immediately. The `Execute` button is not disabled and nothing indicates a request is in flight — so a user who thinks the click missed can press it again and POST the same bundle twice, creating a second set of resources.

The pieces for a fix exist but were never generalized:

- `.spinner` (`crates/ui/assets/app.css:1207-1221`), including a `prefers-reduced-motion` override, is used in exactly one place: the tenants provisioning row (`crates/ui/templates/partials/tenant_rows.html:44-46`).
- `.btn:disabled { cursor: progress }` (`crates/ui/assets/app.css:1651-1654`) implies a busy-button convention that nothing actually drives.
- Only two scripts disable a control during async work — `crates/ui/assets/nl-search.js:128` and `crates/ui/assets/conformance-crud.js:20` — and neither shows a spinner while doing it.
- htmx is loaded but no template uses `hx-indicator`.

## Proposed behavior

A single busy concept, defined once and reused:

- A **busy button** state: the button disables, keeps its width, and shows the `.spinner` in place of or beside its label for the duration of the operation. Re-entrant clicks are ignored.
- A **busy region** state for work that is not tied to one button (parsing and rendering a large bundle): an inline status line with the spinner and a short label, marked `role="status"` so it is announced.
- Both honor `prefers-reduced-motion`, as `.spinner` already does.
- The label is a Fluent string, not hardcoded English.

Applied to `/ui/batch`:

- Picking or dropping a file puts the upload card into the busy region state until the preflight list is on screen.
- Pressing `Execute` puts that button into the busy state until the response renders or the request fails, and further clicks are ignored while it is in flight.

## Acceptance criteria

- [ ] One CSS component plus one small JS helper (or an htmx `hx-indicator` convention) covering both the button and region cases, documented in `crates/ui/README.md` next to the other conventions
- [ ] Dropping a large bundle on `/ui/batch` shows the busy region within one frame of the file being picked, and clears when the preflight renders
- [ ] `Execute` shows the busy state for the duration of the POST and cannot be double-fired
- [ ] Reduced-motion users get the static (non-animating) form
- [ ] Existing ad-hoc busy handling in `nl-search.js:128` and `conformance-crud.js:20` moves to the shared helper
- [ ] A Playwright assertion that the busy state appears and clears on execute

## Evidence

- `crates/ui/assets/batch.js:74-91` — `readFile()`, no busy state around parse
- `crates/ui/assets/batch.js:110-157` — `renderPreflight()`, one synchronous DOM build per entry
- `crates/ui/assets/batch.js:185-204` — `execute()`, no button disable, no in-flight guard; two listeners bound to it
- `crates/ui/assets/app.css:1207-1221` — the existing `.spinner`
- `crates/ui/templates/partials/tenant_rows.html:44-46` — its only current use
- `crates/ui/assets/app.css:1651-1654` — `.btn:disabled { cursor: progress }`

## Suggested approach

1. Add `.btn--busy` (or a `data-busy` attribute) that reserves the label width and swaps in `.spinner`, plus a `.busy-status` inline region with `role="status"`.
2. Add a small helper in a shared asset — something like `withBusy(el, promise)` — that sets the state, ignores re-entrant calls, and clears on settle. Wire `batch.js`'s `execute()` through it.
3. For the parse path, yield to the browser before the heavy loop (`requestAnimationFrame` or a zero timeout) so the busy state actually paints before the main thread blocks. Chunked or virtualized rendering of the entry list would be a better fix for very large bundles but is a separate piece of work.
4. Survey the other long operations for adopters — editor save, saved-query execution, history diff, bulk import/export polling — and file follow-ups rather than converting them all here.

## Scope / out of scope

Build the shared component and adopt it on `/ui/batch`. Converting every other long-running operation in the app is explicitly follow-on work. Rendering performance for very large bundles (virtualization, chunking) is out of scope — this issue is about telling the user that work is happening.

dougc95:
busy-execute-light
busy-execute-dark
## Summary

One busy/working convention for the whole UI, in two lanes, applied first to `/ui/batch`:

- **Fetch-driven scripts** call the new shared `assets/busy.js` (`window.hfsBusy` — the crate's one exported global). `during(buttons, work)` disables the controls and stamps `aria-busy="true"` for the life of the promise `work()` returns; `region(el, label)` reveals a pre-rendered `.busy-status` `role="status"` element. The CSS ring keys off `aria-busy`, so a script cannot get the visuals without the semantics.
- **htmx controls** keep `hx-disabled-elt` (#581) — documented in the README rather than adding indicator machinery no current adopter needs.

`work` is a thunk, not a promise, so the re-entrancy guard runs before the request exists; the thunk wraps the whole render chain, so busy clears when the outcome is on screen, not at response headers. Reduced-motion users get the ring as a static glyph; `.btn--primary` gets an explicit white ring (accent-on-accent vanishes in dark theme). `.provisioning` is renamed into the new `.busy-status` primitive (one region class per the vocabulary rule).

## On /ui/batch

- Picking or dropping a file shows the busy region **before any file bytes arrive**; parse + preflight render defer one paint so the region is actually visible for large bundles. FileReader `error`/`abort` (folder drops, vanished files) clear the region and report instead of spinning forever.
- Execute busies **both** button copies (#675 added a top copy) and disables the Cancels with them — which also fixes a real crash: Cancel mid-flight nulled `bundle` and the settling `renderResponse` threw on it.
- Focus is re-homed on every stage transition (preflight on upload, Done on response, drop zone on reset), using containment checks because the focus fixup for a hidden trigger is asynchronous.
- Labels are Fluent keys (`en`/`es`/`de`), riding the page's existing `data-msg-*` pattern.

## Adopters migrated

`nl-search.js` (whose Enter/chip handlers could double-POST around the disabled button — now gated by the helper's guard) and `conformance-crud.js` (stays inert through its success navigation via a never-settling promise; a failed DELETE re-enables next to its inline error).

## Screenshots

busy-execute-dark
busy-execute-light
## Testing

- `cargo test -p helios-ui` — 127 tests, including new shell-shape and rendered-copy assertions in `router_http.rs` and the `.busy-status` rename in `tenants_http.rs`.
- Playwright: 8 new busy tests — busy appears/clears on execute with a parked POST, no-double-fire (synthetic-event re-entry), the whole-bundle failure path, reduced motion, an axe WCAG 2.2 AA scan of the **held-open** busy state, unreadable-file cleanup, a stale-label race harness, and both adopter migrations. Design-system and no-CDN guards pass.

## Follow-ups (out of scope per the issue)

Remaining long operations to adopt the helper later: editor save, saved-query execution, history diff, bulk import/export polling.

Closes #679

https://claude.ai/code/session_01Rm3dCBykU67pvQAuMnwTNN

### DIFF
--- a/.claude/skills/work-with-ui/SKILL.md
+++ b/.claude/skills/work-with-ui/SKILL.md
@@ -81,12 +81,16 @@ through to the normal REST surface.
 - `templates/partials/` — htmx-swappable fragments, no `<html>` wrapper;
   `{% include %}`d into pages so the first render and the swap emit identical markup.
 - `templates/icons/*.svg` — Figma exports, fills normalized to `currentColor`, inlined.
-- `assets/` — `htmx.min.js` (pinned), `app.css`, `fonts/`, `logo.png`, and the
-  per-page scripts: `theme.js`, `editor.js`, `resources.js`, `saved-queries.js`,
-  `batch.js`, `history.js`, `nl-search.js`, `resource-filter.js`, `conformance-crud.js`.
+- `assets/` — `htmx.min.js` (pinned), `app.css`, `fonts/`, `logo.png`, the
+  shared `busy.js` (#679, the crate's one exported global `window.hfsBusy`),
+  and the per-page scripts: `theme.js`, `editor.js`, `resources.js`,
+  `saved-queries.js`, `batch.js`, `history.js`, `nl-search.js`,
+  `resource-filter.js`, `conformance-crud.js`.
 
 `theme.js` loads **without `defer`**, before first paint, to avoid a FOUC; every
-other script is `defer`.
+other script is `defer`. Busy/working states go through `hfsBusy` for
+fetch-driven code and `hx-disabled-elt` for htmx controls — see
+`crates/ui/README.md` § Busy states.
 
 ## Rules of the road
 
--- a/crates/ui/README.md
+++ b/crates/ui/README.md
@@ -96,13 +96,17 @@ note the version bump in the commit.
 ### Client-side scripts
 
 Each is a small, self-contained IIFE. Page-specific scripts load with `defer`;
-`json-view.js` loads from the shared layout because its delegated behavior is
-used across workspaces. `theme.js` loads **without `defer`**, before first paint,
-to avoid a flash of the wrong theme.
+`json-view.js` and `busy.js` load from the shared layout because their behavior
+is used across workspaces — `busy.js` ahead of every page script (defer runs in
+document order) because page scripts call into it. `theme.js` loads **without
+`defer`**, before first paint, to avoid a flash of the wrong theme. `busy.js`
+is the crate's one exported global (`window.hfsBusy`): unlike the closed IIFEs
+it exists to be called by the others.
 
 | Asset | Owns |
 |---|---|
 | `theme.js` | Light/dark preference: stored choice → OS preference, plus the top-bar toggle |
+| `busy.js` | The shared busy states (#679): `during(buttons, work)` and `region(el, label)` |
 | `saved-queries.js` | Saved queries, the visual search builder, and the `/_user/settings` read/modify/write cycle |
 | `editor.js` | The schema-driven editor loop — posts the document to `/ui/editor/render` and swaps in the server's HTML |
 | `json-view.js` | Delegated folding and accessibility state for every server-rendered JSON view |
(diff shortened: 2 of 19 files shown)


dougc95:
The failing **Test FHIRPath** check is not caused by this PR — `main` is currently red with the same failure.

The conformance coverage gate exits with `COVERAGE-LOST` on 8 tests from the new `primitivesWithoutValue` group in the upstream `tests-fhir-r5.xml` suite: string functions and comparisons on extension-only primitives (e.g. `Patient.name.given.first().length()` where `_given` has an extension but no value) throw type errors instead of returning empty. That defect was filed as #716 and is fixed by afedde242, which currently ships with #710.

Once #710 merges (or the fix is cherry-picked to `main`), re-running the checks here should go green — no rebase needed, since CI tests the merge ref.

smunini (approved):

PR #720, opened 2026-08-26
Honor configured public base URL across HFS links

### LINKED ISSUE #629: HFS silently advertises localhost URLs when the public base URL is not configured
## TL;DR

When `HFS_BASE_URL` is unset, HFS silently treats `http://localhost:8080` as its public origin even when the server is bound or published somewhere else. HFS then puts that unreachable or incorrect origin into pagination links, resource URLs, metadata, `Location` headers, Bulk Data URLs, and subscription URLs. A client can receive a valid first response and fail on the next request, or send search parameters and opaque cursors to the wrong host.

This remains reproducible on `main` at `c76c62cbf5b5018eb30a9aeffbbe7d92da3cb393` as of 2026-08-26. Setting `HFS_BASE_URL` fixes the advertised URLs, which confirms that cursor and storage code are not the cause. HFS cannot infer an external Docker port, reverse-proxy hostname, TLS scheme, or path prefix. It should validate and log the configured public base, warn when the localhost default is unsafe, and keep request headers untrusted unless an explicit trusted-proxy mode exists.

## Summary

When HFS is exposed on a public origin that differs from its listener and `HFS_BASE_URL` is not set, the server silently advertises the default `http://localhost:8080` origin in public FHIR URLs.

Search pagination is the first blocking symptom: clients receive a valid first page, then follow an unreachable `Bundle.link[next].url`. The same configured base URL also feeds `Bundle.link[self]`, `Bundle.entry.fullUrl`, `CapabilityStatement.implementation.url`, `Location` headers, history and compartment bundles, Bulk Data URLs, and subscription-related URLs.

This is a REST and public URL configuration defect. Storage adapters generate opaque cursors but do not choose the URL origin.

Related QA campaign: #556

## Environment

- HFS `0.2.1`
- Revision `3b46edc07b343ff9ea97a694e80147439e89e0ad`
- FHIR R4
- Docker port mapping `18080:8080`
- Runtime-confirmed backend: SQLite
- `HFS_BASE_URL` unset

### 1. Start HFS with a host port different from the container port

```bash
docker run --rm --name hfs-base-url-repro \
  -p 18080:8080 \
  -e HFS_SERVER_HOST=0.0.0.0 \
  -e HFS_SERVER_PORT=8080 \
  -e HFS_STORAGE_BACKEND=sqlite \
  -e HFS_DATABASE_URL=:memory: \
  -e HFS_DEFAULT_FHIR_VERSION=R4 \
  ghcr.io/heliossoftware/hfs:0.2.1
```

Do not set `HFS_BASE_URL`.

### 2. Create enough resources to paginate

```bash
for id in 01 02 03; do
  curl -fsS -X PUT "http://127.0.0.1:18080/Patient/base-url-repro-${id}" \
    -H 'Content-Type: application/fhir+json' \
    --data "{\"resourceType\":\"Patient\",\"id\":\"base-url-repro-${id}\",\"active\":true}"
done
```

### 3. Request a one-item page

```bash
curl -fsS \
  'http://127.0.0.1:18080/Patient?_count=1&_sort=_id' \
  -o first-page.json

jq '.link' first-page.json
```

The response advertises the default origin:

```text
http://localhost:8080/Patient?_count=1&_sort=_id
http://localhost:8080/Patient?_count=1&_sort=_id&_cursor=...
```

### 4. Follow the advertised next link unchanged

```bash
NEXT_URL="$(jq -r '.link[] | select(.relation == "next") | .url' first-page.json)"
curl -v "$NEXT_URL"
```

The request fails before reaching HFS when nothing is listening on host port `8080`.

### 5. Keep the cursor and change only the origin

```bash
NEXT_PATH="${NEXT_URL#http://localhost:8080}"
curl -fsS "http://127.0.0.1:18080${NEXT_PATH}" | jq '.entry | length'
```

This request succeeds. The cursor and storage query are valid; only the advertised origin is wrong.

## Positive control

Restart the same container with:

```bash
-e HFS_BASE_URL=http://127.0.0.1:18080
```

Repeat the search. `self`, `next`, `first`, and `entry.fullUrl` now use `http://127.0.0.1:18080`, and the literal `next` URL returns the following page.

## Campaign result

The observed Patient search returned:

- Page 1: 20 resources
- Page 2: 20 resources
- Page 3: 15 resources
- Total: 55 unique resources
- Duplicate IDs: 0

The advertised `next` URL failed with curl exit code 7 and HTTP status `000`. Replacing only `http://localhost:8080` with `http://127.0.0.1:18080` allowed the complete traversal.

The integrated console follows a pagination link on localhost port 8080 and shows an error

## Cross-backend scope

| Backend mode | Expected scope | Verification status |
| --- | --- | --- |
| `sqlite` | Affected | Runtime confirmed |
| `postgres` | Affected | Shared REST path confirmed; runtime matrix pending |
| `mongodb` | Affected | Shared REST path confirmed; runtime matrix pending |
| `sqlite-elasticsearch` | Affected | Shared REST path confirmed; runtime matrix pending |
| `postgres-elasticsearch` | Affected | Shared REST path confirmed; runtime matrix pending |
| `mongodb-elasticsearch` | Affected | Shared REST path confirmed; runtime matrix pending |
| `s3` | Not applicable | Standalone S3 does not implement search |
| `s3-elasticsearch` | Affected | Shared REST path confirmed; runtime matrix pending |
| Future `SearchProvider` implementations | Affected | REST builds links after adapter search |

For each searchable backend, seed at least three resources, request `_count=1`, inspect `self`, `next`, `first`, and `fullUrl`, follow `next`, then repeat with an explicit `HFS_BASE_URL`.

Backend-specific storage settings change, but the public URL failure is introduced after `storage.search()` returns. The non-SQLite rows above are code-path findings, not claims of completed runtime coverage.

## Deployment scope

Reproduce and cover these public-origin shapes:

- Docker or Kubernetes port publishing where the external port differs from the listener.
- HTTPS termination in a reverse proxy while HFS listens over HTTP.
- A public hostname that differs from the container or pod hostname.
- A public path prefix such as `https://example.test/fhir`.
- URL-path tenant routing where the tenant segment must be preserved.
- Direct local development where the default localhost URL is valid.

`HFS_BASE_URL` should remain authoritative unless a separate trusted-proxy mode is explicitly configured. HFS cannot reliably infer Docker or Kubernetes port mappings from inside the process.

## Additional affected surfaces

With the same configuration, inspect metadata and search URLs:

```bash
curl -fsS http://127.0.0.1:18080/metadata \
  | jq -r '.implementation.url'

curl -fsS 'http://127.0.0.1:18080/Patient?_count=1' \
  | jq '{links: .link, firstFullUrl: .entry[0].fullUrl}'
```

The observed responses advertised `http://localhost:8080` in `CapabilityStatement.implementation.url`, Bundle links, and `entry.fullUrl`.

The same global base URL is used by:

- Search and system-search Bundle links.
- Resource `fullUrl` values.
- Create and update `Location` headers.
- CapabilityStatement implementation URL.
- History and compartment bundles.
- Bulk Data URLs.
- Subscription and WebSocket URLs.

Each surface should be tested because a single fix must produce a consistent public origin across the REST API.

## Code path

- [`HFS_BASE_URL` defaults to `http://localhost:8080`](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/rest/src/config.rs#L918-L920).
- [Search builds its self link from `state.base_url()` after the adapter returns](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/rest/src/handlers/search.rs#L393-L397).
- [`SearchResult::to_bundle` derives paging links and `fullUrl` values from that base](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/persistence/src/core/search.rs#L105-L152).
- [The integrated console follows the supplied absolute paging URL](https://github.com/HeliosSoftware/hfs/blob/3b46edc07b343ff9ea97a694e80147439e89e0ad/crates/ui/assets/saved-queries.js#L1581-L1610).

SQLite, PostgreSQL, MongoDB, and Elasticsearch produce cursor data. None of them selects the public origin. The failure is therefore shared by adapters that reach this REST conversion path.

FHIR R4 treats paging links as opaque URLs that clients should follow as supplied. The console is correct to use the server-provided link, so silently rebasing absolute links in the browser would hide the server defect and could change the destination of a valid remote FHIR server.

## Actual result

HFS starts normally and does not report that it will advertise `http://localhost:8080`.

The server then returns absolute URLs that do not match the reachable deployment origin. Any client that follows the supplied FHIR links loses access after the first page.

The integrated console clears the current result table and shows a generic error after following the unreachable URL.

## Expected result

HFS must not silently advertise a default public origin in a deployment where that origin may be unreachable.

`HFS_BASE_URL` should remain the authoritative public URL unless a separate, documented request-origin or trusted-proxy mode is introduced. HFS must validate and clearly report the public URL selected at startup.

The server must not trust `Host`, `Forwarded`, or `X-Forwarded-*` headers by default. Proxy-derived URLs require an explicit trusted-proxy configuration to prevent host-header injection.

## Impact

- Every page after the first can become inaccessible.
- Every searchable backend inherits the same failure.
- Capability and resource URLs can point clients to the wrong service.
- Search parameters and opaque cursors can be sent to an unintended origin before CORS blocks the response.
- Names, identifiers, dates, and other sensitive search terms may appear in those URLs.
- The failure remains silent until a client follows an advertised URL.

No disclosure or data corruption was confirmed during the campaign.

## Proposed fix

1. Validate `HFS_BASE_URL` at startup.
   - Require an absolute `http` or `https` URL.
   - Reject credentials, query strings, and fragments.
   - Normalize the trailing slash.
2. Record whether the value was explicitly configured.
3. Log the advertised public base URL at startup alongside the bind address.
4. Warn prominently, or fail in a strict mode, when the default localhost URL is retained while HFS binds to a non-loopback interface.
5. Centralize public URL construction in REST and use it across search, metadata, `Location` headers, history, compartments, bulk operations, and subscriptions.
6. Keep forwarded-header support disabled by default. If added, restrict it to configured trusted proxies.
7. Improve the integrated console as defense in depth.
   - Preserve the previous result page when pagination fails.
   - Report the failed origin and mention `HFS_BASE_URL`.
   - Do not silently rewrite absolute FHIR paging links.

## Acceptance criteria

- HFS validates the configured public base URL before serving requests.
- Startup logs include the bind address and advertised public base URL.
- A non-loopback deployment that retains the default localhost base emits an actionable warning or fails under a documented strict mode.
- An explicit public base URL is used consistently by:
  - `CapabilityStatement.implementation.url`
  - `Bundle.link[self|next|previous|first]`
  - `Bundle.entry.fullUrl`
  - `Location` headers
  - History and compartment bundles
  - Bulk and subscription URLs
- Following `Bundle.link[next]` reaches the next page without client-side URL rewriting.
- Search pagination behavior is independent of the storage adapter.
- Forwarded headers cannot change advertised URLs unless trusted-proxy handling is explicitly enabled.
- URL-path tenant routing preserves the tenant prefix in every generated public URL.
- The integrated console does not discard the current page when pagination fails.
- Documentation covers Docker port mappings, reverse proxies, HTTPS termination, and path prefixes.

## Regression test matrix

Run the same contract for every searchable backend:

1. Public origin differs from listener, with explicit `HFS_BASE_URL`.
2. Default base retained on a non-loopback bind.
3. HTTPS public base behind a reverse proxy.
4. Public base with a path prefix.
5. URL-path tenant routing.
6. Spoofed `Host` and `X-Forwarded-*` headers with proxy trust disabled.
7. Trusted forwarded headers if proxy-aware mode is implemented.
8. Pagination through the final page with no missing or duplicate resources.
9. Search, metadata, `Location`, history, compartment, bulk, and subscription URL checks.

Backend suites should continue testing cursor correctness. REST tests should own origin selection and public URL consistency.

Gordex2014:
Closes https://github.com/HeliosSoftware/hfs/issues/629

## Summary

- Validate and normalize `HFS_BASE_URL` at HFS startup, preserve reverse-proxy path prefixes, and append dynamic values as encoded URL segments.
- Use one cached public URL builder for synchronous REST responses, Bulk Export, Bulk Submit, SQL-on-FHIR exports, Subscriptions, WebSocket discovery, and UI Bulk Import callbacks.
- Preserve URL-routed tenant provenance when JWT authentication is enabled, while keeping JWT claims authoritative for tenant validation.
- Keep header-routed asynchronous endpoints reachable in `TenantRoutingMode::Both` by sharing the reserved-path classifier between routing and tenant resolution.
- Make saved-query paging consume the server-provided Bundle links atomically. A failed page fetch preserves the current results and displays a localized `HFS_BASE_URL` diagnostic.
- Document public URL configuration and unsafe loopback/default-port startup warnings.

## URL behavior

| Input or context | Result |
|---|---|
| `HFS_BASE_URL=https://api.example.test/fhir/` | Normalizes to `https://api.example.test/fhir` and retains the `/fhir` prefix |
| URL-routed tenant | Includes the tenant segment in REST and protected asynchronous artifact URLs |
| Header-routed tenant | Does not add a tenant path segment; async status, file, and WebSocket routes remain system namespaces |
| Dynamic tenant, resource, or ID value | Appends the value as an encoded path segment |
| Presigned object-storage URL | Returns it unchanged |
| Invalid scheme, credentials, query, fragment, or missing host | Exits during startup with a configuration error |
| Loopback/default base inconsistent with the listener | Starts and logs a targeted warning with the bind and public addresses |

## Affected behavior

- CapabilityStatement implementation URL.
- Search Bundle paging links and entry `fullUrl`.
- Create/update `Location`, history and compartment links, and batch/transaction response locations.
- Bulk Export, Bulk Submit, and SQL-on-FHIR status and download URLs.
- Subscription notification sources and WebSocket endpoints.
- UI Bulk Import recipient URLs.
- Saved-query result links and page navigation.

Saved-query paging follows the Bundle link exactly, including opaque URLs. Page replacement is atomic. Network errors, non-2xx responses, and non-JSON payloads leave the existing table, count, ordering, and pager intact. The alert displays only the failing origin and points operators to `HFS_BASE_URL`.

## Verification

```text
cargo fmt --all -- --check
git diff --check
cargo check -p helios-rest --features s3,subscriptions
cargo test -p helios-rest -p helios-hfs -p helios-subscriptions -p helios-ui
cargo test -p helios-rest --test public_url_contract
cargo test -p helios-rest --features subscriptions ws::tests
cargo test -p helios-rest extractors::tenant::tests
cargo test -p helios-rest config::tests
npm run test:unit --prefix crates/ui/e2e
npx playwright test tests/queries.spec.ts --project=chromium <focused case>
```

Targeted Bulk Export, Bulk Submit, SQL-on-FHIR, WebSocket, and tenant-routing integration suites passed for both URL-path and header routing.

`cargo clippy ... --all-targets -- -D warnings` remains blocked by existing warnings outside this change in `crates/serde-support/src/lib.rs` and `crates/fhir/build.rs`.

## Review

Read-only reviews covered URL/configuration security, asynchronous protocols, and console paging. Follow-up reviews found no remaining actionable issues after fixes for:

- URL-route tenant provenance after JWT authentication.
- WebSocket token/request tenant mismatches.
- Header-routed async namespaces in `Both` mode.
- Opaque pager links losing resource context.
- Result links dropping configured prefixes or tenant paths.
- Bracketed IPv6 loopback listeners producing false warnings.

### DIFF
--- a/.agents/skills/run-hfs-server/SKILL.md
+++ b/.agents/skills/run-hfs-server/SKILL.md
@@ -33,8 +33,14 @@ HFS_SERVER_PORT=3000 HFS_LOG_LEVEL=debug cargo run --bin hfs
 | `HFS_SERVER_PORT` | `8080` | Server port |
 | `HFS_SERVER_HOST` | `127.0.0.1` | Host to bind |
 | `HFS_LOG_LEVEL` | `info` | Log level: error, warn, info, debug, trace |
-| `HFS_BASE_URL` | `http://localhost:8080` | Base URL for Location headers and Bundle links |
+| `HFS_BASE_URL` | `http://localhost:8080` | Public HTTP(S) base for Location headers and Bundle links |
 | `HFS_DATA_DIR` | `./data` | FHIR data directory, including search parameters |
+
+Set `HFS_BASE_URL` explicitly for containers and reverse proxies. Include any
+published path prefix. HFS rejects non-HTTP(S) values, credentials, query
+strings, and fragments, and it never derives the public base from request
+forwarding headers. A loopback base that conflicts with the listener produces
+a startup warning.
 
 ## Limits
 
--- a/README.md
+++ b/README.md
@@ -233,12 +233,18 @@ AWS_REGION=us-east-1 \
 | `HFS_STORAGE_BACKEND` | `sqlite` | Backend mode: `sqlite`, `sqlite-elasticsearch`, `postgres`, `postgres-elasticsearch`, `mongodb`, `mongodb-elasticsearch`, `s3`, or `s3-elasticsearch` |
 | `HFS_SERVER_PORT` | `8080` | Server port |
 | `HFS_SERVER_HOST` | `127.0.0.1` | Host to bind |
-| `HFS_BASE_URL` | `http://localhost:8080` | Base URL used in `Location` headers and Bundle links |
+| `HFS_BASE_URL` | `http://localhost:8080` | Public HTTP(S) base used in `Location` headers and Bundle links |
 | `HFS_DATABASE_URL` | `fhir.db` | Database URL (SQLite path or PostgreSQL connection string) |
 | `HFS_DATA_DIR` | `./data` | Directory containing FHIR data files (search parameters) |
 | `HFS_SEARCH_PARAM_CACHE_TTL` | `3600` | Seconds between refreshes of the in-memory SearchParameter registry from storage; a param POSTed to one cluster node becomes visible to others within this interval. `0` disables the refresh. |
 | `HFS_DEFAULT_FHIR_VERSION` | `R4` | FHIR version (R4, R4B, R5, R6) |
 | `HFS_LOG_LEVEL` | `info` | Log level (error, warn, info, debug, trace) |
+
+Set `HFS_BASE_URL` to the URL clients use, including any reverse-proxy path
+prefix, for example `https://fhir.example.com/fhir`. HFS accepts only absolute
+`http` or `https` URLs without credentials, query strings, or fragments. It
+removes trailing slashes at startup. HFS does not derive this value from
+`Host`, `Forwarded`, or `X-Forwarded-*` request headers.
 
 **Limits & behavior**
 
--- a/book/src/components/hfs-server.md
+++ b/book/src/components/hfs-server.md
@@ -47,8 +47,14 @@ The server is configured entirely through [environment variables](../configurati
 ```bash
 # Explicit example
 HFS_SERVER_PORT=3000 \
+HFS_BASE_URL=https://fhir.example.com \
 HFS_STORAGE_BACKEND=postgres \
 HFS_DATABASE_URL="postgresql://user:pass@localhost/fhir" \
 HFS_LOG_LEVEL=debug \
   ./hfs
 ```
+
+`HFS_BASE_URL` is the public origin HFS writes into FHIR responses. Include a
+reverse-proxy path prefix when one exists, for example
+`https://fhir.example.com/fhir`. HFS validates and logs the value at startup.
+It warns when a loopback base does not match the listener.
--- a/book/src/configuration/environment-variables.md
+++ b/book/src/configuration/environment-variables.md
@@ -9,9 +9,16 @@ All server behavior is controlled through environment variables. No configuratio
 | `HFS_SERVER_PORT` | `8080` | Server port |
 | `HFS_SERVER_HOST` | `127.0.0.1` | Host to bind |
 | `HFS_LOG_LEVEL` | `info` | Log level: `error`, `warn`, `info`, `debug`, `trace` |
-| `HFS_BASE_URL` | `http://localhost:8080` | Base URL for Location headers and Bundle links |
+| `HFS_BASE_URL` | `http://localhost:8080` | Public HTTP(S) base for Location headers and Bundle links |
 | `HFS_DATA_DIR` | `./data` | Path to FHIR data directory (search parameters) |
 | `HFS_SEARCH_PARAM_CACHE_TTL` | `3600` | Seconds between refreshes of the in-memory SearchParameter registry from storage. In a cluster, a SearchParameter POSTed to one node becomes visible to the others within this interval. `0` disables the periodic refresh. |
+
+`HFS_BASE_URL` must be an absolute `http` or `https` URL with a host. It may
+include a deployment path prefix, such as `https://example.com/fhir`, but it
+must not include credentials, a query string, or a fragment. HFS normalizes a
+trailing slash and rejects invalid values before it starts. Configure the
+public address explicitly behind a proxy. Request `Host`, `Forwarded`, and
+`X-Forwarded-*` headers do not change advertised URLs.
 
 ## Limits
 
--- a/book/src/getting-started/docker.md
+++ b/book/src/getting-started/docker.md
@@ -12,6 +12,7 @@ docker run -p 8080:8080 ghcr.io/heliossoftware/hfs:latest
 docker run -p 8080:8080 \
   -v hfs-data:/data \
   -e HFS_DATABASE_URL=/data/fhir.db \
+  -e HFS_BASE_URL=http://localhost:8080 \
   ghcr.io/heliossoftware/hfs:latest
 
 # With PostgreSQL
@@ -51,6 +52,11 @@ docker build --build-arg BINARY_NAME=fhirpath-server -t fhirpath-server .
 The Dockerfile expects the binary and `data/` files to be pre-staged in the build context. In CI, the binary is built separately and copied in before the Docker build step.
 
 > **Details:** Base image is `debian:bookworm-slim`. The server runs as non-root user `hfs`. Default exposed port is 8080. Host-binding environment variables (`HFS_SERVER_HOST`, `SOF_SERVER_HOST`, `FHIRPATH_SERVER_HOST`) are automatically set to `0.0.0.0` inside the container.
+
+Set `HFS_BASE_URL` to the address used outside the container. For a TLS proxy
+that publishes HFS below `/fhir`, use
+`-e HFS_BASE_URL=https://fhir.example.com/fhir`. HFS uses that configured value
+for response links and does not trust forwarding headers to infer it.
 
 ## Environment Variables
 
--- a/crates/fhirpath/src/evaluator.rs
+++ b/crates/fhirpath/src/evaluator.rs
@@ -1371,6 +1371,21 @@ fn primitive_member_access(
     }
 }
 
+/// Returns true for a FHIR primitive element that has metadata but no system value.
+///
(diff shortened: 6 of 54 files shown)


Gordex2014:
## Manual UI evidence

Query used in all captures: `GET /Patient?_count=1&_sort=_id`

### Before: first page

The server returns one Patient and an absolute pager link targeting the unreachable default public URL.

Before: first page

### Before: pagination failure

Following that link clears the rendered result and leaves only the generic error state.

Before: pagination failure clears results

### After: failed page fetch preserves state

With the same unsafe default, the first result, count, sort, and pager remain visible. The localized alert identifies `http://localhost:8080` and points to `HFS_BASE_URL`.

After: failed page fetch preserves results and shows HFS_BASE_URL diagnostic

### After: configured public URL paginates successfully

With `HFS_BASE_URL=http://127.0.0.1:18080`, Next loads the second Patient and shows no alert.

After: successful second page with configured public URL

PR #723, opened 2026-08-27
SoF conformance: ViewDefinition search parameters and fixture housekeeping

### LINKED ISSUE #570: SoF 3.0.0-ballot: register the ViewDefinition search parameters and advertise them
Becoming a resource brings the standard FHIR read and search machinery with it. 3.0.0-ballot publishes the standard metadata search parameters for ViewDefinition, so a server can be queried for views by `url`, `identifier`, `name`, `status`, `publisher`, `date`, `context` and the remaining metadata parameters, in the same manner as any other canonical resource.

Version 2.0.0 provided no means of discovering the views a server holds, because a logical model has no endpoint. That gap is now closed by the spec, and we should close it here.

## Current state

We store ViewDefinitions and resolve them by canonical URL (`crates/rest/src/handlers/sof/references.rs::resolve_by_canonical_url` issues a `url=` search), so *something* answers a `url` lookup today. What is missing is the registered SearchParameter set that makes `GET /ViewDefinition?...` work generally, and the CapabilityStatement entry advertising it.

The parameter bundle is published upstream at `input/resources/viewdefinition/bundle-ViewDefinition-search-params.xml`, under canonicals of the form `http://hl7.org/fhir/uv/sql-on-fhir/SearchParameter/ViewDefinition-{code}`.

## Work

- Add the ViewDefinition search parameters to `data/search-parameters-{r4,r4b,r5,r6}.json`, the files the registry loads.
- Confirm they survive `SearchParameterRegistry::load_from_spec_file`. Two known traps: `register()` rejects duplicate URLs but `params_by_type` uses `HashMap::insert`, so a same-code parameter under a different URL silently overrides the lookup; and `SearchParameterStatus::Draft` parameters are filtered out by `get_active_params()` (the spec-file loader converts draft→active, the embedded loader does not).
- Advertise them on the ViewDefinition entry in the CapabilityStatement. The spec also says that if a server supports CRUD and search for ViewDefinition, `rest.resource` SHALL carry an entry for it whose `interaction` array includes read, search-type, write, patch, delete and create — and the same for `Library`, of which SQLQuery and SQLView are profiles. Worth checking `build_resource_capability` already covers both types.
- Tests: `GET /ViewDefinition?url=…`, `?name=…`, `?status=…`, and a `_lastUpdated`/`date` range, at least on SQLite.

## Acceptance

- `GET /ViewDefinition?url=http://example.org/…` returns the matching view through the ordinary search path, not only through the SoF resolver.
- The metadata parameters listed above are advertised in the CapabilityStatement's ViewDefinition entry.
- ViewDefinition and Library both carry the full CRUD + search interaction set in `rest.resource`.

### LINKED ISSUE #571: SoF 3.0.0-ballot housekeeping: rename the conformance fixture directory, re-sync fixtures, sweep stale links
Loose ends left over from tracking SQL on FHIR 3.0.0-ballot. None of these change behaviour; they are naming and provenance.

## The conformance-fixture directory name

`crates/sof/tests/sql-on-fhir-v2/` is named for a release we no longer track. The fixtures inside are current — they include `repeat.json` and `row_index.json`, and the README already points at `FHIR/sql-on-fhir.js`, which is where the suite lives now that the specification moved to HL7 under CC0 while the reference implementation stayed with the FHIR Foundation under MIT.

Renaming the directory to something version-neutral (`sql-on-fhir/`, or `conformance/`) touches 13 path references across 6 test files. Purely mechanical; deferred out of the ballot work to keep that diff reviewable.

## Re-sync the fixtures

The README documents a `curl` loop that re-fetches each file by name, plus a note to pick up newly-added upstream files by hand. Worth an actual run — the suite is now versioned with the reference implementation rather than with the specification, so it moves on its own cadence and our copy can drift silently.

While re-syncing, check whether the `PASS_FLOOR` ratchets in the three in-DB conformance suites (`sof_conformance.rs`, `sof_conformance_postgres.rs`, `sof_conformance_mongodb.rs`, all at 132) should move. They are ratchets, so new upstream fixtures will push the total up without pushing the floor.

## Remaining `sql-on-fhir-v2` references in prose

A handful of doc links still point at `build.fhir.org/ig/FHIR/sql-on-fhir-v2`, which is the continuous build of the old repository. The published location is `http://hl7.org/fhir/uv/sql-on-fhir`. Most were updated; a sweep for stragglers (including in generated `crates/fhir/src/r*.rs` doc comments, which come from the vendored StructureDefinition) is worth doing once.

## Package identifier

The guide's package identifier changed from `org.sql-on-fhir.ig` to `hl7.fhir.uv.sql-on-fhir`. We do not consume the package today — `crates/fhir-gen` reads the StructureDefinition directly — so nothing depends on this. Recorded as `IG_PACKAGE_ID` in `crates/sof/src/canonical.rs` for whenever something does.

### LINKED ISSUE #718: Decide: keep or rename crates/sof/tests/sql-on-fhir-v2 (path stability vs version-neutral name)
## Context

Issue #571 asks to rename `crates/sof/tests/sql-on-fhir-v2/` to something version-neutral, since the fixtures now track SQL on FHIR 3.0.0-ballot rather than v2.

However, the directory's own README takes the opposite position — added in c572e551b (2026-08-19, one day after #571 was opened):

> The directory is still named `sql-on-fhir-v2` **for path stability**; the fixtures themselves track the current reference implementation.

So there are two documented, conflicting positions, and the rename should not be executed (or discarded) silently. The M2 housekeeping work will proceed **without** the rename; this issue exists to settle the naming question explicitly.

## The two positions

**Keep `sql-on-fhir-v2` (status quo, per the README):**
- Path stability: ~14 references across 3 crates (`sof`, `rest`, `persistence`) never churn again.
- External references (docs, muscle memory, scripts) keep working.
- The name is cosmetic; the README already explains the mismatch.

**Rename to a version-neutral name (`sql-on-fhir/` or `conformance/`, per #571):**
- The name stops lying about what the fixtures track.
- One mechanical change now vs a permanently confusing name.
- Every future reader pays a small comprehension tax with the current name.

## Ask

Decide one of the two and record it:
- **Keep** -> close this issue and strike the rename bullet from #571 (the rest of #571 — fixture re-sync, `PASS_FLOOR` ratchet, link sweep verification — proceeds either way).
- **Rename** -> the M2 ticket picks it up, and the README note is removed in the same change.

Refs: #571, c572e551b, `crates/sof/tests/sql-on-fhir-v2/README.md:12`.

andyguz17:
## Summary

**ViewDefinition search parameters (#570)**
- Add `data/sql-on-fhir-search-parameters.json`: the 17 `ViewDefinition` SearchParameters (`url`, `name`, `status`, `date`, `version`, `identifier`, `title`, `description`, `publisher`, `jurisdiction`, `resource`, `profile`, and the `context*` family) copied from the current official SQL-on-FHIR IG build (`hl7.fhir.uv.sql-on-fhir` 3.0.0-ballot) under their official core-namespace canonicals (`http://hl7.org/fhir/SearchParameter/ViewDefinition-{code}`). The file is loaded through the existing custom-file mechanism (any `data/*.json` not matching `search-parameters-*.json`), so the verbatim HL7 spec artifacts `data/search-parameters-r*.json` and the loader's production code are untouched.
- One deliberate, documented deviation from the published build: the two composite parameters (`context-type-quantity`, `context-type-value`) reference their components via legacy `uv/sql-on-fhir` canonicals upstream, which would never resolve in the registry (components would be silently skipped and the composites would index nothing). Our copy uses the core canonicals for those four `component.definition` values.
- New tests: loader/registry survival tests per FHIR version (all 17 registered, active, retrievable by code and canonical, benign collision with the embedded fallback), an extractor-level test proving the composite `context` parameters actually index a matching `useContext`, and an integration suite (`crates/rest/tests/view_definition_search_params.rs`, 25 tests) covering `GET /ViewDefinition?url|name|status|date|_lastUpdated`, the CapabilityStatement entry (full interaction set + advertised searchParams; `Library` full interaction set), and a `$reindex` round trip restoring searchability of pre-existing ViewDefinitions.
- Docs: the REST README notes that pre-existing ViewDefinitions require `$reindex` after upgrading to become searchable by the new parameters; the HFS README's data-directory tree lists the new file.

**Conformance fixture housekeeping (#571)**
- Re-sync the SQL-on-FHIR conformance fixtures from upstream (byte-verbatim). Only `tests/repeat.json` changed (upstream fixture-data fix: required `Questionnaire.status` added, an invalid childless `group` item changed to `display`). No fixtures added or removed.
- Rename the fixture directory from `crates/sof/tests/sql-on-fhir-v2/` to the version-neutral `crates/sof/tests/sql-on-fhir/`, as decided in #718: the fixtures track the current spec (3.0.0-ballot), not v2. Pure `git mv` (all 22 fixtures byte-identical), all path references updated, and the README's path-stability note removed in the same change. The one historical citation to the old `FHIR/sql-on-fhir-v2` GitHub repository (a PR reference in `emit.rs`) is kept as provenance.
- Pin the fixture JSON files to LF in `.gitattributes` so byte-verbatim comparison against upstream stays clean on Windows checkouts.
- Conformance pass counts are unchanged after the re-sync and rename: 132/144 on SQLite, PostgreSQL, and MongoDB (`PASS_FLOOR` stays 132; the 12 remaining failures are the documented engine gap).

## Testing
- `cargo fmt --all -- --check`, `cargo clippy -p helios-fhir -p helios-rest -p helios-persistence --all-targets -- -D warnings`
- `cargo test -p helios-fhir search` (default + `--all-features`), `cargo test -p helios-persistence --lib search::extractor`, `cargo test -p helios-rest` (full suite), `cargo test -p helios-sof --tests`
- Conformance suites on SQLite, PostgreSQL, and MongoDB (Docker): 132/144 each, floor met — re-verified after the directory rename
- Key suites re-run after merging `main` back into the branch: all green
- Manual smoke against a live server: create + search by `url`/`name`/`status`/`date`/`_lastUpdated`, CapabilityStatement entry verified

Fixes #570
Fixes #571
Closes #718


crates/fhir/src/search/loader.rs
```diff
@@ -952,4 +1015,146 @@ mod tests {
...
         );
     }
+
+    // ---- ViewDefinition SearchParameters survive the real loader/registry (#570) ----
+
+    /// The 17 SQL-on-FHIR IG canonicals shipped in every
+    /// `search-parameters-r*.json` file, keyed by their `code`.
+    const VIEW_DEFINITION_PARAM_CODES: &[&str] = &[
+        "context",
+        "context-quantity",
```
smunini, line 964: Please examine

smunini (changes requested):

smunini:
Tracking for the upstream composite-canonical inconsistency that `scripts/sync-sof-search-params.py` patches around (the composites' `component.definition` values reference legacy `uv/sql-on-fhir` canonicals no published SearchParameter declares):

- GitHub issue: HL7/sql-on-fhir#403
- chat.fhir.org thread in **#analytics on FHIR**: https://chat.fhir.org/#narrow/stream/179219-analytics-on-FHIR/topic/ViewDefinition.20SearchParameter.20canonicals (also asks whether post-ballot feedback should be moving to HL7 Jira)

Once fixed upstream, the script's canonical rewrite becomes a no-op and `data/sql-on-fhir-search-parameters.json` becomes a fully verbatim copy of the IG package.

https://claude.ai/code/session_01CktNg2DJpWsPFANE1VevR9

smunini (approved):

Issue #682, opened 2026-08-25
ui: Bulk Import dialog keeps typed values after cancel/close, and New Submission toggle shows a default browser focus ring


smunini:
## Summary

Two defects on `/ui/bulk-import`. Closing the "New Submission" dialog leaves everything the user typed in the form, so reopening it shows stale input. And the "New Submission" toggle keeps a browser-default blue focus ring after it is clicked, because `.btn` has no focus style of its own.

## Current behavior

**Stale form state.** Go to `/ui/bulk-import`, press "New Submission", type something into "Submission Name", then dismiss the dialog with the `×`, with Cancel, with Esc, or by clicking outside. Press "New Submission" again — the typed name is still there, along with every other field the user touched.

`addbox.js` closes the dialog by removing the `open` attribute (`crates/ui/assets/addbox.js:11-13`) and nothing calls `form.reset()`. The `<details>` element keeps its subtree in the DOM, so the input values survive. There is no page script for this form at all — `/ui/bulk-import` loads only the layout's shared scripts (`crates/ui/templates/layouts/base.html:12-17`).

The same holds for the "Add Manifest" dialog on `/ui/bulk-import/{id}` (`crates/ui/templates/pages/bulk-import-detail.html:14-50`) and for the Add Tenant panel on `/ui/tenants`. `tenants.js:16-25` resets its form on a *successful create* only; cancelling or closing leaves the input in place.

**Blue focus ring.** The dialog toggle is `<summary class="btn btn--primary">` (`crates/ui/templates/pages/bulk-import.html:13-16`). `app.css` defines `.btn`, `.btn:hover`, and `.btn:disabled` (`crates/ui/assets/app.css:1616-1654`) but no `.btn:focus-visible` or `summary:focus` rule, so the focusable `<summary>` falls back to the user agent's default focus ring — the blue box. It appears after the toggle is clicked and stays after the dialog is dismissed, since focus remains on the summary.

Only four `:focus-visible` rules exist in the whole stylesheet (`.brand`, `.stat--link`, `.addbox__x`, and one more), so this is the general state of focus styling rather than a one-off.

## Expected behavior

- Dismissing the dialog by any route — `×`, Cancel, Esc, click outside — clears the form. Reopening it starts blank.
- A submission that *fails* keeps the user's input, so they can correct and retry (this is the rule `tenants.js` already documents for its error path).
- The toggle shows the app's own focus indicator, matching the other focusable controls, not the UA default ring.

## Evidence

[...]

PR #736, opened 2026-08-27
fix(ui): Batch/Transaction round 3 — single footer, error placement, focus, header count, tab icons

### LINKED ISSUE #726: ui: Batch/Transaction preflight shows a white focus-ring border around its outer edge on the first file drop
## Summary

Dropping a JSON bundle onto the Batch/Transaction upload zone for the first time paints a white border around the outer edge of the preflight stage. It is a browser focus ring on an unstyled `tabindex="-1"` container, not an intentional border.

## Current behavior

1. Open `/ui/batch` in a fresh page load. Do not click anything.
2. Drag a batch or transaction Bundle JSON file onto the drop zone and release.
3. The preflight stage appears with a thin white outline tracing its entire outer edge.

The outline does not appear on subsequent runs in the same page session (Cancel or Done, then drop a second file), nor when the file is chosen by clicking the drop zone and using the file picker.

## Expected behavior

The preflight stage appears with no stray outline. Focus still lands on the revealed stage for screen-reader and keyboard users — that behavior is deliberate and must be preserved — but a pointer-driven drop should not paint a visible ring.

## Evidence

The revealed stage is focused programmatically:

```js
// crates/ui/assets/batch.js:190-198
/* The drop zone was hidden with its stage, stranding focus. ... */
var active = document.activeElement;
if (active === document.body || stages.upload.contains(active)) {
  stages.preflight.focus();
}
```

The focus target is a full-width section carrying `tabindex="-1"`:

```html
<!-- crates/ui/templates/pages/batch.html:52 -->
<section id="batch-preflight" tabindex="-1" hidden>
```

- `#batch-preflight` is the only `tabindex="-1"` element in `crates/ui/templates/`, and nothing in `crates/ui/assets/app.css` styles its focus state — so it falls through to the browser's default focus ring, which traces the whole section. The white half of Chrome's two-tone default ring is what reads as the "white border".
- The `.batch-drop` rules at `crates/ui/assets/app.css:4632-4655` are not involved: that element's border is `1.5px dashed var(--divider)`, and `.batch-drop--over` is removed on `drop` (`batch.js:120-122`).
- Introduced by `3d5227100` — "feat(ui): shared busy indicator, applied to Batch/Transaction (#679)".

The "first time only" part is consistent with focus-ring heuristics: on a fresh load the user has not clicked, so `document.activeElement === document.body` and the browser has no recent pointer interaction to suppress the ring against. After a Cancel or Done click, the subsequent programmatic focus is not treated as focus-visible. Worth confirming during the fix rather than assuming.

## Suggested approach

Give `#batch-preflight` (and `#batch-response`, if it ever becomes a focus target) an explicit focus style rather than leaving it to the UA. Two options:

- `#batch-preflight:focus { outline: none; }` — simplest, and defensible for a container that is only ever focused programmatically as a scroll/announce target, never keyboard-reachable via Tab.
- Or style `:focus-visible` to match the rest of the UI (`2px solid var(--accent)`, `outline-offset: 2px`, as `.btn:focus-visible` does at `app.css:1765-1766`) and rely on the browser to suppress it for pointer-driven focus.

Prefer whichever keeps the outline invisible for the drop path while leaving keyboard navigation unchanged. Verify in both light and dark themes — the ring's visibility differs between them.

## Scope / out of scope

Out of scope: the focus-management logic itself. Landing focus on the revealed stage is correct (#679) and should not be removed to make the outline go away.

### LINKED ISSUE #727: ui: add the design's Actions and Bundle JSON tab icons to the Batch/Transaction preflight
## Summary

The `Actions` and `Bundle JSON` tabs on the Batch/Transaction preflight stage are text-only. The design shows a leading icon on each. Add them.

## Motivation

Icon-plus-label is the established treatment for headings and controls on this page — issue #674 added icons to the preflight and response card heads and replaced the row disclosure glyph with `icons/chevron-down.svg` — but the tab row inside the preflight card was left as bare text:

```html
<!-- crates/ui/templates/pages/batch.html:74-77 -->
<div class="batch-tabs" role="tablist">
  <button type="button" class="batch-tab" id="batch-tab-actions" role="tab" aria-selected="true">{{ i18n.t("batch-tab-actions") }}</button>
  <button type="button" class="batch-tab" id="batch-tab-json" role="tab" aria-selected="false">{{ i18n.t("batch-tab-json") }}</button>
</div>
```

Every other heading on the same screen carries one: `icons/export.svg` on the upload card (`batch.html:37`), `icons/layers.svg` on the plan card (`:72`), `icons/check.svg` on the response card (`:95`). The two tabs sitting directly under the icon-bearing plan heading are the only labels on the stage without one.

## Proposed behavior

Each tab renders a leading 16px icon followed by its existing label, matching the design.

Icon selection is the design call to make. The repo's existing set (`crates/ui/templates/icons/`) already holds plausible candidates — `sliders.svg` or `layers.svg` for `Actions`, `code.svg` for `Bundle JSON` — but if the design specifies glyphs that are not in the set, add the new SVGs to `crates/ui/templates/icons/` in the same 16×16 `viewBox="0 0 16 16"` / `fill="none"` / `currentColor` form as the existing files. No runtime CDN; icons are inlined via `{% include %}`.

Markup follows the card-head idiom:

```html
<button type="button" class="batch-tab" id="batch-tab-actions" role="tab" aria-selected="true">
  <span class="icon">{% include "icons/sliders.svg" %}</span> {{ i18n.t("batch-tab-actions") }}
</button>
```

`.batch-tab` (`crates/ui/assets/app.css:4702-4711`) currently has no `display: flex`, so it will need `display: inline-flex; align-items: center; gap: 8px;` — the same treatment `.card-head h3` uses at `app.css:3624-3632` — for the icon and label to sit on a shared baseline.

The icon inherits `currentColor`, so it should pick up the selected-state color change from `.batch-tab[aria-selected="true"]` (`app.css:4712-4716`) without extra rules. Worth confirming in both themes.

## Acceptance criteria

- [ ] Both tabs render a leading icon matching the design
- [ ] Icon and label are vertically centered and evenly gapped, consistent with `.card-head h3`
- [ ] Icon color tracks the selected/unselected tab states in both light and dark themes
- [ ] Any new SVGs added under `crates/ui/templates/icons/` follow the existing 16×16 / `currentColor` convention
- [ ] Tab labels stay in the Fluent catalogs (`batch-tab-actions`, `batch-tab-json` at `locales/en/main.ftl:548-549`) — the icon is decorative markup, not part of the translated string
- [ ] `role="tab"` / `aria-selected` semantics unchanged; the icon does not become an accessible name

## Scope / out of scope

Out of scope: the history page tab row (`crates/ui/templates/pages/history.html:35-37`), which uses a different `.tab` class and is also icon-less. If the design calls for icons there too, that is a separate issue.

### LINKED ISSUE #729: ui: show the total created count in the Batch/Transaction response card header, left of the status badge
## Summary

Move the total created count from the summary line below the results card header up into the header itself, immediately to the left of the response status badge in the upper-right corner.

## Current behavior

Execute a batch or transaction bundle on `/ui/batch`. The response stage renders the HTTP status badge in the card head's upper right, and the aggregate counts as a separate muted paragraph underneath it:

```html
<!-- crates/ui/templates/pages/batch.html:92-99 -->
<div class="card-head">
  <h3><span class="icon">{% include "icons/check.svg" %}</span> {{ i18n.t("batch-response-heading") }}</h3>
  <span class="batch-badge" id="batch-overall"></span>
</div>
<p class="batch-summary" id="batch-summary"></p>
```

`#batch-overall` gets the numeric status (`batch.js:363`); `#batch-summary` gets a dot-joined tally built at `batch.js:419-424`:

```js
var parts = [];
if (created) parts.push(created + " " + messages.msgCreated);
if (updated) parts.push(updated + " " + messages.msgUpdated);
if (other) parts.push(other + " " + messages.msgOther);
if (failed) parts.push(failed + " " + messages.msgFailed);
summary.textContent = parts.join(" · ");
```

So the single number a user most wants after an import — how many resources were created — is a muted secondary line, spatially separated from the success indicator.

## Expected behavior

The created count reads in the card head, directly left of the status badge, so the header carries the outcome at a glance: `Per-Action Outcomes` on the left, `9 created` `200` on the right.

## Open decision

What happens to the rest of the tally (`updated`, `read/other`, `failed`). Two defensible readings, and this issue should settle one rather than leave it to the implementer:

- Move only `created` into the header and leave `#batch-summary` below with the remaining counts.
- Move the whole tally into the header and drop `#batch-summary`.

The stated ask is the created count specifically, so the first is the default reading. Note that `failed` is the count a user must not miss — if the summary paragraph is kept, it must stay visible even when `created` is the only thing in the header.

## Suggested approach

Add a sibling to `#batch-overall` inside the card head and populate it from the `created` counter already computed at `batch.js:407`. Reuse the `batch-sum-created` catalog key (`locales/en/main.ftl:556`) so the label stays translated; add nothing new to the catalogs if the existing key suffices.

Layout caveat: `.card-head` is `display: flex; justify-content: space-between` (`crates/ui/assets/app.css:3615-3622`), and `.batch-badge` carries `margin-left: auto` (`app.css:4835`). With three children, `space-between` plus that auto margin will not place the count snug against the badge. Either wrap the count and badge in a small flex group, or give the count the `margin-left: auto` and drop it from the badge — but note `.batch-badge` is also used for the per-row status badges (`batch.js:400-404`), where `margin-left: auto` is what pushes them to the right of each row, so do not remove it from the shared class.

Hide the count element when `created` is zero rather than rendering `0 created`, matching how `batch.js:419-424` already omits zero-valued parts.

## Acceptance criteria

- [ ] Created count renders in the response card head, immediately left of the status badge
- [ ] Count is hidden, not zero-valued, when nothing was created
- [ ] Per-row badges keep their existing right alignment
- [ ] The `failed` count remains visible somewhere on the stage in every case
- [ ] Label stays translated via the Fluent catalogs; any new key added to all of `locales/{en,es,de}/main.ftl`
- [ ] Header lines up correctly in both light and dark themes, and with a long localized heading

## Scope / out of scope

Out of scope: changing how the counts are classified (`batch.js:406-410` decides created vs updated vs other from the status code and method). This issue is presentation only.

### LINKED ISSUE #730: ui: Batch/Transaction — drop the bottom Cancel/Execute footer and move the execute error above the Execution Plan card
## Summary

Delete the bottom `Cancel` / `Execute` footer on the Batch/Transaction preflight stage. The top footer becomes the only action row. Move the execute error out of that footer and render it as a full-width alert box directly above the `Execution Plan` card.

## Current behavior

Load `/ui/batch` and drop in a batch or transaction bundle. The preflight stage renders, top to bottom: an action footer, the request strip, the semantics notice, the `Execution Plan` card, then a second action footer that duplicates the first and also holds the error:

```html
<!-- crates/ui/templates/pages/batch.html:56-59 — top footer, kept -->
<div class="batch-footer batch-footer--top">
  <button type="button" class="btn" id="batch-cancel-top">{{ i18n.t("batch-cancel") }}</button>
  <button type="button" class="btn btn--primary" id="batch-execute-top">{{ i18n.t("batch-execute") }}</button>
</div>

<!-- crates/ui/templates/pages/batch.html:82-88 — bottom footer, to be removed -->
<div class="batch-footer">
  <button type="button" class="btn" id="batch-cancel">{{ i18n.t("batch-cancel") }}</button>
  <p class="alert alert--inline" id="batch-execute-error" role="alert" hidden></p>
  <button type="button" class="btn btn--primary" id="batch-execute">{{ i18n.t("batch-execute") }}</button>
</div>
```

The error only appears at the very bottom of the page. On a large bundle the plan card is long, so a failure raised by the top `Execute` button renders far below the fold, out of sight of the button that was pressed.

## Expected behavior

- The bottom `.batch-footer` is gone. `Cancel` and `Execute` exist once, at the top.
- `#batch-execute-error` renders as a full-width alert immediately above the `Execution Plan` card — after the semantics notice at `batch.html:68`, before the `<div class="card">` at `:70`.
- It stays hidden until an execute attempt fails, and is cleared on the next attempt, exactly as now.

## Suggested approach

Template (`crates/ui/templates/pages/batch.html`):

- Delete lines 81-88 (the comment and the bottom `.batch-footer`).
- Insert `<p class="alert" id="batch-execute-error" role="alert" hidden></p>` between the semantics notice and the plan card.
- Drop the `alert--inline` modifier. It exists to shrink the alert into a flex row (`display: inline-block`, no margin — `crates/ui/assets/app.css:1378-1382`); standing alone above a card, the base `.alert` block treatment with its `margin: 0 0 12px` (`app.css:1367-1376`) is what is wanted.

Script (`crates/ui/assets/batch.js`) — the removed IDs are referenced in four places and will be `null` if only the template changes:

- `:46-49` — `executeBtn` / `cancelBtn` lookups for `batch-execute` and `batch-cancel`
- `:317` — `cancelBtn.addEventListener("click", reset)`
- `:356` — `executeBtn.addEventListener("click", execute)`
- `:334` and `:353` — `[executeBtn, executeTopBtn]` and `alsoDisable: [cancelBtn, cancelTopBtn]`, the busy-state disable lists

Collapse each of these to the `-top` element alone. `#batch-execute-error` itself keeps its ID, so `executeError` at `:42` and its uses at `:323`, `:349-350`, and `:377-378` need no change.

Also check `reset()` at `crates/ui/assets/batch.js:305-315`, which restores focus by testing `stages.preflight.contains(active)` — removing the bottom buttons changes which elements can hold focus when Cancel is pressed, though the containment test should still hold.

## Relationship to prior issues

This intentionally narrows two earlier decisions; note them so the change is not read as a regression:

- #675 added the top footer specifically so a large bundle could not push the only actions below the fold. That goal is preserved — the surviving copy is the top one.
- #676 moved errors next to the button that raised them, which is why the error currently sits inside the bottom footer. With the bottom footer gone, above the plan card is the position that keeps the error near the top action row and in the reading path.

## Acceptance criteria

- [ ] Bottom `Cancel` / `Execute` removed; top footer is the only action row on the preflight stage
- [ ] Execute error renders above the `Execution Plan` card as a full-width `.alert`
- [ ] Error is hidden on load, shown on failure, and cleared when Execute is pressed again (`batch.js:323`)
- [ ] The whole-bundle failure path at `batch.js:373-379` — a rolled-back transaction returning an `OperationOutcome` — still surfaces its diagnostics in the relocated box
- [ ] No stale `getElementById` returning `null` for the removed IDs
- [ ] Busy-state disabling still covers both remaining buttons during execute
- [ ] `role="alert"` retained so the failure is announced

## Scope / out of scope

Out of scope: the response stage footer (`batch.html:104-106`), which holds the single `Done` button and is unaffected. Also out of scope: the stage 1 upload error (`#batch-upload-error`, `batch.html:44`), which already sits directly under the drop zone.

### LINKED ISSUE #731: ui: Batch/Transaction reset() does not clear the error boxes, so a stale error is still visible after Cancel
## Summary

An error raised on the Batch/Transaction page survives `Cancel` and is still on screen when the user comes back to start a new import. `reset()` clears the bundle, the file input, and the preflight rows, but never clears either of the page's two error elements.

## Current behavior

1. Go to `/ui/batch` and attempt an import that fails.
2. Press `Cancel`. The page returns to the upload stage.
3. Click the `Batch/Transaction` nav item to start a new import.
4. The error from the previous attempt is still displayed.

## Expected behavior

`Cancel` returns the page to a clean upload stage. No error text from the abandoned attempt remains visible, and re-entering the page never shows a stale error.

## Evidence

The page has two error elements, and `reset()` clears neither:

```js
// crates/ui/assets/batch.js:299-315
function reset() {
  resetJsonRendering();
  bundle = null;
  fileInput.value = "";
  clearPreflight();
  show("upload");
  ...
}
```

```js
// crates/ui/assets/batch.js:134-138 — what clearPreflight() actually covers
function clearPreflight() {
  rows.textContent = "";
  rawJson.textContent = "";
  bundleJsonRenderer = null;
}
```

- `#batch-execute-error` (`crates/ui/templates/pages/batch.html:86`) is set at `batch.js:349-350` and again at `:377-378`, and is only ever hidden at the top of `execute()` (`:323`). Nothing in the cancel path hides it — `show("upload")` merely hides the whole preflight `<section>` around it, leaving `hidden = false` on the element itself. The moment the preflight stage is shown again, the old text is back.
- `#batch-upload-error` (`crates/ui/templates/pages/batch.html:44`) is set by `fail()` at `batch.js:129-132` and hidden only at `readFile()` (`:157`). `reset()` does not touch it, so an upload-stage error stays on screen through a cancel.

Both are the same class of bug as #682 ("Bulk Import dialog keeps typed values after cancel/close").

## Suggested approach

Hide both error elements in `reset()`:

```js
function reset() {
  resetJsonRendering();
  bundle = null;
  fileInput.value = "";
  uploadError.hidden = true;
  executeError.hidden = true;
  clearPreflight();
  show("upload");
  ...
}
```

Clearing `textContent` as well as setting `hidden` is worth doing — a hidden element still holds its text for assistive tech that walks the tree, and `role="alert"` on both makes a later unhide re-announce whatever is left in there.

One thing to confirm while fixing: step 3 of the repro is a full page navigation. `/ui/batch` in the sidebar is a plain `<a href>` (`crates/ui/templates/layouts/base.html:81`) with no `hx-boost` anywhere in the app, so the server re-renders the template with `hidden` on both error elements and a fresh load should not show a stale error. If the error does survive an actual reload, the mechanism is the back/forward cache restoring the live DOM — nothing under `crates/ui/assets/` handles `pageshow`/`event.persisted` today — and the fix needs a `pageshow` handler that re-runs the clear, not just the `reset()` change above. Determine which of the two is happening before closing this out; the `reset()` gap is real either way.

## Acceptance criteria

- [ ] `Cancel` from the preflight stage leaves no execute error visible
- [ ] `Cancel` leaves no upload error visible
- [ ] Returning to `/ui/batch` — by nav click and by browser back/forward — shows a clean upload stage
- [ ] The error elements are empty as well as hidden after a cancel, so `role="alert"` cannot re-announce stale text
- [ ] `execute()`'s existing clear at `batch.js:323` still works for the retry-after-failure path

## Scope / out of scope

Out of scope: changing what the error messages say, and the response stage — `Done` routes through the same `reset()` and is fixed by the same change.

### LINKED ISSUE #732: ui: Batch/Transaction scrolls to the Execute button when a large bundle is dropped — focus() moves the viewport
## Summary

Dropping a large bundle on the Batch/Transaction page scrolls the viewport down to the top of the `Execute` button as the preflight stage appears. Small bundles do not do this. The page should stay where it is.

## Current behavior

1. Go to `/ui/batch`.
2. Drop a bundle JSON with enough entries that the rendered execution plan is taller than the viewport.
3. As the preflight stage appears, the page scrolls; the top action row (`Cancel` / `Execute`) ends up pinned at the top of the viewport.

With a small bundle the page does not move.

## Expected behavior

Revealing the preflight stage does not move the viewport. The user stays where they were, at the top of the page, and reads down into the plan themselves.

## Evidence

The only viewport-moving call in this path is the focus handoff that runs right after the stage is revealed:

```js
// crates/ui/assets/batch.js:188-198
renderPreflight(generation);
show("preflight");
/* The drop zone was hidden with its stage, stranding focus. ... */
var active = document.activeElement;
if (active === document.body || stages.upload.contains(active)) {
  stages.preflight.focus();
}
```

`HTMLElement.focus()` scrolls the focused element into view unless told not to. `#batch-preflight` (`crates/ui/templates/pages/batch.html:52`) begins with the top `Cancel`/`Execute` footer (`:56-59`), so scrolling that section into view lands exactly on the `Execute` button — which matches the reported symptom.

The size dependence follows from the same call. `.content` and `.pane` set no `overflow` (`crates/ui/assets/app.css:550-556`, `:645-648`), so the document is the scroller. With a small bundle the page has no scroll range at all, and nothing can move no matter what `focus()` requests. With a large bundle the rendered plan makes the document scrollable, and the focus scroll becomes visible.

The condition guarding the call is satisfied on both entry paths: a drag-and-drop leaves `document.activeElement === document.body`, and a click-to-browse pick leaves focus on the drop zone, which `stages.upload.contains(active)` matches.

## Suggested approach

Pass `preventScroll` so focus still moves for keyboard and screen-reader users without dragging the viewport:

```js
stages.preflight.focus({ preventScroll: true });
```

The focus handoff itself must stay. It exists because the drop zone is hidden with its stage, which would otherwise strand focus on `<body>` (#679), and the comment at `batch.js:192-196` records why the target is the stage rather than its primary action.

Two things to check while fixing:

- `preventScroll` support is universal in current browsers but the option is silently ignored where it is not honored, so verify against the browsers the Playwright suite runs.
- The other two `focus()` calls in the file have the same exposure and are worth the same treatment if they move the page: `drop.focus()` in `reset()` (`batch.js:314`) and `doneBtn.focus()` after a run completes (`:429`). The response stage can also be long, so the `Done` case is the more likely of the two to be noticeable.

## Acceptance criteria

- [ ] Dropping a large bundle reveals the preflight stage with no viewport movement
- [ ] Focus still lands on `#batch-preflight` — verify with a keyboard-only pass that Tab continues from the revealed stage, not from the top of the document
- [ ] Screen-reader announcement of the revealed stage is unchanged
- [ ] Small-bundle behavior unregressed
- [ ] `drop.focus()` and `doneBtn.focus()` checked for the same jump and fixed if present

## Scope / out of scope

Out of scope: the focus-management design from #679 — where focus lands is correct and should not change, only whether the page scrolls to follow it. Also out of scope: the sticky busy region (`.batch-busy`, `crates/ui/assets/app.css:4814-4824`), which is positioned relative to the viewport by design.

angela-helios:
Closes #726. Closes #727. Closes #729. Closes #730. Closes #731. Closes #732.

One pass over the Batch/Transaction stage, issue by issue:

- **#730** — the bottom `Cancel`/`Execute` footer is gone; the top row is the only action row. `#batch-execute-error` moves to a full-width alert directly above the Execution Plan card (after the semantics notice), so a failed execute is visible next to the button that raised it however long the plan is. `batch.js` drops the second set of wiring.
- **#731** — `reset()` now hides `#batch-upload-error` and `#batch-execute-error`. Hiding a stage never cleared them, so an abandoned attempt's error was still on screen for the next one — same defect family as #682.
- **#732** — the preflight focus handoff passes `{ preventScroll: true }`: on a plan taller than the viewport, `focus()`'s scroll-into-view pinned the Execute row to the top. The handoff itself stays (it is #679's fix).
- **#726** — `#batch-preflight:focus { outline: none }`: the stage's programmatic focus fell through to the browser's default two-tone ring around the whole section. It is `tabindex="-1"`, not keyboard-reachable, so the ring was pure noise.
- **#729** — the created count renders in the response card head, immediately left of the status badge (`3 created`  `200`). Per the issue's default reading, `updated`/`other`/`failed` keep the summary line below, which now hides when it has nothing to say — so `failed` stays visible whenever present.
- **#727** — the Actions / Bundle JSON tabs get their leading icons (`sliders.svg`, `code.svg`, the issue's suggested candidates from the existing set), with the `.batch-tab` flex treatment the issue prescribes.

## Tests

The batch spec moves to the surviving `#batch-execute-top`/`#batch-cancel-top` ids and gains regressions: error placement above the plan card, exactly one footer, stale-error-after-cancel (#731), and a tall-bundle reveal that asserts `scrollY === 0` plus no focus outline (#726/#732). Full `helios-ui` Rust ring, batch + nojs + design-system + a11y (light/dark) Playwright projects green; verified visually against the running server.

### DIFF
--- a/crates/ui/assets/app.css
+++ b/crates/ui/assets/app.css
@@ -4700,6 +4700,9 @@ details.json-fold[open] > summary .icon {
 }
 
 .batch-tab {
+  display: inline-flex;
+  align-items: center;
+  gap: 8px;
   padding: 8px 14px;
   border: 1px solid var(--divider);
   border-radius: 999px;
@@ -4845,6 +4848,27 @@ details.json-fold[open] > summary .icon {
 .batch-summary {
   margin: 4px 0 10px;
   color: var(--muted);
+}
+
+/* The card head's right-hand cluster (#729): created count beside the
+   status badge. */
+.card-head__meta {
+  display: flex;
+  align-items: center;
+  gap: 10px;
+}
+
+.batch-created {
+  color: var(--muted);
+  font-size: 12px;
+  font-variant-numeric: tabular-nums;
+}
+
+/* The stage takes programmatic focus when revealed (#679); it is not
+   keyboard-reachable, so the browser's default ring around the whole
+   section is noise (#726). */
+#batch-preflight:focus {
+  outline: none;
 }
 
 
--- a/crates/ui/assets/batch.js
+++ b/crates/ui/assets/batch.js
@@ -43,10 +43,9 @@
   var outcomes = document.getElementById("batch-outcomes");
   var overall = document.getElementById("batch-overall");
   var summary = document.getElementById("batch-summary");
-  var executeBtn = document.getElementById("batch-execute");
   var executeTopBtn = document.getElementById("batch-execute-top");
-  var cancelBtn = document.getElementById("batch-cancel");
   var cancelTopBtn = document.getElementById("batch-cancel-top");
+  var createdBadge = document.getElementById("batch-created");
   var doneBtn = document.getElementById("batch-done");
   var busyRegion = document.getElementById("batch-busy");
 
@@ -194,7 +193,9 @@
              revealed stage, not its primary action. */
           var active = document.activeElement;
           if (active === document.body || stages.upload.contains(active)) {
-            stages.preflight.focus();
+            /* preventScroll (#732): on a plan taller than the viewport the
+               default scroll-into-view pinned the Execute row to the top. */
+            stages.preflight.focus({ preventScroll: true });
           }
         } finally {
           busy.done();
@@ -301,6 +302,11 @@
     bundle = null;
     fileInput.value = "";
     clearPreflight();
+    /* Errors are page-level elements, not stage content: hiding a stage
+       leaves them set, so an abandoned attempt's error would greet the next
+       one (#731). */
+    uploadError.hidden = true;
+    executeError.hidden = true;
     show("upload");
     /* Done/Cancel hid the stage that held focus; land on the one action
        the upload stage offers (#679). Containment, not body: the focus
@@ -314,7 +320,6 @@
       drop.focus();
     }
   }
-  cancelBtn.addEventListener("click", reset);
   cancelTopBtn.addEventListener("click", reset);
 
   /* ---- stage 3: execute and report ------------------------------------ */
@@ -331,7 +336,7 @@
        crashed the settling renderResponse. Busy holds until the outcome is
        rendered, not merely until response headers arrive. */
     hfsBusy.during(
-      [executeBtn, executeTopBtn],
+      [executeTopBtn],
       function () {
         return fetch("/", {
           method: "POST",
@@ -350,10 +355,9 @@
             executeError.hidden = false;
           });
       },
-      { alsoDisable: [cancelBtn, cancelTopBtn], region: busyRegion, label: messages.msgExecuting }
+      { alsoDisable: [cancelTopBtn], region: busyRegion, label: messages.msgExecuting }
     );
   }
-  executeBtn.addEventListener("click", execute);
   executeTopBtn.addEventListener("click", execute);
   /* The run already happened: Done lands back on a clean upload stage
      rather than offering to re-run a mutation that succeeded (#675). */
@@ -416,12 +420,16 @@
       outcomes.appendChild(li);
     });
 
+    /* The created count reads in the card head next to the status badge
+       (#729); the rest of the tally keeps the summary line, failures above
+       all. */
+    createdBadge.textContent = created ? created + " " + messages.msgCreated : "";
     var parts = [];
-    if (created) parts.push(created + " " + messages.msgCreated);
     if (updated) parts.push(updated + " " + messages.msgUpdated);
     if (other) parts.push(other + " " + messages.msgOther);
     if (failed) parts.push(failed + " " + messages.msgFailed);
     summary.textContent = parts.join(" · ");
+    summary.hidden = !parts.length;
 
     show("response");
     /* The disabled trigger was hidden with its stage; land on the one
--- a/crates/ui/e2e/tests/batch.spec.ts
+++ b/crates/ui/e2e/tests/batch.spec.ts
@@ -72,16 +72,19 @@ test("a transaction bundle uploads, previews, executes, and reports", async ({ p
   await expect(page.locator("#batch-json .json-view")).toBeVisible();
   await page.locator("#batch-tab-actions").click();
 
-  // The execute error slot lives inside the footer, next to its button (#676).
-  await expect(page.locator(".batch-footer #batch-execute-error")).toHaveCount(1);
+  // The execute error slot sits above the plan card (#730), and the bottom
+  // footer is gone — Cancel/Execute exist once, at the top.
+  await expect(page.locator("#batch-execute-error + .card")).toHaveCount(1);
+  await expect(page.locator("#batch-preflight .batch-footer")).toHaveCount(1);
 
   // Execute: outcomes per entry plus the aggregate summary.
-  await page.locator("#batch-execute").click();
+  await page.locator("#batch-execute-top").click();
   await expect(page.locator("#batch-response")).toBeVisible();
   await expect(page.locator("#batch-outcomes .batch-row")).toHaveCount(2);
   await expect(page.locator("#batch-outcomes .batch-badge").first()).toContainText("201");
-  await expect(page.locator("#batch-summary")).toContainText("2");
-  await expect(page.locator("#batch-summary")).toContainText(/created/i);
+  // The created count reads in the card head, beside the status badge (#729).
+  await expect(page.locator("#batch-created")).toContainText("2");
+  await expect(page.locator("#batch-created")).toContainText(/created/i);
   await expect(page.locator("#batch-overall")).toHaveClass(/--ok/);
 
   // Done is the one way out (#675): back to a clean upload stage. Done hid
@@ -213,7 +216,7 @@ test("an invalid replacement clears the old preview and cannot execute stale or 
 
   // Exercise the defensive guard even though the hidden control cannot be
   // reached by a user while the upload stage is visible.
-  await page.locator("#batch-execute").evaluate((button: HTMLButtonElement) => button.click());
+  await page.locator("#batch-execute-top").evaluate((button: HTMLButtonElement) => button.click());
   expect(executeRequests).toBe(0);
   await expect(page.locator("#batch-upload-error")).toBeVisible();
 });
@@ -320,7 +323,7 @@ test("a non-bundle file is rejected with a message, not a crash", async ({ page 
 // The transient states are made deterministically observable: FileReader
 // delivery and the execute POST are both parked behind manual releases.
 
-const FOOTER_CONTROLS = ["#batch-execute", "#batch-execute-top", "#batch-cancel", "#batch-cancel-top"];
+const FOOTER_CONTROLS = ["#batch-execute-top", "#batch-cancel-top"];
 
 test("picking a file shows the busy region before any file bytes arrive", async ({ page }) => {
   await page.goto("/ui/batch", { waitUntil: "networkidle" });
@@ -373,12 +376,12 @@ test("execute busies the whole footer, ignores re-entrant clicks, and lands focu
   await page.goto("/ui/batch", { waitUntil: "networkidle" });
   await page.locator("#batch-file").setInputFiles(bundleFile("batch"));
   await expect(page.locator("#batch-preflight")).toBeVisible();
-  await page.locator("#batch-execute").click();
+  await page.locator("#batch-execute-top").click();
 
   // Both Execute copies spin, and the Cancels go inert with them: a
   // mid-flight Cancel raced the settling response and crashed on the nulled
   // bundle before #679.
-  await expect(page.locator("#batch-execute")).toHaveAttribute("aria-busy", "true");
+  await expect(page.locator("#batch-execute-top")).toHaveAttribute("aria-busy", "true");
   await expect(page.locator("#batch-execute-top")).toHaveAttribute("aria-busy", "true");
   for (const control of FOOTER_CONTROLS) await expect(page.locator(control)).toBeDisabled();
   await expect(page.locator("#batch-busy")).toBeVisible();
@@ -387,7 +390,7 @@ test("execute busies the whole footer, ignores re-entrant clicks, and lands focu
   // The default-motion busy button: the label yields to the animated ring,
   // and the filled primary gets the explicit white ring (accent-on-accent
   // vanishes in dark theme).
-  const busyStyle = await page.locator("#batch-execute").evaluate((button) => ({
+  const busyStyle = await page.locator("#batch-execute-top").evaluate((button) => ({
     color: getComputedStyle(button).color,
     content: getComputedStyle(button, "::after").content,
     animation: getComputedStyle(button, "::after").animationName,
@@ -401,7 +404,7 @@ test("execute busies the whole footer, ignores re-entrant clicks, and lands focu
   // Re-entrant activation cannot double-POST: a real click on a disabled
   // button dispatches nothing, and a synthetic event that does reach the
   // handler is ignored by the busy guard.
-  await page.locator("#batch-execute").evaluate((button: HTMLButtonElement) => button.click());
+  await page.locator("#batch-execute-top").evaluate((button: HTMLButtonElement) => button.click());
   await page
     .locator("#batch-execute-top")
     .evaluate((button) => button.dispatchEvent(new MouseEvent("click", { bubbles: true })));
@@ -444,10 +447,10 @@ test("a whole-bundle failure clears the busy state and re-enables the footer", a
 
   await page.goto("/ui/batch", { waitUntil: "networkidle" });
   await page.locator("#batch-file").setInputFiles(bundleFile("batch"));
-  await page.locator("#batch-execute").click();
+  await page.locator("#batch-execute-top").click();
 
   // The busy state is genuinely entered before the failure lands…
-  await expect(page.locator("#batch-execute")).toHaveAttribute("aria-busy", "true");
+  await expect(page.locator("#batch-execute-top")).toHaveAttribute("aria-busy", "true");
   await expect(page.locator("#batch-busy")).toBeVisible();
   release();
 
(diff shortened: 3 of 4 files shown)


angela-helios:
Manually verified on a local build: tall-bundle drop holds the viewport (#732) with no stray focus ring (#726), single top footer with the failed-execute alert landing above the plan (#730), Cancel clears the stale error (#731), created count in the response head (#729), and the tab icons (#727). All good.

smunini (approved):

PR #747, opened 2026-08-27
perf(postgres): take crud and search to first in the published benchmark field, and cut import CPU 53%

mauripunzueta:
Nine rounds of Postgres write-path and search work, measured against the
HealthSamurai FHIR server performance benchmark. **crud and search now lead the
published field; import is second, behind Microsoft alone.**

## Where this lands

Throughput, RPS. HFS crud/search from run `33400765064` (the quietest host of
the campaign); import from run `33446308268`, which carried the round-9 import
work but landed on a badly contended host — so 8.4 is a floor, not a fair read.

| suite | **HFS** | aidbox | microsoft | medplum | hapi | rank |
|---|--:|--:|--:|--:|--:|:--|
| crud | **6,391** | 5,550 | 3,726 | 1,698 | 2,978 | **1st** |
| search | **5,896** | 3,562 | 3,632 | 2,369 | 1,040 | **1st** |
| import | **8.4** | 5.31 | **15.42** | 4.08 | 2.91 | 2nd |

Search p99 by published category (ms, lower is better) — at or better than the
best server in the field in 11 of 12 categories:

| category | **HFS** | aidbox | microsoft | medplum | hapi |
|---|--:|--:|--:|--:|--:|
| composite Observation | **15** | 44 | 15 | 65 | 198 |
| quantity Observation | **15** | 54 | 15 | 71 | 149 |
| token Encounter | **13** | 33 | 15 | 65 | 170 |
| token Observation | **13** | 25 | 15 | 79 | 176 |
| date Encounter | **13** | 26 | 15 | 68 | 169 |
| date Observation | **14** | 26 | 14 | 70 | 165 |
| date Patient | **14** | 28 | 14 | 76 | 228 |
| reference Encounter | **14** | 25 | 15 | 64 | 173 |
| reference Observation | **13** | 25 | 14 | 64 | 152 |
| string Organization | **13** | 24 | 14 | 67 | 119 |
| string Patient | **15** | 23 | 15 | 68 | 89 |
| reference MedicationRequest | 17 | 24 | **15** | 66 | 148 |

## ⚠ How to read a number from this branch

**The environment moves these results by more than any commit here does.** Our
Postgres runs on a *shared* 4-CPU/11 GB Docker host; the published field ran on
a dedicated 8 vCPU/30 GB database host. Six runs of one unchanged branch:

| run | host containers | pg ms/call | crud | search |
|---|--:|--:|--:|--:|
| 33400765064 | 4 | 1.92 | **6,391** | **5,896** |
| 33407314287 | 6 | 1.90 | 6,194 | 5,831 |
| 33416756530 | 4 | 2.89 | 5,993 | 5,482 |
| 33409768172 | 6 | 3.40 | 4,948 | 4,268 |
| 33412052977 | 13-28 | 5.45 | 5,096 | 4,817 |
| 33419509028 | 7 | 6.70 | 4,418 | 3,475 |

Same code, 1.90 → 6.70 ms/call of environment, crud 4,418 → 6,391. The workflow
now prints `pg ms/call` plus the Docker host's loadavg and container count in
the step summary and flags anything over 3 ms/call as not comparable
(`ec5a4ee0a`). **Read that line before reading a delta**; a same-agent pair is
not sufficient, because Docker host load is independent of which runner got the
job.

## What actually moved it

**Search.** Seek-only token indexes replacing a 2,283 MB scan-bound one (v30,
v31); prefix-range string search instead of a table scan (v30); reference-base
normalisation, deleting the `LIKE` that existed because it was missing (v32); a
trigram index for `:contains` (v33). A one-sided `value-quantity` `gt`/`ge` with
no true matches was planned off the table-wide histogram and walked all 512,311
entries of the parameter slice; an uncorrelated `EXISTS` over the same predicate
becomes an InitPlan and gates the scan off behind a One-Time Filter — that
shape's p99 improved 14.53x raw / 11.20x normalized. It is tautological given
the outer predicate, so no row can be added or removed.

**Write path.** Fewer index rows (composite rows no query can reach, contained
`_id` rows nothing can read, full-text content nobody searches, `_id`/
`_lastUpdated` rows that merely restate the resource); statements prepared once
rather than 11M times; the delete path's version read folded into its CTE; PUT
no longer reads the row twice and `If-Match` is atomic; a tombstone stores no
resource body and a vread of one is 410; autovacuum tuned to keep up with
`search_index` (v35).

**Import.** Profiling showed HFS, not Postgres, was the wall by 4.4x — 1,815 µs
of CPU per resource against Postgres's 409. Two fixes in the extractor:
`get_polymorphic_fields` scanned all 8,514 `FIELD_TYPES` entries on every member
access that misses (now `partition_point` + break), and `Type.member`
deep-copied the whole resource ~40x per resource (now borrowed through a `Cow`).
**Extraction 652.5 → 200.3 µs/resource (3.26x)**, with `extracted_values`
identical at 142,891 before and after. HFS CPU per resource 1,815 → 851 (-53%).

**Harness correctness.** `fhir-benchmark.yml` passed `-c max_wal_size=1GB` and
left `checkpoint_timeout` at Postgres's 5-minute default, so every run since the
workflow was written checkpointed several times more often than aidbox's did.
Now four inputs defaulting to upstream's exact 8GB/2GB/256MB/15min. Clean
same-agent A/B: crud +7.5%, import +9%, search +1.7% (noise).

**Also fixed, not a perf change:** `phonetic` was answering with the `name`
match it is not.

## Not yet established

- **`525f89ab4` (UUIDv7 for server-assigned ids) is unverified end to end.** An
  isolated `COPY` test says -36%, but local could not reproduce it (16 cores, so
  HFS is still the wall there) and CI has only ever seen it on contended hosts.
  **If a clean run shows import down, revert that commit alone.**
- **4GB vs 8GB `max_wal_size` was never A/B'd** — both 8GB runs were contended.
  With `shared_buffers = 2GB` (upstream had 10GB) there is a plausible mechanism
  by which 8GB is worse here and not upstream.

## Ceiling

crud and search are at the roof of this campaign; what remains there is the
host, not the code. Import is not: after the extractor fix Postgres becomes the
limiter at 629.8 s of exec on 4 cores against a 276 s wall, so **removing HFS
from the path entirely caps us at ~12.9** against Microsoft's 15.42. Closing the
rest needs Postgres-side row reduction, and the only lever that size is the
`combo-*` duplication — **which must not be taken**: those rows are exactly what
make composite Observation 15 ms against aidbox's 44. The one lever that helps
import *and* crud while touching no search plan is a narrow-key surrogate for
`idx_search_resource` (measured -64% write cost on that index); the full bigint
surrogate is blocked by a paging tie-break that loses rows, analysed and
deliberately not shipped.

## State of this branch

- **`main` is merged in and the branch is conflict-free.**
- Schema migrations **v18 → v36**. Both sides had independently invented v17 —
  main's adds the Bulk Submit store (#772), ours denormalizes the pagination
  sort key onto `search_index`. Main's is already shipped, so its number cannot
  move and ours shifted up one. Appending main's at the end instead was rejected:
  a database already at 17 would skip our `ALTER TABLE search_index ADD COLUMN
  last_updated` and every later index migration reading that column would fail.
- Verified against **PostgreSQL 18.6**: a fresh database migrates v1 → v36 with
  no errors and ends with `bulk_provider_submissions`, `search_index_layout` and
  `search_index.last_updated` all present — neither side's v17 work was lost.
  Re-running against the migrated database is a clean no-op.
- `cargo clippy --all-targets --all-features -- -D warnings`, `cargo fmt` and
  the canonical-precedence check are clean on the merged tree.

⚠ **Every measurement in this PR was taken on `801e3370e`, before the merge.**
Main moved 305 commits, including in the REST search path merged here, so the
numbers above describe the pre-merge tree and want re-confirming on the merged
one.


mauripunzueta:
Dispatched run [33128380492](https://github.com/HeliosSoftware/hfs/actions/runs/33128380492) against `281f5757c` — the consolidated branch, which is the first run to carry every round plus the contention probe.

It answers the two things the PR body flags as open:

- **Round 3's search question.** The previous round-3 run lost 29% search throughput while Postgres did less work and each query got faster. That run's plan capture was also invalidated by a bad merge (fixed in `27bcfdfc3`), so this is the first clean look.
- **Whether the numbers can be trusted at all.** `host-contention.txt` now records the Docker host's load average and container count at each suite boundary, so this run states its own conditions instead of needing an external poller.

New plan-capture sections in this run: `AM..` (the composite fast path — the SQL the server actually executes, which earlier sections did not model) and `BM..` (base `value-quantity`, the shape nothing modelled and the one that alternates between a 29 ms and a 400 ms plan).

smunini (approved):

PR #777, opened 2026-08-28
ui: unify data-table column alignment

### LINKED ISSUE #758: ui: Tenant Maintenance right-aligns the Resources column; it should be left-aligned
## Summary

On Tenant Maintenance (`/ui/tenants`), the **Resources** column is right-aligned. It should be left-aligned, header and cells together.

## Current behavior

The Resources header and every cell under it carry `class="col-num"`, which is right-aligned:

```css
.col-num {
  text-align: right;
  font-variant-numeric: tabular-nums;
}
```

So the counts sit hard against the Created column while the Tenant column beside them starts at the left, and the column reads as detached from its own header.

The column is also not purely numeric. Depending on row state that same `col-num` cell holds:

- a count such as `12`, `1.4K`, or `2.31M` (`crates/ui/src/tenants.rs:106-115`)
- a spinner plus "Provisioning" text while the tenant is being created (`crates/ui/templates/partials/tenant_rows.html:43-48`)
- a red "Failed" tag (`crates/ui/templates/partials/tenant_rows.html:52-54`)

Right alignment suits none of the latter two, and the column visibly jumps between shapes as the 3s poll swaps rows in.

## Expected behavior

The Resources column — `<th>` and all three `<td>` variants — is left-aligned, matching the Tenant and Created columns. Digits stay tabular so counts remain vertically comparable.

## Evidence

- `crates/ui/templates/partials/tenant_rows.html:19` — `<th class="col-num">{{ i18n.t("tenants-col-resources") }}</th>`
- `crates/ui/templates/partials/tenant_rows.html:43`, `:52`, `:69` — the three `<td class="col-num">` variants (provisioning, failed, count)
- `crates/ui/assets/app.css:1504-1507` — the `.col-num` rule

## Suggested approach

Do **not** change `.col-num` itself. It is shared with three other tables, all of which are genuinely numeric and want right alignment:

- `crates/ui/templates/pages/capability-statement.html:86-98` — search param / include / revinclude counts
- `crates/ui/templates/pages/subscriptions.html:67-100` — last 24h / sent / failures
- `crates/ui/templates/pages/index.html:197` — the dashboard's per-type counts

Editing the shared rule would silently re-align all of them.

Instead, scope the change to this table — either drop `col-num` from the four spots in `tenant_rows.html` in favor of a left-aligned class that keeps `font-variant-numeric: tabular-nums`, or add a page-scoped override. Whichever way, apply it to the `<th>` and all three `<td>` branches so the header and every row state stay in step.

## Scope / out of scope

- In scope: the Resources column on `/ui/tenants`.
- Out of scope: the Capability Statement, Subscriptions, and dashboard tables. If those should change too, that is a separate decision about `.col-num` as a whole.

Gordex2014:
Closes #758

## What changed

- Align ordinary and numeric data-table headers and cells to the left throughout the UI.
- Keep tabular figures for numeric columns.
- Keep action columns right-aligned, with matching header and cell alignment.
- Keep empty rows centered, including when an empty cell also has the action-column class.
- Document and test the shared alignment rules.

Tenant resource counts now line up with the `Resources` header. Capability tables, dashboard data, subscription metrics, and dynamic query results use the same rules without route-specific overrides. Diff and history grids are unchanged because they do not use the data-table component.

## Validation

- `cargo test -p helios-ui` with 230 passing tests
- `cargo build -p helios-hfs --features ui,subscriptions`
- Focused Chromium tests
- Full Playwright suite with 288 passing and 4 skipped tests
- Manual checks at desktop and compact sizes, plus dark theme and dynamic FHIR results

## Risk

The CSS changes only affect `.data-table` and its documented column classes. There are no backend, API, persistence, or dependency changes.

### DIFF
--- a/crates/ui/README.md
+++ b/crates/ui/README.md
@@ -242,7 +242,7 @@ These are the shared primitives. Before styling anything, reach for one; add to
 | `.kv-grid` | Responsive two-column key/value layout; collapses to one column on compact viewports. |
 | `.page-head`, `.page-head__title`, `.page-head__lede`, `.page-head--row` | Page heading block; the only `<h1>` treatment; `--row` puts an action on the right. |
 | `.back-link` | In-page return link with theme-safe normal, visited, hover, and focus states. |
-| `.table-wrap` > `.data-table`, `.data-table__empty`, `.table-foot` | The table, always in its scroll wrapper; empty-state row; footer with pagination. |
+| `.table-wrap` > `.data-table`, `.col-num`, `.col-actions`, `.data-table__empty`, `.table-foot` | The table, always in its scroll wrapper: ordinary headers and data align left; `.col-num` uses tabular figures without changing alignment; `.col-actions` aligns right; empty-state rows stay centered; the footer hosts pagination. |
 | `.empty-state` | The same centered, muted empty treatment for non-table content. |
 | `.field`, `.field__label`, `.field__input`, `.field__hint`, `.field__hint--error` | A labelled form field. |
 | `.addbox`, `.addbox--modal`, `.addbox__panel`, `.addbox__head`, `.addbox__x`, `.addbox__actions` | The `<details>` disclosure for create/add flows; `--modal` centers it as a dialog. |
--- a/crates/ui/assets/app.css
+++ b/crates/ui/assets/app.css
@@ -1502,11 +1502,10 @@ select.field__input {
 }
 
 .col-num {
-  text-align: right;
   font-variant-numeric: tabular-nums;
 }
 
-.col-actions {
+.data-table .col-actions {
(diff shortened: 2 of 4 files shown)


Gordex2014:
## Visual evidence

Tenant Maintenance at 1280 × 800 on the final branch.

- The `Resources` header and both resource counts are left-aligned.
- Numeric values keep tabular figures.
- The `Actions` header and cells remain right-aligned by design.

Manual QA also covered 390 × 844, Capability Statement numeric columns, the dashboard table, dark-theme Subscriptions, and dynamic FHIR results.

Tenant Maintenance with left-aligned resource counts

smunini (approved):

PR #778, opened 2026-08-28
fix(ui): standardize detail field spacing

### LINKED ISSUE #759: ui: Server summary fields on the Capability Statement page have no vertical spacing between them
## Summary

The Server Summary card on `/ui/capability-statement` stacks its seven labeled fields with **no vertical separation at all**. The gap inside a field (label to value, 5px) is larger than the gap between fields (0px), so the block reads as one continuous run of alternating text instead of seven distinct rows.

## Current behavior

Load `/ui/capability-statement`. The first card, "Server summary", renders seven `.detail__field` rows — Description, URL, FHIR version, Status, Kind, Date, Formats — directly inside a `.card__body`:

```html
<div class="card__body">
  <div class="detail__field"><span>DESCRIPTION</span><div>…</div></div>
  <div class="detail__field"><span>URL</span><div>…</div></div>
  …
</div>
```

`.detail__field` is a flex column whose `gap: 5px` separates its own label from its own value, and it carries **no margin of its own** (`crates/ui/assets/app.css:3143-3148`). Everywhere else that pattern is used, the *parent* supplies the separation:

- `.detail` (the sticky sidebar) — `gap: 12px` (`crates/ui/assets/app.css:2976-2985`)
- `.form-panel` — `gap: 12px` (`crates/ui/assets/app.css:2989-2994`)
- `.kv-grid` — `gap: 14px 18px` (`crates/ui/assets/app.css:3265-3270`)

`.card__body` supplies none. It is `padding: 14px` and nothing else — no `display: flex`, no `gap` (`crates/ui/assets/app.css:755-757`). So the rows butt directly against each other, and the next field's uppercase label sits flush under the previous field's value.

The single column is a second problem: seven short values (`4.0.1`, `active`, `instance`, a date) each take a full-width row, so the card is far taller than its content needs and the values float alone in a wide empty band.

## Expected behavior

The Server Summary fields are separated from one another by more space than separates a label from its own value, so each field reads as a unit. Short values sit two-up rather than each claiming a full-width row, with the long ones (Description, URL) spanning the full width.

## Evidence

- `crates/ui/templates/pages/capability-statement.html:16-24` — the seven `.detail__field` rows inside a bare `.card__body`
- `crates/ui/assets/app.css:3143-3148` — `.detail__field` has `gap: 5px` internally and no outer margin
- `crates/ui/assets/app.css:755-757` — `.card__body { padding: 14px; }`, no flex, no gap
- `crates/ui/assets/app.css:3265-3270` — `.kv-grid`, the container this pattern is normally given

Compare the two pages that render the same "box of labeled fields" correctly, both of which wrap the rows in `.kv-grid`:

- `crates/ui/templates/pages/compartments.html:51-54`
- `crates/ui/templates/pages/bulk-import-detail.html:38-41`

Server Summary is the odd one out.

## Suggested approach

Wrap the seven rows in the existing `.kv-grid` rather than inventing new spacing. That yields the 14px row / 18px column gap, a two-column layout that collapses to one column under 1100px (`crates/ui/assets/app.css:3353-3355`), and it aligns this card with Compartments and the Bulk Import detail page.

Then mark the two long fields `detail__field--wide` (`crates/ui/assets/app.css:3272-3274`, `grid-column: 1 / -1`) so Description and the implementation URL span both columns instead of being squeezed into a half-width cell and wrapping badly.

Note `.kv-grid` carries `margin-bottom: 14px`, which will double up with `.card__body`'s own 14px bottom padding. Check that in the browser and drop one of the two if the foot of the card ends up too deep.

**Do not fix this by adding `display: flex; gap` to `.card__body`.** It is shared by many cards whose bodies are forms, textareas, empty states, and prose (`crates/ui/templates/pages/sql-view-definitions.html`, `sql-library.html`, `sql-export.html`, and others), and turning it into a flex column would reflow all of them.

## Scope / out of scope

- In scope: the Server summary card on `/ui/capability-statement`.
- Out of scope: the other cards on that page (Interactions, Operations, Resources) and the `.detail__field`-in-`.card__body` instances on `crates/ui/templates/pages/sql-export.html:63-83` and `sql-files.html:23-27`, which have the same underlying cause. Worth fixing, but those pages are getting a design pass in #756 — leave them to it rather than spot-patching twice.

Gordex2014:
## Summary

- establish one parent-owned spacing contract for every visible `.detail__field` surface
- render Capability Statement metadata as a responsive two-column grid with wide Description and Base URL fields
- apply the shared stack/flush primitives to SQL Export, SQL Files, and Bulk Import without broad `.card__body` changes
- document the invariant and responsive composition with an ASCII representation
- add HTTP and Playwright regression coverage across all affected pages and 1440/1240/390px viewports

## Validation

- `cargo test -p helios-ui` — 230 passed
- affected Playwright ring — 26 passed
- code review: Standards and Spec axes clean after remediation
- manual HITL approved before PR creation

## Known pre-existing failure

The full Playwright run exposed `queries.spec.ts:45` (results pagination). It reproduces unchanged on the pinned base commit `b64503ff`, so it is not caused by this diff.

Closes #759


Gordex2014:
## Visual evidence

Before and after at 1440 x 900.

**Before**
capability-statement-1440x900-real

**After**
capability-statement-1440x900-real

smunini (approved):

PR #780, opened 2026-08-29
fix(hts): implement FHIR string search for terminology names and titles

### LINKED ISSUE #719: hts: name/title search on CodeSystem/ValueSet/ConceptMap ignores FHIR string-search spec, uses exact match instead of partial
### Summary
`GET /CodeSystem`, `GET /ValueSet`, and `GET /ConceptMap` filter on `name`/`title` using SQL exact equality. Both parameters are `string`-typed FHIR SearchParameters, which the spec requires to default to case-/accent-insensitive prefix matching — HTS implements neither the default behavior nor the `:contains`/`:exact` modifiers.

### Current Behavior
- **Exact match only:** all six backend search paths (`crates/hts/src/backends/{sqlite,postgres}/{code_system,value_set,concept_map}.rs`) build `name = ?` / `title = ?` WHERE clauses — no prefix, no substring, no case/accent folding.
- **Modifiers silently dropped:** `ResourceSearchQuery` (`crates/hts/src/types.rs`) is a typed struct with fixed field names. A request like `?name:contains=foo` doesn't deserialize into any field, so axum's `Query` extractor drops it without error — the request returns unfiltered results, not an error, with no indication the modifier was ignored.
- **Admin UI inherits it:** `crates/hts-ui`'s CodeSystem/ValueSet/ConceptMap browsers forward `?name=`/`?title=` verbatim, so an operator searching by partial name gets zero results unless they type the exact string.

### Root Cause
Per [FHIR R4 `string` search parameters](https://hl7.org/fhir/R4/search.html#string):

> "By default, a field matches a string query if the value of the field equals or **starts with** the supplied parameter value, after both have been normalized by case and combining characters." `:contains` matches anywhere; `:exact` requires exact case/accents.

`name` and `title` are registered as `type: string` on CodeSystem/ValueSet/ConceptMap — confirmed directly against the official HL7 R4 spec bundle vendored in this repo (`crates/fhir-gen/resources/R4/search-parameters.json`, `SearchParameter/conformance-name` and `SearchParameter/conformance-title`, both `"type": "string"`, both listing `CodeSystem`/`ValueSet`/`ConceptMap` in `base`). Same facts on the public spec pages: [codesystem.html#search](https://hl7.org/fhir/R4/codesystem.html#search), [valueset.html#search](https://hl7.org/fhir/R4/valueset.html#search), [conceptmap.html#search](https://hl7.org/fhir/R4/conceptmap.html#search). `url` (`uri`) and `status` (`token`) are different parameter types and correctly remain exact-match — this gap is specific to the two `string` params.

Previously flagged as a known limitation in `edson/docs/hts-ui-design.md` §7.2.1.0/§7.2.1.1, which documents the same code paths.

### Expected Behavior
- Unmodified `?name=`/`?title=` on the FHIR API performs case-/accent-insensitive prefix matching, per spec default.
- `:contains` and `:exact` modifiers are supported on `name`/`title`.
- An unsupported modifier returns a `400 OperationOutcome` instead of being silently ignored.
- `url`/`status` matching is unchanged.
- **In `crates/hts-ui`, the CodeSystem/ValueSet/ConceptMap browser search forms default `name`/`title` search to `:contains` semantics** (not bare prefix) — the UI should send `name:contains=`/`title:contains=` by default, so admins get substring search out of the box for the common "find by partial name" case.

### Scope
Touches both `crates/hts` (backend: default prefix match + `:contains`/`:exact` modifier support) and `crates/hts-ui` (the `:contains`-by-default UI behavior above is required; how it's wired — e.g. always appending `:contains`, vs. a future explicit mode toggle — is for the implementer to decide).

Base this on `feat/551-hts-ui` — `crates/hts-ui` doesn't exist on `main` until that branch merges, so there's no other base with the UI code to change. Cut a **new branch off of `feat/551-hts-ui`** rather than committing directly to it, to keep this work out of the #551 UI PR it's stacked on.

Gordex2014:
Closes #719.

### Summary

- implement case- and accent-insensitive FHIR prefix matching for bare `name` and `title` searches;
- support `:contains` and case-/accent-sensitive `:exact` modifiers;
- return `400 OperationOutcome` for unsupported modifiers on advertised terminology search parameters;
- ignore empty parameters before modifier/control validation, while preventing normalization-empty values from becoming match-all predicates;
- apply string filtering before pagination in SQLite and PostgreSQL, then hydrate only the selected resources in a consistent read snapshot;
- preserve exact matching and existing fast paths for `url`, `version`, and `status`;
- keep public HTS UI URLs on bare `name`/`title`, while sending `name:contains`/`title:contains` only on the internal FHIR request.

### Implementation notes

- Matching is centralized in a backend-neutral Rust module using Unicode decomposition, combining-mark removal, and Unicode case folding.
- `:exact` compares the decoded Unicode string without case or accent normalization.
- SQLite uses a deferred read transaction and chunks selected-ID hydration in batches of 900.
- PostgreSQL uses a repeatable-read transaction and preserves the authoritative table ID when hydrating JSON.
- Existing resource-shape behavior is unchanged: SQLite CodeSystem/ValueSet summary and unfiltered synthesis, SQLite ConceptMap full hydration, and PostgreSQL full hydration are preserved.
- FHIR's recommended punctuation and non-significant-whitespace normalization remains out of scope for this issue.

### Validation

- `cargo fmt --all --check`
- `cargo check -p helios-hts --no-default-features --features sqlite,R4`
- feature checks for SQLite with R4B, R5, and R6
- combined SQLite/PostgreSQL R4 check
- new SQLite HTTP/storage contract: 5/5
- existing SQLite CodeSystem/ValueSet/ConceptMap search tests: 16/16
- new PostgreSQL contract: 1/1
- existing PostgreSQL search tests: 9/9
- `cargo test -p helios-hts-ui --lib`: 74/74
- new Playwright title-substring cases: 3/3
- manual API and UI before/after QA on seeded SQLite

### Manual QA result

- bare prefix `name=example` now returns the four expected seeded CodeSystems;
- `title:contains=Anatomy Code` returns only `ExampleCodeSystem`;
- percent-encoded `name%3Acontains` is decoded and matched;
- unsupported `name:not` returns HTTP 400 with an `OperationOutcome`;
- an empty `name=` is ignored;
- exact `url + status` behavior remains unchanged;
- CodeSystem, ValueSet, and ConceptMap UI title substrings each return exactly the expected row while the browser URL retains the bare `title` key.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3472,6 +3472,7 @@ dependencies = [
  "tower-http",
  "tracing",
  "tracing-subscriber",
+ "unicode-normalization",
  "uuid",
  "zip 0.6.6",
 ]
--- a/crates/hts-ui/e2e/tests/code-systems.spec.ts
+++ b/crates/hts-ui/e2e/tests/code-systems.spec.ts
@@ -43,6 +43,32 @@ test.describe("HTS CodeSystem browser (§7.2)", () => {
     ).toBeVisible();
   });
 
+  test("Title substring keeps the browser URL bare and selects ExampleCodeSystem", async ({
+    page,
+  }) => {
+    await page.goto("/ui/hts/code-systems?title=Anatomy%20Code");
+
+    const browserUrl = new URL(page.url());
+    expect(browserUrl.searchParams.get("title")).toBe("Anatomy Code");
+    expect(browserUrl.searchParams.has("title:contains")).toBe(false);
+    await expect(page.locator("#filter-title")).toHaveValue("Anatomy Code");
+    await expect(
+      page.getByText("No CodeSystems match these filters."),
+    ).toHaveCount(0);
+
+    const rows = page.locator("table tbody tr");
+    await expect(rows).toHaveCount(1);
+    await expect(
+      rows.getByRole("link", { name: "ExampleCodeSystem", exact: true }),
+    ).toBeVisible();
+    await expect(
+      rows.getByRole("cell", {
+        name: "Example Anatomy Code System",
+        exact: true,
+      }),
+    ).toBeVisible();
+  });
+
   test("load-more appends the next page beforeend without replacing rows", async ({ page }) => {
     // The seed loader injects 34 CodeSystems total (ex-cs-1 + ex-cs-2..
     // ex-cs-31 + ex-cs-source + ex-cs-target + ex-cs-limbs). The default
--- a/crates/hts-ui/e2e/tests/concept-maps.spec.ts
+++ b/crates/hts-ui/e2e/tests/concept-maps.spec.ts
@@ -85,6 +85,32 @@ test.describe("HTS ConceptMap browser (§7.5)", () => {
     await expect(page.locator("#hts-cm-rows tbody > tr").first()).toBeVisible();
   });
 
+  test("Title substring keeps the browser URL bare and selects ExampleCM", async ({
+    page,
+  }) => {
+    await page.goto("/ui/hts/concept-maps?title=Source-to-Target");
+
+    const browserUrl = new URL(page.url());
+    expect(browserUrl.searchParams.get("title")).toBe("Source-to-Target");
+    expect(browserUrl.searchParams.has("title:contains")).toBe(false);
+    await expect(page.locator("#filter-title")).toHaveValue("Source-to-Target");
+    await expect(
+      page.getByText("No ConceptMaps match these filters."),
+    ).toHaveCount(0);
+
+    const rows = page.locator("table tbody tr");
+    await expect(rows).toHaveCount(1);
+    await expect(
+      rows.getByRole("link", { name: "ExampleCM", exact: true }),
+    ).toBeVisible();
+    await expect(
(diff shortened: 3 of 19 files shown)


Gordex2014:
## QA evidence

Before the fix, partial title searches returned no rows:

before-cs-empty
before-vs-empty
before-cm-empty

https://github.com/user-attachments/assets/c328624d-242c-4021-80c9-50b09c627f5b

After the fix, the attached recording verifies:

- CodeSystem `Anatomy Code` returns exactly `ExampleCodeSystem`.
- ValueSet `Limbs Value` returns exactly `ExampleLimbsVS`.
- ConceptMap `Source-to-Target` returns exactly `ExampleCM`.
- Public UI URLs keep the bare `title` parameter.

API validation also confirmed bare prefix matching, encoded `:contains`, `400 OperationOutcome` for unsupported modifiers, ignored empty values, and unchanged exact `url + status` behavior.

smunini (approved):

Issue #764, opened 2026-08-28
ui: a submission whose manifest the recipient fails to process stays "In Progress" forever — poll_status never sets status


smunini:
## Summary

A submission whose manifest the recipient cannot process stays "In Progress" forever. `poll_status()` never assigns `submission.status` in any branch, so once a submission reaches `in-progress` the only things that can move it are the user's own Abort and Complete buttons. A recipient reporting failure — by any mechanism — does not.

## Reproduction

1. `/ui/bulk-import` → New Submission → create one.
2. On the detail page, Add Manifest with a URL the recipient will accept at kick-off but fail to process (unreachable host, 404, malformed manifest body).
3. Submit the manifest. The recipient accepts the kick-off with a 2xx, so `crates/ui/src/bulk_import.rs:1110` sets `status = "in-progress"` and status polling starts.
4. Wait.

The submission row reads **In Progress** indefinitely. The run log records what happened; the status never changes, and the Abort/Complete buttons stay the only way out.

## Root cause

`poll_status()` (`crates/ui/src/bulk_import.rs:959-1019`) mutates `progress`, `result`, and `poll_url` — and never `status`. Every branch leaves the submission `in-progress`:

| Poll outcome | What happens | `status` after |
|---|---|---|
| `200` — recipient finished, manifest reports `error[]` entries | `result` records `outputs`/`errors`, `poll_url` cleared (`:986-1002`) | `in-progress` |
| `4xx`/`5xx` | logs "polling stopped", clears `poll_url` (`:1010-1017`) | `in-progress` |
| network error | early return, nothing recorded (`:968-972`) | `in-progress` |
| `429` | backs off (`:1004-1009`) | `in-progress` |

The 200 case is the important one. The recipient's completion manifest carries the failure, and the code **already counts it** — `let errors = manifest["error"].as_array().map(Vec::len).unwrap_or(0);` (`:988`) — stores it in `result.errors`, and renders it in the status card as an "Errors" figure. A submission can therefore display `Errors: 3` while its status reads In Progress.

The 4xx/5xx case is worse in one respect: `poll_url` is cleared but `completed_at` is never set, so `crates/ui/templates/partials/bulk_import_status.html` falls to its last branch and renders a bare `<div id="bulk-status"></div>`. The status card disappears entirely, and the row still says In Progress.

### Two related defects found while tracing

[...]

PR #781, opened 2026-08-29
feat(ui): one-shot bulk submissions

angela-helios:
One-shot Bulk Submit UI, per the mockups Steve approved: a submission carries exactly one manifest, and **Create & Submit** fires the `$bulk-submit` kick-off immediately.

## Create dialog

- Name, **Manifest URL**, and authentication (with Test authentication) up front.
- Submitter identity, output format, and file request headers fold behind an **Advanced options** disclosure (native `<details>` — works without JS).
- The **Submission ID input is gone** (always generated, unique per submitter) and the **FHIR base URL input is gone** (derived from the manifest URL's origin at kick-off, matching the SMART reference provider).

## Detail page

- The manifest URL joins the summary card; the manifest list, its sort menu, and **Add Manifest** are removed.
- Summary actions are **Edit** and **Delete**. **Abort** moves onto the in-progress status card, and the **Complete** button is gone entirely — the server controls terminal status from poll results (#764/#765 transitions, shipped in #772).
- Removed routes and handlers: per-manifest add/remove/submit/replace/abort, submission-level complete and submit-all, and the `?sort=` query. `replacesManifestUrl` support in `helios-rest` is untouched; only the UI affordance is gone.

## Tests

- `bulk_import_http.rs` reworked to the one-shot flow — 25 tests, including: create fires the kick-off and logs the outcome, advanced options ride the kick-off, `fhirBaseUrl` falls back to the manifest origin, abort is status-only, terminal submissions reject status changes without side effects, and the in-progress status card offers Abort while the finished card does not.
- Playwright: layout/theme/dialog specs updated; the a11y ring now audits the open Advanced fold; the no-JS ring proves the whole one-shot create works without JavaScript.
- Locale catalogs drop the 19 retired keys in en/es/de (parity kept).

Stacked on #772 (uses its tenant-scoped `BulkProviderStore`); GitHub will retarget to `main` when that merges.


angela-helios:
## Screenshots

**Create dialog** — name, manifest URL, auth; Advanced collapsed:

create dialog

**Advanced options open** — submitter identity, format, file request headers:

advanced open

**Detail after Create & Submit** — manifest URL in the summary, Edit/Delete only, Abort on the in-progress card, kick-off in the log:

detail

**List**:

list

Issue #782, opened 2026-08-29
subscriptions: per-subscription 24h delivery series for the Last 24 hrs sparkline


angela-helios:
The Subscriptions frame (405:2 in Brett's Figma) draws a per-subscription sparkline in the LAST 24 HRS column. `SubscriptionsSnapshot` (`crates/observability/src/subscriptions.rs`) carries only aggregate 24-hour counters (`delivered_24h`, `first_try_24h`, `failed_24h`) — there is no time series to draw, and #767 renders the real count instead of faking a curve (#555's rule).

To land the sparkline the engine needs a small rolling per-subscription series (e.g. hourly delivery buckets over the last 24h) surfaced through the snapshot:

- `SubscriptionRow` gains something like `delivered_hourly: [u64; 24]` maintained next to the existing rolling counters
- The UI renders it as an inline SVG polyline (no JS, no external libs), green for active, red for failing, dotted amber for idle — per the frame

Extending the snapshot contract is engine work, which #767 scoped out; this issue tracks it.

Issue #586, opened 2026-08-19
subscriptions: delivery instrumentation — per-window counts and first-try success rate


angela-helios:
The second half of #580. The operator page ships with the delivery-window figures deliberately absent because the engine records no delivery history: nothing counts outcomes over time, so the design's **DELIVERED IN 24 H** card (with its first-try success rate) and the table's **LAST 24 HRS** column cannot be filled honestly.

## Proposal

In-memory rolling buckets in the engine, matching the dashboard's `DashboardWindow::LastDay` shape (48 × 30-minute epoch-aligned buckets):

- Per subscription: delivery outcomes bucketed as success / retryable error / permanent error, plus first-try vs after-retry on success — recorded at the dispatch call sites the engine already runs through (`dispatch_with_retry`, the websocket broadcast, email, messaging).
- Exposed through the `helios-observability` subscriptions provider: per-row `sent_24h` / `fails_24h`, and tenant totals `delivered_24h` + first-try rate.
- UI: the fourth card becomes DELIVERED IN 24 H with the rate as its qualifier, and the table gains the LAST 24 HRS column — per the design.

In-memory means the window resets with the process, like the rest of the engine's volatile state — same trade-off, stated in the module docs. A persisted delivery log (durable across restarts, queryable per subscription) stays out of scope here; if operators need it, that is its own issue with cross-backend storage work.

## Acceptance criteria

- [ ] Delivery outcomes recorded at every channel's dispatch path, bucketed per subscription.
- [ ] Ring arithmetic covered by unit tests (rotation, epoch alignment, window boundaries, counter overflow).
- [ ] Provider carries the per-row and total figures; absent (`None`) when the engine is off — never zeroes posing as data.
- [ ] The card and column render real figures, localized en/es/de, covered in the Rust ring and the e2e specs.
- [ ] No measurable dispatch-path overhead (a couple of atomic bumps per delivery).

Related: #580 (the page), #555 (the honesty rule this follows).

PR #783, opened 2026-08-29
feat(ui): subscriptions nav always present; page built out to the frame

### LINKED ISSUE #767: ui: Subscriptions nav entry is hidden unless the engine is running, and the page needs building out to the design
## Summary

Two parts, both on the Subscriptions page:

1. The left-nav entry is hidden unless the subscriptions engine happens to be running, so on a default `hfs` build the page has no way in — even though the page already knows how to explain itself when the engine is off.
2. Build the page out to the design.

Design frame: <https://www.figma.com/design/CcLtq79cH2aHv4Ii9aNQTP/Untitled?node-id=405-283>

---

### Current behavior

`crates/ui/templates/layouts/base.html:156-162`:

```html
<!-- Visible only when the subscriptions engine is advertised (#580). -->
{% if status.subscriptions_enabled() %}
<a class="nav-item" href="/ui/subscriptions" ...>
{% endif %}
```

`subscriptions_enabled()` (`crates/ui/src/lib.rs:390-392`) returns `helios_observability::subscriptions::enabled()`, which is true only while a `SubscriptionsProvider` is registered (`crates/observability/src/subscriptions.rs:76-78`). Registration happens in exactly one place — `crates/rest/src/lib.rs:875-880` — behind two gates:

- the `subscriptions` cargo feature (`crates/rest/src/lib.rs:773`), and
- `HFS_SUBSCRIPTIONS_ENABLED`, which **defaults to false**: `.unwrap_or(false)` (`crates/rest/src/lib.rs:775-777`).

So on a default run the nav entry does not exist and the feature is invisible.

### Why this is the wrong gate

The page **already handles the engine being off**. `crates/ui/templates/pages/subscriptions.html:11-13`:

```html
{% if !available %}
<div class="card notice">{{ i18n.t("subs-unavailable") }}</div>
{% else %}
```

`available` is `snapshot.is_some()` (`crates/ui/src/subscriptions.rs:80`), and the module doc calls this "an explained unavailable state" (`:8`). Hiding the nav entry makes that state **unreachable through the UI** — you can only see it by typing the URL. The explanation exists and nobody can find it.

Subscriptions is also the only feature that hides itself this way. Bulk Import, Bulk Export, and Tenants all render an unconditional nav entry and degrade on the page — `crates/ui/templates/pages/bulk-export.html:12-13` is the same `{% if !available %}` shape with the same intent (`bulk-export-unavailable`). Their nav entries at `crates/ui/templates/layouts/base.html:86-92` carry no condition.

### Expected behavior

The Subscriptions entry is always present in the Tools section. With the engine off, the page explains that it is off and how to turn it on — naming `HFS_SUBSCRIPTIONS_ENABLED` and the `subscriptions` build feature, since "unavailable" alone does not tell an operator what to do.

Worth considering: whether the entry should render in a disabled/muted state rather than as a live link when the engine is off. That is a design call; the frame may already answer it.

---

### Current state

`crates/ui/templates/pages/subscriptions.html` (111 lines) renders four `stat` cards and one `data-table` of subscriptions — id/topic, channel/endpoint, a status tag, and three counters. It is entirely read-only: no create, edit, delete, retry, or drill-down. Sort is a plain-links menu (`:47-57`).

### Known gaps to check against the frame

- **Two stale comments claim data is missing that has since landed.** `crates/ui/src/subscriptions.rs:12` says "Deliberately absent: per-window delivery counts and success rates", and `crates/ui/templates/pages/subscriptions.html:15-16` says the delivery-window card "waits for engine instrumentation (#580)". Both are out of date — #586 landed that instrumentation, and `last24` (`crates/ui/src/subscriptions.rs:38`), `delivered` (`:50`), and `rate` (`:53`, computed at `:91-94`) are all wired. Remove the comments, and confirm the cards present the data the frame asks for.
- **No per-subscription detail view.** The table is terminal; there is nowhere to go for one subscription's history or errors.
- **No actions.** If the frame shows retry, pause, or delete affordances, none exist today, and each needs a REST path decided — `crates/rest/src/handlers/subscriptions.rs` is the surface.
- **Column alignment.** `subs-col-last24`, `-sent`, and `-fails` use `col-num` (`crates/ui/templates/pages/subscriptions.html:67-69`, `:97-99`), which is right-aligned. #758 and #760 are changing that shared class's usage on other pages; whatever convention those settle on should apply here rather than diverging.
- **Status tag colors** already use `tag--active` / `tag--error` / `tag--idle` / `tag--requested` / `tag--muted` (`:84-96`). Confirm the frame's palette matches before adding variants.

### Constraints

Server-rendered HTMX, no SPA framework, no bundler, vendored assets only, and a working no-JS path asserted by the `nojs` Playwright project — `crates/ui/README.md:15-45` and `:75-80`. Any new copy needs keys in all three catalogs under `locales/`; `crates/ui/src/i18n.rs:323` enforces parity.

## Acceptance criteria

- [ ] The Subscriptions nav entry is present regardless of whether the engine is running
- [ ] With the engine off, the page names `HFS_SUBSCRIPTIONS_ENABLED` and the `subscriptions` feature rather than saying only "unavailable"
- [ ] A decision is recorded on whether the nav entry renders disabled when the engine is off
- [ ] The page matches the linked frame; anything in the frame that needs engine or REST work not yet present is filed as a follow-up rather than faked
- [ ] The two stale "deliberately absent" comments are removed
- [ ] Column alignment follows whatever convention #758/#760 settle on
- [ ] New strings present in `locales/en`, `locales/es`, and `locales/de`
- [ ] `nojs` Playwright project passes for `/ui/subscriptions`

## Scope / out of scope

- In scope: the nav entry and the `/ui/subscriptions` page.
- Out of scope: changing the default of `HFS_SUBSCRIPTIONS_ENABLED`. Showing the nav entry does not mean starting the engine — that stays an operator decision.
- Out of scope: engine work behind any figure the frame asks for and the snapshot does not carry. `SubscriptionsSnapshot` (`crates/observability/src/subscriptions.rs`) is the contract; extending it is a separate change.

angela-helios:
Closes #767.

## Nav entry

Always present in Tools, like Bulk Import / Bulk Export / Tenants — the conditional `subscriptions_enabled()` gate and its `Status` plumbing are removed. **Decision recorded:** the entry stays a live link with the engine off rather than rendering disabled — that matches how every other feature page degrades, and the frame shows no disabled nav treatment. With the engine off, the page now names `HFS_SUBSCRIPTIONS_ENABLED=true` and `--features subscriptions` instead of a bare "unavailable".

## Page vs the frame (405:2)

- Cards reordered to **Active / Failing / Idle / Delivered in 24 h**, with status dots on the first three labels and the frame's sub-lines (`delivering`, `Needs attention`, `No clients`, `{rate}% first try`).
- The STATUS column renders **dots with plain labels** (Active / Error / 0 clients / Requested / Off) instead of pill tags; the publication-status block reclaims `.tag--active` now that the subscription pills are gone.
- A failing row carries the frame's **red left edge** (`row--alert`).
- The sort control gets the SORT label / value / chevron treatment; options stay the honest three (Status / Most sent / Fail streak) since the snapshot has no recency field.
- Numeric columns inherit #758's left alignment via the shared `col-num`.
- The two stale "deliberately absent" comments are removed — #586 landed the 24-hour instrumentation and the cards/columns show it.

**Not faked:** the frame's per-row 24-hour sparkline needs a per-subscription time series `SubscriptionsSnapshot` does not carry. Filed as #782; the column shows the real count until the engine grows the series.

## Tests

- `subscriptions_http.rs` reworked: the off state asserts the switch and build feature are named and the nav entry is present either way; the live phase asserts dots, the red row edge, and that the pills are gone.
- Playwright `subscriptions.spec.ts` restructured: the nav test never skips; a new off-state test covers the explained copy (runs on default local builds); the live-state tests keep their engine-on skip for CI.
- New strings landed in `en`/`es`/`de` (parity enforced); a11y, design-system, no-cdn, and nojs rings green.

### DIFF
--- a/crates/ui/assets/app.css
+++ b/crates/ui/assets/app.css
@@ -2846,12 +2846,12 @@ details.json-fold[open] > summary .icon {
 
 /* FHIR publication-status tags (HTS browsers + resource detail headers).
    Distinct from the .tag--stored/config/conflict family: those describe
-   configuration/lint outcomes, these describe a resource's own status.
+   configuration/lint outcomes, these describe a resource's own status. */
+.tag--active {
+  background: var(--ok-soft);
+  color: var(--ok);
+}
 
-   `active` is deliberately absent here: the subscription-state block below
-   already defines `.tag--active` with the identical treatment, and defining
-   it twice is what `design-system.spec.ts` forbids. FHIR's `active` status
-   and a live subscription happen to want the same pill. */
 .tag--draft {
   background: var(--warn-soft);
   color: var(--warn);
@@ -2868,22 +2868,38 @@ details.json-fold[open] > summary .icon {
   border: 1px solid var(--surface-border);
   color: var(--muted);
 }
-/* Subscription states (#580): the cards and the STATUS column share this
-   vocabulary — active delivers, idle has no listeners, error needs attention. */
-.tag--active {
-  background: var(--ok-soft);
-  color: var(--ok);
+
+/* Subscription states (#580, reshaped to the frame in #767): the cards and
+   the STATUS column share a dot vocabulary — green delivers, amber has no
+   listeners, red needs attention. */
+.status-dot {
+  display: inline-block;
+  width: 8px;
+  height: 8px;
+  margin-right: 7px;
+  border-radius: 50%;
+  vertical-align: baseline;
 }
 
-.tag--idle,
-.tag--requested {
-  background: var(--warn-soft);
-  color: var(--warn);
+.status-dot--ok {
+  background: var(--ok);
 }
 
-.tag--error {
-  background: var(--danger-soft);
-  color: var(--danger);
+.status-dot--danger {
+  background: var(--danger);
+}
+
+.status-dot--warn {
+  background: var(--warn);
+}
+
+.status-dot--muted {
+  background: var(--muted);
+}
+
+/* A failing row carries the frame's red left edge. */
+.data-table tr.row--alert > td:first-child {
+  box-shadow: inset 3px 0 0 var(--danger);
 }
 
 /* Bulk Export job states (#537) — the same status-pill vocabulary. */
@@ -2913,6 +2929,21 @@ details.json-fold[open] > summary .icon {
 
 .subs-sort summary.btn::-webkit-details-marker {
   display: none;
+}
+
+.subs-sort__label {
+  color: var(--muted);
+  font-size: 9px;
+  letter-spacing: 0.05em;
+  text-transform: uppercase;
+}
+
+.subs-sort__chevron {
+  transition: transform 120ms ease;
+}
+
+.subs-sort[open] .subs-sort__chevron {
+  transform: rotate(180deg);
 }
 }
 
--- a/crates/ui/e2e/tests/subscriptions.spec.ts
+++ b/crates/ui/e2e/tests/subscriptions.spec.ts
@@ -7,30 +7,39 @@ import { test, expect } from "../pages/fixtures";
 // with an injected provider (crates/ui/tests/subscriptions_http.rs).
 //
 // CI builds hfs with the `subscriptions` feature; a local binary built with
-// plain defaults compiles the engine out, so these skip on the unavailable
-// state rather than failing the local run.
+// plain defaults compiles the engine out, so the live-state tests skip on the
+// unavailable state rather than failing the local run. The nav entry and the
+// explained off state are engine-independent (#767) and never skip.
 test.beforeEach(async ({ page }) => {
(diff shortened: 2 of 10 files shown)


angela-helios:
## Screenshots

**Engine off** — nav entry present, the page names the switch and build feature:

engine off

**Engine on** — cards per the frame (dots, order, sub-lines), dot-styled status column ready; populated rows are covered by the Rust ring with an injected provider, and the 24 h sparkline is tracked in #782:

engine on

smunini (approved):

PR #809, opened 2026-08-31
fix(ui,rest,persistence): bulk-import status refresh under a live ingest

### LINKED ISSUE #765: ui: a successful bulk import stays "In Progress" — poll_status never sets status=completed on a clean 200
## Summary

A bulk import that finishes cleanly still reads **In Progress**. On the simplest possible case — one patient, one manifest, recipient returns a `200` status manifest with zero errors — the status card renders "processing finished" with an output count, while the submission's status stays `in-progress` and the list row keeps saying In Progress. The only way to reach Completed is for the user to press the Complete button by hand.

Whether that manual step is correct is the open question this issue also needs to answer: see [Open question](#open-question) below.

## Current behavior

1. `/ui/bulk-import` → New Submission.
2. Add one manifest pointing at a small export (a single Synthea patient is enough).
3. Submit. The recipient accepts the kick-off with a 2xx, `crates/ui/src/bulk_import.rs:1110` sets `status = "in-progress"`, and status polling starts.
4. The recipient finishes ingesting and its poll URL returns `200` with a status manifest — `output[]` populated, `error[]` empty.

Result: the status card shows `Result: processing finished <timestamp>`, `Outputs: N`, `Errors: 0`, and the submission simultaneously reads **In Progress**. It stays that way indefinitely unless a user clicks Complete.

## Expected behavior

A submission whose recipient reported successful completion should not read In Progress. Either the UI transitions it to Completed itself, or — if the sender is genuinely required to signal completion explicitly — the UI should stop presenting the submission as still processing and should make the required Complete action obvious rather than leaving it as an ungated button.

## Evidence

- `crates/ui/src/bulk_import.rs:987-1002` — the `200` branch of `poll_status()` writes `result` (`completedAt`, `outputs`, `errors`), clears `progress` (`:996`) and `poll_url` (`:997`), and logs "processing finished". It never assigns `submission.status`.
- `crates/ui/src/bulk_import.rs:302` — `is_terminal()` is `matches!(status, "completed" | "stopped")`, so nothing about a finished poll makes the submission terminal.
- `crates/ui/src/bulk_import.rs:1167` — `completed` status is reachable only through `POST /ui/bulk-import/{id}/complete`, a manual button.
- `crates/ui/src/bulk_import.rs:525` — `can_finalize: s.status == "in-progress"`, and `set_status` refuses anything else (`:1187`). So the manual press is the sole transition.
- `crates/ui/templates/partials/bulk_import_status.html:14` — the result card renders off `completed_at`, which is independent of `status`. This is why the page can show a finished result and an In Progress status at the same time.

## Open question

The kick-off vocabulary (`.claude/skills/bulk-data-submit/SKILL.md:28`) is `in-progress` (default) / `completed` / `stopped`, sent as `submissionStatus` on a status-only kick-off (`kickoff_parameters`, `crates/ui/src/bulk_import.rs:686`). A submission is a container that can accept more than one manifest, so `completed` plausibly means "no further manifests are coming" — a deliberate sender decision, not something derivable from one manifest finishing. If so, the current code is spec-correct and the defect is purely presentational.

That needs to be settled before the fix is written. Reference thread: https://chat.fhir.org/#narrow/channel/179250-bulk-data/topic/STU4.20in.20CI.20build/with/613246799

Decide explicitly:

- Does the Data Provider signal `completed`, or may a Data Consumer's completion manifest imply it?
- If a submission may carry multiple manifests, what should the UI show between "all submitted manifests finished" and "the sender declared the submission complete"? That state currently has no vocabulary.
- Does auto-completing after a single manifest foreclose adding a second manifest to the same submission? `add_manifest` rejects terminal submissions (`crates/ui/src/bulk_import.rs:646`), so auto-completion is not a free change.

## Suggested approach

Contingent on the answer above. If completion stays a deliberate sender action, the smallest honest fix is a distinct display state for "manifests finished, submission not yet declared complete" so the row stops claiming work is ongoing, plus surfacing Complete as the obvious next step once polling has finished. If the transition can be inferred, assign `submission.status` in the `200` branch of `poll_status()` and confirm `add_manifest`'s terminal guard is still the behavior you want.

Either way the vocabulary comment at `crates/ui/src/bulk_import.rs:106` needs updating, and any new status label needs `en`/`es`/`de` entries — `crates/ui/src/i18n.rs` enforces parity.

## Acceptance criteria

- [ ] The intended sender-side completion flow is written down (in the book or the bulk-data-submit skill), citing the spec or the Zulip thread
- [ ] A clean single-manifest import no longer reads In Progress after the recipient reports completion
- [ ] The status card and the list row cannot disagree — no "processing finished" alongside In Progress
- [ ] Whatever transition is chosen has test coverage for the clean-success path

## Scope / out of scope

- In scope: the submitting client's status tracking, `crates/ui/src/bulk_import.rs`, and documenting the intended flow.
- Out of scope: the Data Consumer ingestion engine. Ingestion itself works; this is about how the sender's UI reports it.

## Related

#764 covers the same function for the failure paths (`4xx`/`5xx`, network errors, and `200` with a non-empty `error[]`). This issue is the clean-success path and the flow question, which #764's acceptance criteria do not cover. The two will likely be fixed together in `poll_status()` — coordinate so the status vocabulary lands once.

### LINKED ISSUE #788: ui: rename the Create Bulk Submission dialog's "Create & Submit" button to "Submit"
## Summary

The primary button in the "Create Bulk Submission" dialog reads "Create & Submit". It should read "Submit".

## Current behavior

`/ui/bulk-import` → "New Submission" opens the "Create Bulk Submission" dialog. Its primary action is labelled "Create & Submit".

## Expected behavior

The button reads "Submit".

## Evidence

- `crates/ui/templates/pages/bulk-import.html:45-48` — the dialog's action row:

```html
<div class="addbox__actions">
  <button type="button" class="btn" data-addbox-close>{{ i18n.t("ui-cancel") }}</button>
  <button type="submit" class="btn btn--primary">
    {{ i18n.t("bulk-import-create-submit") }}
  </button>
</div>
```

- `locales/en/main.ftl:591` — `bulk-import-create-submit = Create & Submit`
- `crates/ui/templates/pages/bulk-import.html:47` is the only use of that message id.

## Suggested approach

The message id is used in exactly one place, so the narrowest change is to update the value rather than introduce a new key:

- `locales/en/main.ftl:591` — `bulk-import-create-submit = Submit`
- `locales/es/main.ftl:588` — currently `Crear y enviar`; the matching one-word form is `Enviar`
- `locales/de/main.ftl:588` — currently `Erstellen & Senden`; the matching form is `Senden`

If a contributor prefers the key name to match the new label, renaming it to `bulk-import-submit` touches the same three locale files plus the one template line.

One test asserts the current text and must be updated in the same change:

- `crates/ui/e2e/tests/nojs/progressive-enhancement.spec.ts:145` — `await create.getByRole("button", { name: "Create & Submit" }).click();`

## Notes

The handler behind the button does create *and* kick off the submission — `crates/ui/src/bulk_import.rs:364` documents it as "one-shot create: a submission carries exactly one manifest, and creating it fires the kick-off immediately", and `crates/ui/src/bulk_import.rs:412` calls `submit_one_with_id` before saving. The rename is a labelling change only; no handler behaviour changes.

## Out of scope

The "Edit Submission" dialog and the submission detail page actions keep their current labels.

### LINKED ISSUE #789: ui: Processing card on the submission detail page has 10px of extra space below the Abort button
## Summary

On the Bulk Import submission detail page (reached from All Submissions), the Processing card carries 10px of extra whitespace below the Abort button, so the card is bottom-heavy and does not match the summary card directly above it.

## Current behavior

Open `/ui/bulk-import`, click a submission that is still in progress. The Processing card renders with an Abort button, and the space below the button is 28px while the space above the "PROCESSING" label is 18px:

```
┌──────────────────────────────┐
│  PROCESSING                  │  ← 18px card padding
│  Waiting for the recipient…  │
│         ↕ 10px               │
│  [ ⏹ Abort ]                 │
│         ↕ 10px  ← extra      │
│         ↕ 18px card padding  │
└──────────────────────────────┘
```

## Expected behavior

The Processing card's bottom spacing matches its top — the same treatment the summary card above it already gets.

## Evidence

- `crates/ui/templates/partials/bulk_import_status.html:13` — the Abort row is a plain `.detail__actions`:

```html
{% if can_abort %}
<div class="detail__actions">
  <form method="post" action="/ui/bulk-import/{{ id }}/abort">
    <button type="submit" class="btn">{% include "icons/stop-circle.svg" %}{{ i18n.t("bulk-import-abort") }}</button>
  </form>
</div>
{% endif %}
```

- `crates/ui/assets/app.css` — `.detail__actions { display: flex; gap: 8px; margin: 10px 0; }`. The `10px` bottom margin stacks on top of the card's padding.
- `crates/ui/assets/app.css:3012` — `.panel { padding: 18px }` via `.card panel bulk-import-section`, so the total below the button is 10 + 18 = 28px against 18px above the label.
- `crates/ui/assets/app.css:3023` — the summary card's action row already neutralizes this:

```css
.bulk-import-summary__actions {
  ...
  margin: 0;
}
```

  applied at `crates/ui/templates/pages/bulk-import-detail.html:28` (`class="detail__actions bulk-import-summary__actions"`). The Processing card's Abort row never got the equivalent.

- `crates/ui/assets/app.css` — `.kv-grid--flush { --kv-grid-margin-bottom: 0 }`, so the 10px above the Abort button comes entirely from `.detail__actions`' top margin; that gap is intentional and should stay.

## Suggested approach

Zero the bottom margin on the Abort row rather than changing `.detail__actions` itself — that class is also used by `crates/ui/templates/pages/search-parameters.html:165` and `crates/ui/templates/pages/compartments.html:71`, which are not part of this fix.

Either add a modifier class to `crates/ui/templates/partials/bulk_import_status.html:13` alongside `.detail__actions` and give it `margin-bottom: 0`, or scope a rule to `.bulk-import-section > .detail__actions:last-child`. The second also covers any future action row added to these cards.

## Scope

The Processing card only shows an Abort row while `can_abort` is true; the completed-result variant of the same partial has no action row and is unaffected.

### LINKED ISSUE #790: ui: Bulk Import polls the recipient every 5s against a 10-per-60s limit and ignores Retry-After, so progress never refreshes
## Summary

The Bulk Import status card polls the recipient every 5 seconds, but the `$bulk-submit` status endpoint allows 10 polls per 60s and asks clients to come back after 120s. The UI therefore spends almost all of its time throttled: the progress number never refreshes, and the submission run log fills with `Status poll throttled (429); backing off.` lines.

## Current behavior

`crates/ui/templates/partials/bulk_import_status.html:4` sets a fixed cadence:

```html
hx-trigger="every 5s"
```

That is 12 polls/minute. The recipient's default limit is 10 per 60s (`crates/rest/src/config.rs:688-689`), so roughly every sixth poll onward inside a window is rejected with `429`. The recipient also returns `Retry-After: 120` on the `202` progress response (`retry_after_secs: 120`, `crates/rest/src/config.rs:683`) and a `Retry-After` delta on the `429` itself (`crates/rest/src/handlers/bulk_submit.rs:136-141`).

`poll_status` never reads either header — the only header it looks at is `x-progress` (`crates/ui/src/bulk_import.rs:877`). The `429` arm just appends a log line and returns, leaving the next poll to fire 5s later at the same cadence:

```rust
429 => {
    push_log(
        submission,
        "Status poll throttled (429); backing off.".to_string(),
    );
}
```
`crates/ui/src/bulk_import.rs:920-925` — the log message says "backing off", but nothing actually backs off.

Observed on a live ~19M-resource submission (24 NDJSON files). The run log after 7 minutes:

```
2026-08-31T13:31:35.652Z Status poll throttled (429); backing off.
2026-08-31T13:26:34.577Z Status poll throttled (429); backing off.
2026-08-31T13:26:29.531Z Status poll throttled (429); backing off.
2026-08-31T13:24:39.559Z Status: processing 0% complete
2026-08-31T13:24:34.446Z Bulk status kick-off request
```

One successful progress read in the whole run. Polling the recipient directly confirms both sides:

```
$ curl -i http://localhost:8080/bulk-submit-status/<token>
HTTP/1.1 429 Too Many Requests
retry-after: 1
{"resourceType":"OperationOutcome","issue":[{"severity":"error","code":"throttled",
 "details":{"text":"too many status polls (max 10 per 60s); honour the Retry-After header"}}]}

# after waiting out the window
HTTP/1.1 202 Accepted
x-progress: processing 0% complete
retry-after: 120
```

## Expected behavior

The UI honors the recipient's `Retry-After` and stops issuing polls it knows will be rejected. A throttled poll should not be logged as a run-log event at all, or at most once per throttling episode — it is transport noise, not submission history.

## Evidence

- `crates/ui/templates/partials/bulk_import_status.html:4` — hardcoded `hx-trigger="every 5s"`
- `crates/ui/src/bulk_import.rs:858-947` — `poll_status`; reads `x-progress` only, ignores `Retry-After`
- `crates/ui/src/bulk_import.rs:920-925` — the `429` arm, which logs "backing off" without backing off
- `crates/rest/src/handlers/bulk_submit.rs:112-141` — `check_poll_rate_limit`, returns the delta-seconds
- `crates/rest/src/config.rs:683,688-689` — defaults: `retry_after_secs: 120`, `poll_rate_limit: 10`, `poll_rate_window_secs: 60`

## Suggested approach

Record a "not before" instant on the `Submission` from whichever `Retry-After` the recipient last sent (both the `202` and the `429` carry one), and have `status_fragment` skip the poll when the clock has not reached it — the fragment already gates on `poll_url` being non-empty (`crates/ui/src/bulk_import.rs:1090`), so this is the same shape of guard. Then drive the htmx cadence from that value rather than a literal, e.g. render `hx-trigger="every {{ poll_interval_secs }}s"` so the card re-renders promptly while polling the recipient only as often as it permits.

Suppressing the repeated throttle log line matters independently: every `push_log` grows the submission document that gets written back to storage on each poll, so the noise has a write cost as well as a readability cost.

## Scope / out of scope

The recipient side is behaving correctly — no change to the limiter or its defaults is wanted here. `crates/ui/templates/partials/bulk_export_card.html:7` uses the same 5s literal for Bulk Export; that path has no equivalent server-side limiter, so it is out of scope unless it turns out to be throttled too.

### LINKED ISSUE #791: sqlite: put_provider_submission uses a DEFERRED transaction, so every provider-submission write fails with "database is locked" during a bulk ingest
## Summary

`SqliteBackend::put_provider_submission` opens a DEFERRED transaction that reads before it writes. Under WAL, the read-to-write upgrade fails with `SQLITE_BUSY_SNAPSHOT` and the configured `busy_timeout` does not cover it, so every provider-submission write fails for as long as a bulk ingest is writing. The provider-side submission record then stops being persisted entirely.

## Current behavior

During a large `$bulk-submit` ingest the server log repeats every few seconds, indefinitely:

```
2026-08-31T13:31:20.493338Z  WARN helios_ui::bulk_import: failed to persist bulk-import submission
  submission=48d610f3-b51f-45e7-9956-be63fe965e51
  error=internal error in sqlite: write provider submission: database is locked
```

Every poll of the status card tries to save the updated submission (status, progress, run log) and every one of those saves is lost. The record on disk stays frozen at whatever was written before the ingest got busy, and `save_or_warn` only warns, so nothing surfaces in the UI.

The cause is the transaction mode. `put_provider_submission` does `SELECT version ...` and then `INSERT ... ON CONFLICT DO UPDATE` inside a transaction started with rusqlite's default:

```rust
let txn = conn
    .transaction()
    .map_err(|e| backend_err(format!("begin provider submission transaction: {e}")))?;
```
`crates/persistence/src/backends/sqlite/bulk_provider.rs:118-120`

`Connection::transaction()` is DEFERRED. The `SELECT` takes a WAL read snapshot; if the ingest worker commits before the `INSERT` runs, the upgrade returns `SQLITE_BUSY_SNAPSHOT`, and SQLite does not invoke the busy handler for that code — so `busy_timeout_ms` (default 30000, `crates/persistence/src/backends/sqlite/backend.rs:149`) never applies. With a worker committing continuously, the upgrade loses the race essentially every time.

This is already understood elsewhere in the same backend. `storage.rs` documents it and takes the write lock up front:

```rust
// IMMEDIATE, not the DEFERRED default: this transaction reads (the count
// below) before it writes. Under WAL a deferred transaction takes a read
// snapshot on that first read and then fails the read-to-write upgrade
// with SQLITE_BUSY_SNAPSHOT if another connection committed in between —
// and the busy handler is *not* invoked for that code, so the configured
// busy_timeout does not cover it. Taking the write lock up front does.
```
`crates/persistence/src/backends/sqlite/storage.rs:958-966`

`bulk_provider.rs` did not get the same treatment.

## Expected behavior

A provider-submission write succeeds while a bulk ingest is running, waiting out the writer backlog within `busy_timeout_ms` rather than failing immediately. The submission's persisted state should track its in-memory state.

## Evidence

- `crates/persistence/src/backends/sqlite/bulk_provider.rs:118-120` — DEFERRED `conn.transaction()`
- `crates/persistence/src/backends/sqlite/bulk_provider.rs:122-131` — the `SELECT version` that takes the read snapshot
- `crates/persistence/src/backends/sqlite/bulk_provider.rs:143-154` — the `INSERT ... ON CONFLICT` whose upgrade fails; source of the `write provider submission: {e}` message
- `crates/persistence/src/backends/sqlite/storage.rs:958-966` — the same hazard, diagnosed and fixed with `TransactionBehavior::Immediate`
- `crates/persistence/src/backends/sqlite/backend.rs:149` — `default_busy_timeout_ms() = 30000`, which this code path does not benefit from
- `crates/ui/src/bulk_import.rs:205-215` — `save_or_warn`, which swallows the failure into a `warn!`

Reproduced against a ~19M-resource, 24-file manifest on the SQLite backend; the recipient-side ingest itself was healthy throughout (`bulk_entry_results` growing at ~67 rows/sec, 0 failures) while every provider-submission write failed.

## Suggested approach

Switch to `conn.transaction_with_behavior(rusqlite::TransactionBehavior::Immediate)`, matching `storage.rs`. `crates/persistence/src/backends/sqlite/user_settings.rs:35` and `:215` use the plain `transaction()` default as well and are worth checking for the same read-then-write shape while in here.

Worth deciding separately whether `save_or_warn` should stay silent to the user when a save fails repeatedly — a submission whose state has not persisted for minutes is arguably worth surfacing, not just logging.

## Scope / out of scope

SQLite only. The Postgres, MongoDB, and S3 `bulk_provider` implementations do not have this failure mode.

angela-helios:
Closes #765, closes #790, closes #791 — plus the two cosmetic ones on the same page: closes #788, closes #789.

Steve's "still In Progress" was three faults stacked on top of each other:

## 1. Every provider-store write was lost during a live ingest (#791)

`SqliteBackend::put_provider_submission` opened a **DEFERRED** transaction that reads the version before writing. Under WAL the read→write upgrade fails with `SQLITE_BUSY_SNAPSHOT` — a code the busy handler is never invoked for, so `busy_timeout` doesn't apply — and with a bulk ingest committing continuously the race lost essentially every time. Status, progress, and the run log were computed and then silently dropped (`save_or_warn` only warns). Now **IMMEDIATE**, exactly like the documented precedent in `storage.rs`.

## 2. The UI burned the poll budget and ignored Retry-After (#790)

The card polled every 5s against the recipient's 10-per-60s limit. Now `Retry-After` (from the `202` and from the `429`) materializes as `nextPollAt` in the submission document: the htmx cadence keeps refreshing **this server's** view cheaply, and the recipient is contacted only when its window has passed. Throttled polls no longer spam the run log. The recipient's `X-Progress` also gains a number that actually moves — the percentage is manifest-granular (a one-shot submission reads 0% until the very end), so it now appends the summed processed-entries counter: `processing 0% complete (14,200 entries ingested)`.

## 3. Errors were read from the wrong manifest key

The STU4 status manifest lists OperationOutcome files under **`outcome`**; the UI read the bulk-export-style **`error`** key only, so a **truncated ingest reported "finished cleanly"**. Both keys count now, `outcome` entries weighted by their `countSeverity` (a warnings-only outcome file doesn't spoil a completion).

## Verification — 20,000-Patient one-shot run (as requested)

| | before | after |
|---|---|---|
| provider-store writes | lost every few seconds (`database is locked`) | **0 lost** across the whole run |
| run log | full of "throttled (429)" lines | clean |
| progress card | frozen at "processing 0%" | "(N entries ingested)" advancing on the 120s Retry-After cadence |
| final state | stuck In Progress | **Completed on its own**, 20,000/20,000 ingested |

Tests: the sqlite concurrency suite gains a provider-writes-under-ingest smoke; `bulk_import_http` gains Retry-After hold/backoff tests, the STU4 `outcome` vocabulary cases (severity-aware), and keeps the export-style `error` fallback covered — 28 tests green, plus the ui/rest/persistence suites.


angela-helios:
## Screenshots

**Create dialog** — the button reads Submit (#788):

dialog

**Processing card mid-run** — live '(N entries ingested)' counter (#790), Abort flush with the card padding (#789):

card

smunini (approved):

angela-helios:
Two follow-ups from Angela's review, pushed as 3f97c883c:

- The summary card's **STATUS cell now updates live**: the polling fragment carries the status label out-of-band (hx-swap-oob), so Completed/Failed lands without a page reload.
- The processing line is now a **progress bar** — determinate when the recipient reports a percentage, indeterminate sweep while only the entry counter moves (reduced-motion aware):

progress bar

Release-build throughput measured with the 20k run: **~43.5 resources/s** (10k ≈ 3.8 min). Filed #813 for the ingest batching work — the per-entry autocommit overhead is the bound, not this PR's scope.

angela-helios:
The two remaining states, captured from merged main (the follow-up commit 3f97c883c is part of this PR's merge):

**Completed** — the summary STATUS flips live via the out-of-band swap, no reload:

completed

**Determinate progress** — a submission with a completed manifest and one still streaming reads 50%, bar filled accordingly with the live entry counter below:

percent

PR #811, opened 2026-08-31
ui: fix Capability Statement version links and incremental JSON rendering

### LINKED ISSUE #797: ui: Capability Statement resource links resolve to the current FHIR release, not the server's version
## Summary

On the Capability Statement page, the per-resource links point at the unversioned canonical `http://hl7.org/fhir/StructureDefinition/{Type}`, which hl7.org resolves to whatever the *current* published FHIR release happens to be. They should point at the documentation for the FHIR version the server is actually running, per the official version directory: https://hl7.org/fhir/directory.html

## Current behavior

1. Start `hfs` built for R4, open `/ui/capability-statement`.
2. In the resource table, click a resource type — e.g. Patient.

The link goes to `http://hl7.org/fhir/StructureDefinition/Patient`, which redirects to the current published release (R5 today, and R6 once that publishes). An R4 server therefore sends the reader to the wrong version's documentation, and the destination silently changes as HL7 publishes new releases.

The link is also `http://`, not `https://`, and it is a StructureDefinition canonical — a resource identifier — rather than a documentation page.

## Expected behavior

The link resolves to the resource's page in the version the server is serving: `https://hl7.org/fhir/R4/patient.html` for an R4 server, `.../R4B/`, `.../R5/`, and so on for the others.

## Evidence

The version-correct link is already implemented — it is just never reached for HFS's own CapabilityStatement.

- `crates/ui/src/capability.rs:247-256` — the advertised `profile` wins, and the versioned page is only a fallback:

```rust
let resource_href = safe_canonical_href(advertised_profile).unwrap_or_else(|| {
    if registry
        .resolve(&resource_type)
        .is_some_and(|schema| editor::is_resource(&schema))
    {
        resource_definition_href(version, &resource_type)
    } else {
        String::new()
    }
});
```

- `crates/rest/src/handlers/capabilities.rs:338` — HFS advertises a `profile` for **every** resource:

```rust
"profile": format!("http://hl7.org/fhir/StructureDefinition/{}", resource_type),
```

  so `safe_canonical_href` always succeeds and the fallback never runs against this server's own statement.

- `crates/ui/src/capability.rs:322-335` — `safe_canonical_href` only checks the scheme is http/https and that a host is present. It does not care whether the URL is version-appropriate.

- `crates/ui/src/capability.rs:286-297` — `fhir_docs_root` already maps each version to its documentation root:

```rust
FhirVersion::R4 => "https://hl7.org/fhir/R4/",
FhirVersion::R4B => "https://hl7.org/fhir/R4B/",
FhirVersion::R5 => "https://hl7.org/fhir/R5/",
FhirVersion::R6 => "https://build.fhir.org/",
```

- `crates/ui/src/capability.rs:427-443` — the existing test `core_resource_fallbacks_follow_each_enabled_version_root` asserts the correct per-version URLs, but only for a statement with **no** `profile`, which is not the shape HFS produces.

## Suggested approach

Prefer the version's own documentation page for a core resource type, and fall back to the advertised profile only when the type is not a core resource of that version — i.e. invert the current precedence at `crates/ui/src/capability.rs:247`.

That keeps the useful behaviour for genuinely custom profiles (`https://example.org/Patient|2.0` → `https://example.org/Patient`, asserted at `crates/ui/src/capability.rs:583`) while making every core type link to the right version's docs. A statement that advertises the plain `http://hl7.org/fhir/StructureDefinition/{Type}` canonical carries no information the version root does not already give.

The narrower alternative — special-casing `hl7.org/fhir/StructureDefinition/` canonicals and rewriting them to the version root — works too, but leaves the same trap for any other statement that names an hl7.org canonical.

Worth settling while in there: `FhirVersion::R6` maps to `https://build.fhir.org/`, the continuous build rather than a published version. Check https://hl7.org/fhir/directory.html for the URL the current R6 ballot publishes under and use that if one exists; the CI build is not a stable link.

Also consider whether `crates/rest/src/handlers/capabilities.rs:338` should advertise a version-qualified profile canonical at all, or drop the field where it adds nothing. That is a separate, server-side question — the UI fix should not depend on changing what the server advertises, since the page must render third-party CapabilityStatements correctly too.

## Acceptance criteria

- [ ] On an R4 build, the Patient row links to `https://hl7.org/fhir/R4/patient.html`.
- [ ] The same holds for R4B and R5 against their own roots.
- [ ] A resource advertising a genuinely custom profile still links to that profile.
- [ ] A type that is not a core resource of the active version, with no usable profile, still renders unlinked.
- [ ] A test covers a statement that *does* advertise `http://hl7.org/fhir/StructureDefinition/{Type}` — the case the existing fallback test does not reach.

## Scope

The per-resource links in the resource table (`crates/ui/templates/pages/capability-statement.html:110`). The system-interaction links (`system_interaction_href`, `crates/ui/src/capability.rs:299-307`) already build on `fhir_docs_root` and are version-correct. The operation-definition links at `capability-statement.html:68` point at OperationDefinition canonicals, which are frequently non-HL7 (e.g. `http://sql-on-fhir.org/OperationDefinition/run`, asserted at `crates/ui/src/capability.rs:394`) and are not part of this change.

### LINKED ISSUE #798: ui: Capability Statement raw JSON is never syntax-highlighted — the statement always exceeds the 4,000-line render budget
## Summary

The "Raw CapabilityStatement (JSON)" fold on `/ui/capability-statement` renders as an unhighlighted monospace blob. The highlighted, line-numbered, foldable viewer was built for this page (#760 item 1), but HFS's own CapabilityStatement exceeds the render budget, so the plain `<pre>` fallback is what actually ships.

## Current behavior

Start `hfs`, open `/ui/capability-statement`, expand "Raw CapabilityStatement (JSON)". The statement appears as plain escaped text — no colors, no line numbers, no folds.

The template picks between the two renderings — `crates/ui/templates/pages/capability-statement.html:134-139`:

```html
{% if raw_fallback %}
<pre class="detail__code">{{ raw }}</pre>
{% else %}
{% include "partials/json-view.html" %}
{% endif %}
```

`raw_fallback` is set when `json_view::try_lines` refuses the document — `crates/ui/src/lib.rs:2775-2794`:

```rust
const MAX_LINES: usize = 4_000;
const MAX_ESTIMATED_HTML_BYTES: usize = 2 * 1024 * 1024;
match json_view::try_lines(&statement, json_view::RenderOptions {
    include_paths: false,
    budget: Some(json_view::RenderBudget { max_lines: MAX_LINES, max_estimated_html_bytes: MAX_ESTIMATED_HTML_BYTES }),
}) {
    Ok(lines) => (Some(view), lines, String::new(), false),
    Err(_) => (Some(view), Vec::new(), serde_json::to_string_pretty(&statement).unwrap_or_default(), true),
}
```

## Why the budget is always exceeded

Estimated from the data files, not measured — worth confirming by loading the page and checking whether the highlighted branch ever renders.

`data/search-parameters-r4.json` yields **1,706** (resource type, search parameter) rows across **135** base types. On top of that, `build_common_search_params` (`crates/rest/src/handlers/capabilities.rs:507`) adds `_id`, `_lastUpdated`, `_tag`, `_profile`, `_security`, `_text` and more to *every* resource — roughly 135 × 8 ≈ 1,080 more.

Each `searchParam` entry is at minimum 6 pretty-printed lines (`{`, `name`, `type`, `definition`, `documentation`, `}`) — `crates/rest/src/handlers/capabilities.rs:410-417` — and gains several more when modifier extensions are attached (`capabilities.rs:421`). Before counting the 9 interactions per resource (`capabilities.rs:323-333`), the per-resource scalar fields (`capabilities.rs:336-350`), and the `searchInclude` / `searchRevinclude` arrays (`capabilities.rs:355`), that is already on the order of **15,000+ lines** against a 4,000-line ceiling.

So the fallback is not an edge case for large deployments — it is the only branch that runs.

#760 anticipated exactly this and left the decision open:

> Use `json_view::try_lines` with a `RenderBudget`, not the unbounded `lines()`. […] a CapabilityStatement is exactly the kind of document that blows past it. Decide what the page does when the budget is exceeded: falling back to today's `<pre>` is reasonable; a 413 is not.

The fallback was the right call for safety; the result is that the highlighting is unreachable in practice.

## Expected behavior

The raw statement renders with the highlighted, line-numbered, foldable viewer — the same one the Resource Editor uses — for a normal HFS CapabilityStatement.

## Suggested approach

Options, roughly in order of preference:

1. **Raise the budget for this page.** The 4,000-line / 2 MiB pair was chosen for `crates/ui/src/lib.rs:1820-1824`, a POST fragment endpoint that takes a user-supplied body, where the comment says the limit exists "so the configured request-body limit cannot be used for amplification." This page has no user-supplied document — it renders the server's own statement — so the amplification argument does not apply, and the constants can be sized to fit a real statement. Measure a full R4 statement first and set the ceiling above it with headroom; keep *some* ceiling so a pathological third-party statement cannot wedge the page.

2. **Render the fold lazily.** Keep the budget, but load the highlighted view into the `<details>` on first expand rather than with the page, so the cost is only paid when someone asks for it. This also removes the JSON from the initial page weight, which is worth something regardless.

3. **Render collapsed by default.** `json_view` already emits fold state; starting deep nodes folded cuts the rendered line count sharply while keeping the whole document reachable. More work than (1).

Whichever is chosen, the fallback path should stay, and it should say why it fired rather than silently degrading — a short note above the `<pre>` explaining that the statement was too large to highlight.

## Acceptance criteria

- [ ] On a default R4 `hfs`, expanding the raw JSON fold shows the highlighted viewer with line numbers and working fold arrows.
- [ ] The same holds for the largest FHIR version the build enables.
- [ ] A statement that still exceeds whatever ceiling is chosen falls back to `<pre>` with a visible explanation, not silently.
- [ ] The fragment endpoint at `crates/ui/src/lib.rs:1820` keeps its existing budget — its constants are there for a different reason and should not be relaxed alongside this page's.

## Related

- #760 — delivered the viewer for this page and flagged the budget question

Gordex2014:
Closes #797.
Closes #798.

## Batch status

- [x] #797: link core resources to documentation for the active FHIR version.
- [x] #798: render large CapabilityStatements with bounded, incremental syntax highlighting.

## What changed

- Core resource links now follow the active FHIR version. R6 uses the checked-in `6.0.0-ballot4` documentation corpus.
- The normal CapabilityStatement page stays lightweight and does not materialize the raw document.
- Small JSON subtrees continue to use the existing foldable `json_view` renderer.
- Large objects and arrays render one bounded level at a time through a CapabilityStatement-only GET fragment route.
- Arrays and large objects page at 100 items. Pagination replaces the current page instead of appending it.
- Only one nested branch remains active. Collapsing a node aborts its requests and unloads all descendants.
- Without JavaScript, the existing link opens plain JSON instead of a highlighted DOM.
- Batch, the resource editor, `/ui/json-view/render`, and the shared JSON renderer keep their existing behavior and limits.

## Rendering flow

```text
CapabilityStatement page
        |
        +-- no JavaScript -----> plain JSON fallback
        |
        `-- JavaScript --------> GET bounded fragment
                                   |
                    +--------------+--------------+
                    |                             |
              large container               small subtree
                    |                             |
          paged one-level outline       existing json_view
                    |
          expand one active branch
                    |
          collapse: abort + unload
```

## Bounds

- 100 items per object or array page.
- 1,000 lines and 1 MiB for a fully highlighted subtree.
- 1 MiB maximum rendered fragment response.
- JSON Pointer limited to 1,024 bytes and 32 segments.
- A single active branch keeps the practical live-DOM ceiling below about 4,200 rows.

## Versioned resource destinations

| Active version | Patient destination |
| --- | --- |
| R4 | `https://hl7.org/fhir/R4/patient.html` |
| R4B | `https://hl7.org/fhir/R4B/patient.html` |
| R5 | `https://hl7.org/fhir/R5/patient.html` |
| R6 | `https://hl7.org/fhir/6.0.0-ballot4/patient.html` |

Matching stays narrow. Exact HL7 core canonicals resolve to the active-version resource page; safe custom profiles remain link targets, and unsafe or unknown destinations remain plain text.

## Verification

- `cargo test -p helios-ui`: 251 passed.
- Explicit non-default R5 test with all FHIR features: passed.
- Combined HFS build with R4, R4B, R5, R6, SQLite, and UI: passed.
- Frontend unit tests: 8 passed.
- CapabilityStatement Chromium tests: 4 passed.
- No-JavaScript fallback test: passed.
- Batch and editor JSON-view regression tests: passed.
- Formatting and diff checks: passed.
- Security/performance, UI/accessibility, and FHIR/version reviews: no findings after remediation.

Browser QA at 1280 x 800 exercised `root -> rest -> 0 -> resource` on R4, R4B, R5, and R6. Each first resource page held 100 items and 1,045 live DOM nodes. R4 pagination replaced that page with 47 items and reduced the DOM to 568 nodes; collapsing the root removed every rendered JSON fragment. Syntax token colors, focus restoration, the single-scroll layout, and all four versioned Patient links were verified.

## Notes

- Fragment requests rebuild the current tenant/version CapabilityStatement instead of caching it, so they do not introduce new staleness rules.
- Existing single-version validator feature propagation still prevents an extra R4B-only workspace check; the combined supported-version build passes, and this PR does not change validator code.
- Browser QA evidence is attached below.

## Evidence

- [R4 syntax highlighting and pagination screenshot](https://github.com/user-attachments/assets/489c53c5-066b-4a89-8daf-64280fcd2572)
- [R4 interaction recording](https://github.com/user-attachments/assets/ecab37af-3b60-49ca-b6c9-dcbcebbebaa7)


Gordex2014:
## Evidence

R4 syntax highlighting and pagination:

qa-r4-highlighted-page2

R4 interaction recording:

https://github.com/user-attachments/assets/ecab37af-3b60-49ca-b6c9-dcbcebbebaa7

smunini (approved):

PR #815, opened 2026-08-31
perf(persistence): batch the sqlite bulk-submit ingest into per-batch transactions

### LINKED ISSUE #813: perf: bulk-submit sqlite ingest is ~43 resources/s in release — per-entry autocommit overhead
## Measurement

One-shot `$bulk-submit` of 20,000 minimal Patients, single NDJSON file, local file server, SQLite backend, on the #809 branch:

| build | throughput | 10,000 patients | 20,000 patients |
|---|---|---|---|
| debug | ~17.5/s | ~9.5 min | ~19 min |
| **release** | **~43.5/s** | **~3.8 min** | **7.7 min** |

The rate is flat across the run (no degradation as tables grow), so the cost is per-entry overhead, not indexing scale.

## Where the time goes

`process_single_entry` + `process_entries` do, per resource, each on its own pooled connection with its own implicit (fsynced) transaction:

1. `read` — exists check for create-vs-update
2. `create` — resource row + history row + search index + FTS
3. `record_change` — rollback log insert
4. `store_entry_result` — per-line receipt insert
5. (per batch of 100) manifest counter updates

That is 4+ WAL commits per resource; at SQLite fsync cost that bounds throughput far below what the hardware can do. The per-statement connection scoping was deliberate (#646/#711 — a connection held across awaits deadlocked two workers), so the fix is not "hold a connection", it is explicit batching:

## Proposal

- Wrap each batch of `process_entries` (100 entries) in **one IMMEDIATE transaction** on one connection: entry writes, change records, and entry-result inserts commit together. One fsync per 100 resources instead of 400+.
- Prepared-statement reuse inside the batch.
- The #711 deadlock concern stays addressed: the connection is held across sync statement runs only, never across awaits — batch inputs are collected before the transaction opens.

Expected order-of-magnitude: several hundred to a few thousand resources/s on the same hardware.

Context: #809's fixes made large ingests *correct* (status persists, polls throttle, completion lands); this issue is about making them *fast*. Numbers from the run Angela requested (10k–20k patients).

angela-helios:
Closes #813.

`process_entries` paid four-plus autocommit fsyncs per resource (exists-read, create/update with history+index, rollback record, per-line receipt — each on its own pooled connection). Each batch now runs through the same `SqliteTransaction` the FHIR transaction-Bundle path uses — **one IMMEDIATE transaction for the batch's reads, writes, history, and index rows** — with the rollback records and entry receipts following in one batched transaction each: three fsyncs per hundred resources instead of four hundred.

The #646/#711 rule stands: the connection is held across sync statement runs only, never across awaits. Entry semantics are unchanged (import-mode updates, updates-not-allowed skips, per-entry errors without aborting the batch, max_errors behavior, soft-deleted corner).

## Measured (20,000-Patient one-shot, release build)

| | baseline (#809) | batched |
|---|---|---|
| throughput | ~43.5/s | **~109/s (2.5x, full index parity)** |
| 10,000 patients | ~3.8 min | **~92 s** |
| 20,000 patients | 7.7 min | **184 s** |

Rate is flat across the run. Suites green: sqlite bulk-submit concurrency (which also dropped 45s → 13s), persistence lib (804), crud, and the rest bulk_submit/jwe/oauth integration + lib suites.

### DIFF
--- a/crates/persistence/src/backends/sqlite/backend.rs
+++ b/crates/persistence/src/backends/sqlite/backend.rs
@@ -72,6 +72,11 @@ fn load_tenant_stored_params(
 static MEMORY_DB_COUNTER: AtomicU64 = AtomicU64::new(0);
 
 /// SQLite backend for FHIR resource storage.
+///
+/// Cloning is cheap — the pool and the registries are shared handles — and
+/// lets a [`SqliteTransaction`](super::transaction::SqliteTransaction) carry
+/// its parent backend so both write paths index identically.
+#[derive(Clone)]
 pub struct SqliteBackend {
     pool: Pool<SqliteConnectionManager>,
     config: SqliteBackendConfig,
--- a/crates/persistence/src/backends/sqlite/bulk_submit.rs
+++ b/crates/persistence/src/backends/sqlite/bulk_submit.rs
@@ -635,17 +635,31 @@ impl BulkSubmitProvider for SqliteBackend {
 
         let mut results = Vec::new();
         let mut error_count = 0u32;
+        let mut changes: Vec<SubmissionChange> = Vec::new();
+        let mut aborted_on_max_errors = false;
+
+        // One IMMEDIATE transaction per batch (#813): entry rows, history,
+        // and search-index writes for the whole batch land on a single fsync
+        // instead of four-plus autocommits per resource. The connection is
+        // held across sync statement runs only — batch inputs arrived as a
+        // Vec, and the change/receipt inserts run in their own batches after
+        // commit — so #646/#711's held-across-awaits deadlock stays out.
+        let mut txn = <Self as crate::core::TransactionProvider>::begin_transaction(
+            self,
+            tenant,
+            crate::core::TransactionOptions {
+                fhir_version: Some(FhirVersion::default_enabled()),
+                ..Default::default()
+            },
+        )
+        .await?;
 
         for entry in entries {
             // Check if we've hit max errors
             if options.max_errors > 0 && error_count >= options.max_errors {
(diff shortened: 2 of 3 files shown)


angela-helios:
Both findings were real — fixed in df672b61f:

**1. Index parity**: `SqliteTransaction` now delegates to the backend's own `index_resource` (dynamic extraction with minimal fallback, **FTS content**, **contained resources**) after `delete_search_index`, so transactional writes index exactly like direct ones — this also closes the same long-standing gap for FHIR transaction Bundles, which shared the reduced path. SearchParameter writes inside a transaction set a flag and **invalidate the tenant's cached registry after the commit** that makes them readable (create guarded by `create_affects_overlay`, update/delete unconditional, mirroring the storage path).

**2. Atomic bookkeeping**: the rollback records and per-line receipts moved **inside the batch transaction** on its own connection (`SqliteTransaction::with_connection`). They commit — or vanish — with the resource writes they describe: no crash window can leave durable writes without their rollback pre-image, and receipts can no longer be lost for a committed batch. This is stronger than the pre-batching ordering, which could orphan a rollback row.

**Corrected numbers** (20,000-Patient one-shot, release): **184s, ~109 resources/s — 2.5x** over the ~43.5/s baseline with identical index output. The earlier 241s/241/s figure was flattered by the missing FTS/contained work; PR description updated.

New regression test `ingested_entries_index_like_direct_writes_and_commit_their_bookkeeping` asserts FTS rows, contained-resource index rows, and same-transaction rollback+receipt rows after an ingest. Full persistence + rest bulk-submit suites green.

smunini (approved):

PR #818, opened 2026-08-31
fix(ui): the history version rail heads its list with "Versions"

### LINKED ISSUE #796: ui: History version rail heads the list with the resource id; it should read "Versions"
## Summary

The version rail on the History page heads its list with the resource's `Type/id`. Replace that with the static word "Versions".

## Current behavior

At `/ui/history`, load an instance (e.g. `Patient` / `85`). The rail above the version list reads `Patient/85`, and the line immediately below it reads `/Patient/85/_history`.

Before anything is loaded, the same element reads "Pick an instance".

- `crates/ui/templates/pages/history.html:43-50`

```html
<aside class="card history__rail" aria-label="{{ i18n.t("history-versions-label") }}">
  <div class="history__rail-head">
    <span class="icon">{% include "icons/history.svg" %}</span>
    <span id="history-subject">{{ i18n.t("history-pick-instance") }}</span>
  </div>
  <div class="history__rail-path" id="history-path"></div>
  <div class="history__versions" id="history-versions"></div>
</aside>
```

- `crates/ui/assets/history.js:61` and `:65` — both the success and the error path overwrite it:

```js
subject.textContent = type + "/" + id;
```

## Expected behavior

The rail head reads "Versions" at all times — before a resource is loaded and after.

The identity of the resource is not lost: `#history-path` directly beneath it already carries `/{type}/{id}/_history` (`crates/ui/assets/history.js:48`), and the locate form at the top of the page still shows the type and id the user entered.

## Suggested approach

The string already exists — `locales/en/main.ftl:200`, `history-versions-label = Versions` — and is currently used only as the `aria-label` on the enclosing `<aside>`. Reusing it for the visible head means no new locale key in any of the three files, and the `aria-label` then duplicates a visible label, so it can be dropped in favour of `aria-labelledby` pointing at the head (or removed, since the visible text now names the region).

Changes:

1. `crates/ui/templates/pages/history.html:46` — render `{{ i18n.t("history-versions-label") }}` instead of `{{ i18n.t("history-pick-instance") }}`, and drop the `id="history-subject"` if nothing else needs to address it.
2. `crates/ui/assets/history.js:28`, `:61`, `:65` — remove the `subject` lookup and both assignments. No `data-msg-*` entry is needed on `#history` (`history.html:19-22`) because the text becomes server-rendered and never changes.
3. `history-pick-instance` (`locales/en/main.ftl:201`, `locales/es/main.ftl:197`, `locales/de/main.ftl:197`) becomes unused — `crates/ui/templates/pages/history.html:46` is its only reference — so remove it from all three locale files.

## Scope

The rail head only. The `/{type}/{id}/_history` path line below it, the locate form, the Instance / Type Feed / System Feed tabs, and the diff panel are unchanged.

angela-helios:
Closes #796.

The rail head is now the static **Versions** label at all times — reusing `history-versions-label`, so no new locale keys — with the aside pointing at it via `aria-labelledby` instead of the duplicated `aria-label`. `history.js` no longer rewrites the head on load or error; the resource identity stays in the `/{type}/{id}/_history` path line beneath it and in the locate form. The now-unused `history-pick-instance` strings are pruned from en/es/de.

History spec updated to assert the static head plus the path line; history 5/5 and a11y 48/48 green.

### DIFF
--- a/crates/ui/assets/history.js
+++ b/crates/ui/assets/history.js
@@ -25,7 +25,6 @@
   if (!root || !window.fetch) return;
 
   var messages = root.dataset;
-  var subject = document.getElementById("history-subject");
   var pathEl = document.getElementById("history-path");
   var versionsEl = document.getElementById("history-versions");
   var controls = document.getElementById("history-controls");
@@ -58,11 +57,9 @@
       .then(function (bundle) {
         versions = parseBundle(bundle);
         if (!versions.length) throw new Error("not-found");
-        subject.textContent = type + "/" + id;
         render();
       })
       .catch(function (error) {
-        subject.textContent = type + "/" + id;
         versionsEl.textContent = "";
         controls.hidden = true;
         diffEl.innerHTML =
--- a/crates/ui/e2e/pages/history.ts
+++ b/crates/ui/e2e/pages/history.ts
@@ -16,6 +16,9 @@ export class HistoryPage {
   get subject(): Locator {
     return this.page.locator("#history-subject");
   }
+  get path(): Locator {
+    return this.page.locator("#history-path");
+  }
   get controls(): Locator {
     return this.page.locator("#history-controls");
   }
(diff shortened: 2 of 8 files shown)


angela-helios:
CI note: the failing `shared back links` design-system spec is red on **main** itself — it expects the page-head action slot that one-shot (#781) removed; a merge crossing left it asserting a layout that no longer exists. The repair (consumers declare whether they carry an action) lives in #817; the identical commit is cherry-picked here so this PR's CI can go green — git deduplicates it when #817 merges.

smunini (approved):

PR #819, opened 2026-08-31
feat(subscriptions,ui): the Last 24 hrs column draws the frame's sparkline

### LINKED ISSUE #782: subscriptions: per-subscription 24h delivery series for the Last 24 hrs sparkline

angela-helios:
Closes #782.

The half-hour delivery ring from #586 already held the series the frame's sparkline needs — it was only ever summed. `DeliveryStats::series` now exposes it chronologically (48 buckets, zeros where nothing was recorded), `SubscriptionRow` carries it as `delivered_series`, and the operator page scales it **server-side into SVG polyline points** — an inline `<svg>`, no client script, per the crate's no-JS rules.

Per the frame (405:2): **green** for active, **red** for failing, **dotted amber** for a listener-less websocket. The delivered count stays as the SVG's accessible name and tooltip. A snapshot without a series renders the plain count — no fabricated flat line (#555).

Tests: a chronological-series unit test on the ring; the subscriptions page test asserts the state-tinted polylines, the accessible name, and that a series-less row keeps the count. Full ui/observability/subscriptions suites green; e2e subscriptions + a11y + design-system green except the pre-existing back-link spec red on main, whose fix is already in #817.

### DIFF
--- a/crates/observability/src/subscriptions.rs
+++ b/crates/observability/src/subscriptions.rs
@@ -45,6 +45,10 @@ pub struct SubscriptionRow {
     /// Deliveries abandoned in the last 24 hours (permanent errors and
     /// exhausted retries).
     pub failed_24h: u64,
+    /// `delivered_24h` as a chronological series — one count per half-hour
+    /// bucket, oldest first (#782). Empty when the engine carries no series;
+    /// the UI then renders no sparkline rather than a fabricated flat line.
+    pub delivered_series: Vec<u64>,
 }
 
 /// The engine's current inventory for one tenant.
@@ -104,6 +108,7 @@ mod tests {
                     delivered_24h: 12,
                     first_try_24h: 11,
                     failed_24h: 1,
+                    delivered_series: vec![0, 4, 8],
                 }],
             }
         }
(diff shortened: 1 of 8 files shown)


angela-helios:
CI note: the failing `shared back links` design-system spec is red on **main** itself — it expects the page-head action slot that one-shot (#781) removed; a merge crossing left it asserting a layout that no longer exists. The repair (consumers declare whether they carry an action) lives in #817; the identical commit is cherry-picked here so this PR's CI can go green — git deduplicates it when #817 merges.

smunini (approved):

PR #846, opened 2026-09-01
fix(ui): separate All Resources from the resource grid

### LINKED ISSUE #828: ui: Add vertical spacing below "All Resources" on the Bulk Export page
## Summary
On the Bulk Export page, the "All Resources" checkbox sits flush against the grid of individual resource type checkboxes below it. It needs a bit of vertical spacing so it reads as a separate, higher-level control rather than the first entry in the type list.

## Current behavior
Navigate to `/ui/bulk-export`. In the "Resource Types" card, the "All Resources" checkbox is rendered as a `.typegrid__item` label immediately followed by the `.typegrid` grid of resource types. The only separation is the grid's own `padding: 6px 0`, so "All Resources" visually blends into the first row of the type list.

## Expected behavior
A visible gap below "All Resources" that separates it from the resource type grid, making it clear it is a master toggle for the list rather than one of the types.

## Evidence
- `crates/ui/templates/pages/bulk-export.html:90` — the "All Resources" label, immediately followed by the `.typegrid` block at line 91
- `crates/ui/assets/app.css:5782` — `.typegrid { gap: 6px 18px; padding: 6px 0; }` is the only thing currently separating the two
- `crates/ui/assets/app.css:5789` — `.typegrid__item` carries no margin, so the shared class gives "All Resources" no extra breathing room

## Suggested approach
Give the "All Resources" label its own modifier class (e.g. `typegrid__item typegrid__all`) in `bulk-export.html:90` and add a `margin-bottom` rule near the `.typegrid` block in `app.css` (~line 5782). Avoid adding the margin to `.typegrid__item` itself, since that class is reused for every resource type in the grid.

## Scope / out of scope
Spacing only. No change to the checkbox behavior, the no-JS progressive enhancement path noted in the comment at `bulk-export.html:88-89`, or the layout of the resource type grid itself.

Gordex2014:
## Summary

- add an 8 px top margin to the Bulk Export resource grid, producing a nominal 14 px separation from the master toggle
- add responsive geometry regression coverage at desktop and mobile viewport sizes
- preserve the existing grid layout and checkbox behavior

## Tests

- `cargo test -p helios-ui`
- `cargo build -p helios-hfs --features ui`
- `npm run test:unit`
- focused Bulk Export Playwright suite: 11 passed
- design-system Playwright suite: 14 passed
- full Playwright suite: 317 passed, 6 skipped

## QA

- verified the 14 px separation at 1280×800 and 390×844
- verified initial, uncheck, recheck, and Clear checkbox states

Closes #828

### DIFF
--- a/crates/ui/assets/app.css
+++ b/crates/ui/assets/app.css
@@ -5783,6 +5783,7 @@ details.json-fold[open] > summary .icon {
   display: grid;
   grid-template-columns: repeat(auto-fill, minmax(220px, 1fr));
   gap: 6px 18px;
+  margin-top: 8px;
   padding: 6px 0;
 }
 
--- a/crates/ui/e2e/tests/bulk-export.spec.ts
+++ b/crates/ui/e2e/tests/bulk-export.spec.ts
@@ -31,6 +31,28 @@ test("All Resources is the enhanced default", async ({ bulkExport }) => {
   ).toBe(true);
 });
 
+test("All Resources is visually separated from the resource grid", async ({
+  page,
+  bulkExport,
+}) => {
+  for (const viewport of [
+    { width: 1280, height: 800 },
+    { width: 390, height: 844 },
+  ]) {
+    await page.setViewportSize(viewport);
+    await bulkExport.goto();
+
+    const allResourcesBottom = await bulkExport.allResources.evaluate(
+      (input) => input.closest("label")!.getBoundingClientRect().bottom,
+    );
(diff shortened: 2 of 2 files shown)


Gordex2014:
## Visual evidence

### Before, desktop, 1280x800
Before desktop

### After, desktop, 1280x800
After desktop

The mobile viewport was also verified at 390x844 with the same measured 14 px gap.

smunini (approved):

PR #847, opened 2026-09-01
fix(ui): center the busy-button ring

### LINKED ISSUE #786: Busy-state button ring renders 2px off center (Batch/Transaction Execute)
## Summary

The busy ring drawn inside a button while it is working (`.btn[aria-busy="true"]::after`) sits 2px right and 2px below the button's center. It is most visible on the Batch/Transaction page's Execute button, but the rule is shared, so every busy button is affected.

## Current behavior

On `/ui/batch`, pick a bundle, then press Execute. The button's label is replaced by the spinning ring, and the ring is offset down-and-right rather than centered in the button.

## Expected behavior

The ring is centered in the button, both horizontally and vertically.

## Evidence

The offset is a box-model mismatch: the ring's centering math assumes a 14px total size, but the pseudo-element renders at 18px.

- `crates/ui/assets/app.css:1474-1481` — the ring is sized as a 14×14 box with a 2px border:

```css
.spinner,
.btn[aria-busy="true"]::after {
  width: 14px;
  height: 14px;
  border-radius: 50%;
  border: 2px solid var(--outline);
  border-top-color: var(--accent);
  animation: spin 0.8s linear infinite;
}
```

- `crates/ui/assets/app.css:1501-1506` — it is centered by subtracting half of 14px:

```css
.btn[aria-busy="true"]::after {
  content: "";
  position: absolute;
  top: calc(50% - 7px);
  left: calc(50% - 7px);
}
```

- `crates/ui/assets/app.css:206-208` — the reset is `* { box-sizing: border-box; }` only. The universal selector does not match pseudo-elements, and `box-sizing` is not an inherited property, so `::after` falls back to the initial `content-box`. The ring's rendered size is therefore 14 + 2 + 2 = **18px**, and centering it needs a `-9px` offset, not `-7px` — leaving it 2px off on each axis.

Note that `.spinner` (used by the standalone `.busy-status` region, `crates/ui/templates/pages/batch.html:113`) is a real `<span>`, so `*` does match it and it renders at the intended 14px. Only the in-button pseudo-element ring is affected.

## Suggested approach

Pick one:

1. Add `box-sizing: border-box;` to the shared `.spinner, .btn[aria-busy="true"]::after` rule so the declared 14px is the rendered size and the existing `-7px` math is correct. This also keeps the in-button ring the same visual size as the standalone `.spinner`.
2. Extend the reset at `crates/ui/assets/app.css:206` to `*, *::before, *::after`. Broader blast radius — it would change any other pseudo-element that currently relies on `content-box`, so it needs a wider visual check.
3. Change the offsets to `calc(50% - 9px)`. Fixes the centering but leaves the in-button ring 18px while the standalone spinner is 14px.

Option 1 is the narrowest fix.

## Scope

The affected rule is shared, not Batch-specific: any control driven into the busy state by `crates/ui/assets/busy.js` (`during()` sets `aria-busy="true"`) shows the same offset ring.

dougc95:
## Summary

Fixes #786.

The `*` box-sizing reset does not match pseudo-elements, so the `.btn[aria-busy=true]::after` ring rendered content-box at 18px (14px + 2px borders per side) while the `calc(50% - 7px)` centering offsets assumed 14px, leaving the ring 2px down-and-right of center.

This takes option 1 from the issue — the narrowest fix: declare `box-sizing: border-box` on the shared `.spinner, .btn[aria-busy=true]::after` rule, so the declared 14px is the rendered size, the existing `-7px` math is correct, and the in-button ring stays the same visual size as the standalone `.spinner`.

## Verification

Measured against the built stylesheet in Chromium: the ring computes to 14x14 with offset (0, 0) from the button center; forcing `box-sizing: content-box` back reproduces the reported 18px box with a (+2, +2) offset. Screenshots of the running app to follow in a comment.

### DIFF
--- a/crates/ui/assets/app.css
+++ b/crates/ui/assets/app.css
@@ -1493,6 +1493,10 @@ select.field__input {
 
 .spinner,
 .btn[aria-busy="true"]::after {
+  /* Explicit border-box (#786): the `*` reset does not match pseudo-elements,
+     and content-box would render the ::after ring at 18px — off center against
+     the -7px offsets below. */
+  box-sizing: border-box;
   width: 14px;
   height: 14px;
   border-radius: 50%;


dougc95:
Verified against the running server (debug build, `/ui/batch`, dark theme, Chromium via the e2e Playwright toolchain):

- Loaded a 3-entry batch bundle to reach the preflight stage, then held the Execute button in the `aria-busy=true` state busy.js sets during the POST.
- **With the fix:** the ring computes to 14x14 (`box-sizing: border-box`) and sits at exactly (0, 0) offset from the button center.
- **Pre-fix behavior reproduced** by overriding the rule back to `content-box`: the ring renders 18x18 with the reported (+2px, +2px) down-and-right offset.

Before/after close-ups captured at 4x device pixel ratio — attaching below.

dougc95:
before-bug
after-context
after-fixed

smunini (approved):

PR #849, opened 2026-09-01
feat(persistence,rest,ui): byte-based bulk-submit ingest progress

### LINKED ISSUE #827: Bulk Import shows "0% complete" alongside a racing indeterminate progress bar for single-manifest submissions
## Summary

While a single-manifest bulk import streams in, the Bulk Import processing card
reports `processing 0% complete (13853 entries ingested)` and renders a
*sweeping* (indeterminate) progress bar that races across the card on a 1.6s
loop. The two halves contradict each other: the text asserts a precise
percentage while the bar animates as though the percentage is unknown, and the
one number that is actually moving — the entry counter — never drives the bar.

## Current behavior

With a submission whose manifest has a single `output` entry (the common
one-shot case), for the entire duration of the ingest:

- the status line reads `processing 0% complete (N entries ingested)`, with `N`
  climbing into the thousands,
- the percentage stays pinned at `0`,
- the progress bar sweeps left-to-right on a continuous 1.6s animation.

The bar only becomes determinate once the percentage exceeds 0, which for a
single-manifest submission happens only when the whole submission is finished.

## Expected behavior

The bar and the text agree, and the bar reflects real progress. Concretely, one
of:

- drive the bar from the entry counter when a per-file expected `count` is
  known (the Bulk Data manifest carries `count` per output entry), or
- suppress the misleading `0%` from the text while the bar is indeterminate, so
  the card reads e.g. `processing (13853 entries ingested)`.

Either way, a sweeping bar and a stated percentage should not appear together.

## Evidence

Server-side percentage is manifest-granular, so a one-output submission is 0%
until the end — `crates/rest/src/handlers/bulk_submit.rs:763`:

```rust
let pct = if manifests.is_empty() {
    0
} else {
    (manifests.iter().filter(|m| m.status.is_terminal()).count() * 100) / manifests.len()
};
```

and the string is assembled at `crates/rest/src/handlers/bulk_submit.rs:792`:

```rust
format!("processing {pct}% complete ({entries} entries ingested)")
```

The UI drops a `0` percentage on the floor — `crates/ui/src/bulk_import.rs:1176`:

```rust
fn progress_percent(progress: &str) -> Option<u8> {
    let rest = progress.strip_prefix("processing ")?;
    let digits: String = rest.chars().take_while(char::is_ascii_digit).collect();
    let pct: u8 = digits.parse().ok().filter(|p| *p > 0 && *p <= 100)?;
    Some(pct)
}
```

`filter(|p| *p > 0)` turns `0%` into `None`, which selects the indeterminate
branch in `crates/ui/templates/partials/bulk_import_status.html:15`:

```html
<div class="progress-track progress-track--indeterminate" role="progressbar" ...>
```

and that class is what races — `crates/ui/assets/app.css:3173`:

```css
.progress-track--indeterminate .progress-fill {
  position: absolute;
  width: 30%;
  animation: progress-sweep 1.6s ease-in-out infinite;
}
```

Note the text is rendered from the raw `X-Progress` string independently of
`percent` (`bulk_import_status.html:19`), which is why the discarded `0%` still
reaches the user as text while the bar goes indeterminate.

## Reproduction

1. Serve a Bulk Data manifest with a **single** `output` entry pointing at an
   NDJSON file of ~15k resources:
   ```bash
   python3 -m http.server 8000   # serving manifest.json + Patient.ndjson
   ```
2. Kick off a submission (`fhirBaseUrl` is required alongside `manifestUrl`):
   ```bash
   curl -X POST 'http://localhost:8080/$bulk-submit' \
     -H 'Content-Type: application/json' -H 'Prefer: respond-async' \
     -d '{"resourceType":"Parameters","parameter":[
       {"name":"submitter","valueIdentifier":{"system":"http://example.org/submitters","value":"c1"}},
       {"name":"submissionId","valueString":"s1"},
       {"name":"fhirBaseUrl","valueUrl":"http://localhost:8000"},
       {"name":"manifestUrl","valueUrl":"http://localhost:8000/manifest.json"}]}'
   ```
3. Open the Bulk Import page for that submission and watch the processing card
   while the file streams in.

Observed on a live import: `processing 0% complete (13853 entries ingested)`
with the bar sweeping continuously.

## Suggested approach

The cheapest correct fix is in the UI: keep `progress_percent` returning `None`
for `0`, but also strip the `0% complete` fragment from the displayed text so
the indeterminate bar is not paired with a percentage claim.

The better fix drives the bar from real work: the manifest's per-output `count`
is already known at ingest time, so `processed_entries / expected_entries` gives
a genuine determinate percentage for the single-manifest case. That would make
the common one-shot import show smooth progress instead of a sweep. This likely
means carrying an expected-entry total on the manifest record in
`helios-persistence` and emitting an entry-based percentage from
`bulk_submit.rs` rather than the manifest-count ratio.

## Scope / out of scope

Out of scope: the manifest-granular percentage is correct and useful for
*multi*-manifest submissions; this issue is about the single-manifest case and
the text/bar contradiction, not about removing manifest-granular reporting.

### LINKED ISSUE #848: ui/rest: bulk import progress bar stays at 0% (indeterminate sweep) for one-shot submissions — report byte-based progress
## Problem

On a one-shot bulk import (a single manifest, one big NDJSON file) the UI progress bar sweeps indeterminately for the entire run and the reported percentage stays at 0% until the very end. The percentage behind `X-Progress` is manifest-granular — completed manifests / total manifests — so a single-manifest submission is 0% for its whole life and jumps straight to done. The entry counter moves, but the bar itself never fills, which reads as "nunca termina de cargar".

## Proposal: byte-based progress

The recipient knows how many bytes each output file advertises (`Content-Length`) and how many bytes the ingestion engine has consumed. Report the percentage as consumed/total bytes:

- `SubmitInputFetcher::open_file_stream` returns the stream **plus its content length** (`None` for sizeless streams, e.g. gzip-decompressed bodies).
- The worker wraps each file stream in a counting reader and flushes `bytes_processed`/`bytes_total` to the manifest row every few seconds from the lease keeper's task, so a status poll sees movement *while* a large file streams.
- The status endpoint computes the percentage from summed bytes (capped at 99 until terminal), falling back to the old manifest-count fraction when no totals are known.
- The UI renders 0% as a determinate empty bar that fills within seconds — the indeterminate sweep is reserved for recipients that report no percentage at all.

Needs a schema addition (`bytes_processed`/`bytes_total` on manifests) across the SQLite, PostgreSQL, MongoDB, and S3 backends.

angela-helios:
Closes #848
Closes #827

## What

The in-progress percentage behind `X-Progress` was manifest-granular (completed manifests / total), so a one-shot submission read **0% for its entire run** and the UI bar swept indeterminately without ever filling. The percentage is now **byte-based**: consumed bytes over the files' advertised sizes, updated every few seconds while a file streams.

## How

- **Fetcher** — `SubmitInputFetcher::open_file_stream` returns the stream plus its `Content-Length` (`None` for sizeless streams, e.g. gzip-decompressed bodies; JWE-buffered files report the decrypted length).
- **Worker** — each output file is wrapped in a counting `AsyncBufRead`; the `LeaseKeeper`'s task flushes `bytes_processed`/`bytes_total` to the manifest row every ~3s (plus after each file), through the same fenced update discipline as every other worker write. One sizeless file poisons the total for the whole manifest, so a partial total never inflates the percentage.
- **Storage** — `bytes_processed`/`bytes_total` on manifests across all four backends: SQLite (schema v18 migration), PostgreSQL (`ADD COLUMN IF NOT EXISTS`), MongoDB, S3.
- **Status endpoint** — percentage = summed bytes ratio, capped at 99 until terminal; falls back to the manifest-count fraction when no totals are known. The `(N entries ingested)` suffix stays.
- **UI** — a reported 0% now renders a determinate empty bar (it fills within seconds); the indeterminate sweep is reserved for recipients that report no percentage at all.

## Verified

Live run against a release build, 20,000 Patients (2.7 MB NDJSON, single manifest, single file): the poll reported 1% → 43% → 54% → … → 92% → completed, monotonically, alongside the entry counter. Suites: persistence lib (805), sqlite concurrency (6), rest `bulk_submit` (21, incl. a new byte-percentage test) + `bulk_submit_jwe` (6), ui `bulk_import_http` (28). Worker unit test asserts `bytes_total`/`bytes_processed` land at the file's full size.

### Mid-run: determinate bar at 54%
bar

### Completed, with the percentage climbing in the submission log
log

### DIFF
--- a/crates/persistence/src/backends/mongodb/bulk_submit.rs
+++ b/crates/persistence/src/backends/mongodb/bulk_submit.rs
@@ -202,6 +202,8 @@ fn decode_manifest(doc: &Document) -> StorageResult<SubmissionManifest> {
         total_entries: doc.get_i64("total_entries").unwrap_or(0).max(0) as u64,
         processed_entries: doc.get_i64("processed_entries").unwrap_or(0).max(0) as u64,
         failed_entries: doc.get_i64("failed_entries").unwrap_or(0).max(0) as u64,
+        bytes_processed: doc.get_i64("bytes_processed").unwrap_or(0).max(0) as u64,
+        bytes_total: doc.get_i64("bytes_total").unwrap_or(0).max(0) as u64,
     })
 }
 
@@ -769,6 +771,8 @@ impl BulkSubmitProvider for MongoBackend {
             processed_entries: 0,
             failed_entries: 0,
             lease_expiry: None,
+            bytes_processed: 0,
+            bytes_total: 0,
         };
 
         let mut document = manifest_filter(tenant, submission_id, &manifest.manifest_id);
@@ -784,6 +788,8 @@ impl BulkSubmitProvider for MongoBackend {
         document.insert("failed_entries", 0_i64);
         document.insert("fencing_token", 0_i64);
         document.insert("last_processed_line", 0_i64);
+        document.insert("bytes_processed", 0_i64);
+        document.insert("bytes_total", 0_i64);
 
         self.manifests()
             .await?
@@ -1342,6 +1348,22 @@ impl SubmitWorkerStorage for MongoBackend {
                 "processed_entries": processed_entries as i64,
                 "failed_entries": failed_entries as i64,
                 "last_processed_line": last_processed_line as i64,
+            }},
+        )
+        .await
+    }
+
+    async fn update_manifest_bytes(
+        &self,
+        lease: &ManifestLease,
+        bytes_processed: u64,
+        bytes_total: u64,
+    ) -> Result<(), LeaseError> {
+        self.fenced_update(
+            lease,
+            doc! { "$set": {
+                "bytes_processed": bytes_processed as i64,
+                "bytes_total": bytes_total as i64,
             }},
         )
         .await
--- a/crates/persistence/src/backends/mongodb/bulk_submit.rs
+++ b/crates/persistence/src/backends/mongodb/bulk_submit.rs
@@ -1670,14 +1670,69 @@ impl SubmitWorkerStorage for MongoBackend {
     }
 
     async fn count_active_submissions(&self, tenant: &TenantContext) -> StorageResult<u64> {
-        self.submissions()
+        let tenant_id = tenant.tenant_id().as_str();
+        let cursor = self
+            .submissions()
             .await?
-            .count_documents(doc! {
-                "tenant_id": tenant.tenant_id().as_str(),
+            .find(doc! {
+                "tenant_id": tenant_id,
                 "status": SubmissionStatus::InProgress.to_string(),
             })
             .await
-            .map_err(|e| internal_error(format!("count active submissions: {e}")))
+            .map_err(|e| internal_error(format!("count active submissions: {e}")))?;
+        let subs = collect(cursor).await?;
+        if subs.is_empty() {
+            return Ok(0);
+        }
+
+        let keys: Vec<Document> = subs
+            .iter()
+            .filter_map(|d| {
+                Some(doc! {
+                    "submitter": opt_str(d, "submitter")?,
+                    "submission_id": opt_str(d, "submission_id")?,
+                })
+            })
+            .collect();
+        let cursor = self
+            .manifests()
+            .await?
+            .find(doc! { "tenant_id": tenant_id, "$or": keys })
+            .await
+            .map_err(|e| internal_error(format!("count active submissions: {e}")))?;
+        // (submitter, submission_id) → has a non-terminal manifest. Presence in
+        // the map at all means the submission has manifests.
+        let mut manifests: std::collections::HashMap<(String, String), bool> =
+            std::collections::HashMap::new();
+        for m in &collect(cursor).await? {
+            let (Some(submitter), Some(submission_id)) =
+                (opt_str(m, "submitter"), opt_str(m, "submission_id"))
+            else {
+                continue;
+            };
+            let live = matches!(
+                m.get_str("status").unwrap_or_default(),
+                "pending" | "processing"
+            );
+            *manifests.entry((submitter, submission_id)).or_insert(false) |= live;
+        }
+
+        Ok(subs
+            .iter()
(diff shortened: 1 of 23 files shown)


angela-helios:
Follow-up pushed (efdde53): live testing surfaced a flush-starvation problem the first commit had.

**Symptom** — for the first ~20s of a run the bar sat at 0% while the entry counter climbed into the thousands, then the percentage jumped erratically. Ground truth (reading `bulk_manifests` every second during a 20k ingest): the lease keeper's standalone `UPDATE` starved behind back-to-back batch transactions on SQLite — it waited ~20s for the write lock and then landed a value read ~20s earlier. Two completed manifests were also left with `bytes_processed` at 95% of total: a stale in-flight keeper flush outraced the final one.

**Fix** — the per-batch counts `UPDATE` already wins the write lock right after every batch commit, so the byte counters now ride it: `BulkProcessingOptions` carries the shared `ByteProgress` handle and the SQLite batch bookkeeping persists bytes together with the entry counters. Every `update_manifest_bytes` implementation is now monotonic (`MAX` / `GREATEST` / `$max` / `max`), so a late keeper flush can never walk progress backwards; the keeper flush remains as the fallback path for the other backends, which don't share SQLite's single-writer contention.

Re-probed after the fix (same 20k dataset, 1s DB sampling): bytes advance in lockstep with the entry counter, one batch per tick —

```
t=21s entries=100 bytes=13390/2728890 (0%)
t=22s entries=200 bytes=26890/2728890 (0%)
t=23s entries=300 bytes=40390/2728890 (1%)
t=25s entries=500 bytes=67390/2728890 (2%)
t=27s entries=700 bytes=94390/2728890 (3%)
t=29s entries=900 bytes=121390/2728890 (4%)
```

Testing also uncovered an unrelated pre-existing leak (submissions never marked `completed` by the UI hold tenant slots until kick-offs 429) — filed separately as #850.

smunini (approved):

PR #852, opened 2026-09-01
ui: validate Bulk Export name and custom Since instant

### LINKED ISSUE #830: ui: Validate Bulk Export name and custom Since instant, drop the Since format hint
## Summary
On the Bulk Export builder, drop the "Custom instant" hint text and replace it with real validation: the custom `_since` value must be a valid FHIR `instant`, and the export Name must not be empty.

## Current behavior
At `/ui/bulk-export/new`:

- The Custom instant field carries a hint that restates the format instead of enforcing it:
  ```
  bulk-export-field-since-custom-hint = Used when Since is Custom. RFC 3339, e.g. 2026-08-01T00:00:00Z.
  ```
  The field already has a `placeholder="2026-08-01T00:00:00Z"` showing the same example, and JavaScript already disables the input unless Since is set to Custom (`crates/ui/assets/bulk-export.js:30-33`), so the hint duplicates what the UI already conveys.
- Whatever is typed into Custom instant is passed straight through as `_since` with no format check. `since_instant` trims and returns the raw string (`crates/ui/src/bulk_export.rs:715-724`):
  ```rust
  "custom" => custom.trim().to_string(),
  ```
  Submitting `not-a-date` (or an empty value with Since = Custom) creates a job and sends a malformed `_since` to `$export`; the failure only surfaces later as a failed job.
- The Name field is never validated. `start` copies `form.name.trim()` onto the job (`crates/ui/src/bulk_export.rs:1000`) without checking for emptiness, so an unnamed export lands in the Exports list with a blank name.

## Expected behavior
- The `bulk-export-field-since-custom-hint` text is removed from the form.
- When Since is set to Custom, the entered value is validated as a FHIR `instant` before the job is created. An invalid or empty value re-renders the page with a field-level error and a `400`, and no export is kicked off.
- Submitting with an empty Name re-renders the page with a field-level error and a `400`, and no export is kicked off.

## Evidence
- `crates/ui/templates/pages/bulk-export.html:128-132` — the Custom instant field and its `field__hint` span
- `locales/en/main.ftl:721` — the hint string to remove
- `crates/ui/src/bulk_export.rs:715-724` — `since_instant`, which passes the custom value through unvalidated
- `crates/ui/src/bulk_export.rs:1000` — `name: form.name.trim().to_string()` in the `start` handler, with no emptiness check
- `crates/ui/src/bulk_export.rs:980-998` — the existing validation pattern to follow: `parse_patient_refs` failure re-renders `bulk_export_page` with an error message and sets `StatusCode::BAD_REQUEST`
- `locales/en/main.ftl:706` — `bulk-export-patient-invalid`, the existing error-message key of that shape

## Suggested approach
- Delete the `field__hint` span from `crates/ui/templates/pages/bulk-export.html` and the `bulk-export-field-since-custom-hint` key from all three catalogs (`locales/{en,es,de}/main.ftl` — they must carry the same message set; see `crates/ui/src/i18n.rs:345` and `locales/README.md`). Keep the placeholder.
- Change `since_instant` to return `Result<String, _>` (or add a validating wrapper) so the `custom` arm parses the value as an instant rather than passing it through. `chrono::DateTime::parse_from_rfc3339` is already available in this module; FHIR `instant` additionally requires a timezone offset and second-level precision, which RFC 3339 parsing gives.
- Add the Name and Since checks to `start` alongside the existing `parse_patient_refs` check, reusing the same re-render-with-`400` path so the entered values are not silently discarded. Two new i18n keys, e.g. `bulk-export-name-required` and `bulk-export-since-invalid`, in all three catalogs.
- Consider a `required` attribute on the Name input and a `pattern`/`required` on Custom instant for native client-side feedback, but the server-side check is the one that must exist — the form is designed to work without JavaScript.
- Tests to extend: `crates/ui/tests/bulk_export_http.rs` for the two new `400` paths, and `crates/ui/e2e/tests/bulk-export.spec.ts` for the client-side behavior.

## Scope / out of scope
Only the Bulk Export builder form. The `day`/`week`/`month` presets already generate well-formed instants via `to_rfc3339_opts` and need no validation. The other narrow-down fields (`_elements`, `_typeFilter`) are not in scope.

Gordex2014:
## Summary

- require a nonblank export name and validate active, nonempty Custom Since values before creating a job
- keep validation silent until the first Start Export attempt, then revalidate fields as the user edits them
- show localized inline messages, invalid borders, accessible descriptions, and deterministic first-error focus
- preserve submitted form state on server rejection and keep the same behavior without JavaScript
- follow the bundled FHIR R4 instant grammar, including leap seconds and the `±14:00` timezone boundary
- remove the redundant Custom format hint

## Testing

- `node --check crates/ui/assets/bulk-export.js`
- `cargo test -p helios-ui`
- `cargo build -p helios-hfs --features ui`
- Chromium Bulk Export Playwright suite: 16 passed
- No-JavaScript Bulk Export Playwright suite: 5 passed
- Safari manual validation of initial, submitted, reactive, reset, and FHIR instant boundary states

Closes #830


Gordex2014:
## Safari evidence

### Reactive Custom Since validation

Custom Since inline validation error in Safari

### Reactive Name validation

Name inline validation error in Safari

PR #876, opened 2026-09-02
fix(persistence,rest): _lastUpdated on Postgres; server-populated meta in response bodies

### LINKED ISSUE #871: search(postgres): _lastUpdated returns 500 on every prefix and precision
Found in the #448 Postgres leg. Every `_lastUpdated` search fails with an opaque 500 on the Postgres backend — all precisions (`2026-09-01`, `2026-09-01T00:00:00Z`) and all prefixes (`eq`/`gt`/`ge`/`le`):

```
GET /Patient?_lastUpdated=2026-09-02          → 500
GET /Patient?_lastUpdated=gt2026-09-01T00:00:00Z → 500
```

Server log:

```
ERROR helios_rest::error: internal error while processing request
  error.detail=internal error in postgres: Failed to execute search: error serializing parameter 2
```

## Cause

`build_last_updated_condition` (`crates/persistence/src/backends/postgres/search/query_builder.rs:645`) binds the raw query string as a **text** parameter directly against the `last_updated` column:

```rust
conditions.push(SqlFragment::with_params(
    format!("last_updated {} ${}", op, param_num),
    vec![SqlParam::text(&value.value)],
));
```

`last_updated` is `TIMESTAMPTZ`, and tokio-postgres refuses to serialize a text param into it — hence "error serializing parameter 2" on every query.

Two layers to fix:
1. Bind a real timestamp (or cast) so the query executes at all.
2. Apply the same **precision-range semantics** the date-parameter path already has (the range helper at the top of the same file): `eq` at day precision must mean `>= day AND < day+1`, not a text `=`. The #463 date fix explicitly covered `_lastUpdated`'s separate code path on SQLite; Postgres needs the equivalent.

SQLite answers the same queries correctly. The #448 battery runs 22 checks on this backend; these are the only 2 failures.

### LINKED ISSUE #873: fhir: server-populated meta (versionId, lastUpdated) missing from every response body — UI UPDATED column renders blank
Found in the #448 Postgres leg, but it is server-wide: **both** SQLite and Postgres return stored resources without server-populated `meta.versionId` / `meta.lastUpdated` on every read path:

```
GET /Patient/{id}                → meta absent (SQLite) / only source-supplied meta.profile (PG)
GET /Patient/{id}/_history/1     → meta absent
GET /Patient/{id}/_history       → entries' resources: meta absent
GET /Patient?...                 → searchset entries: meta absent
```

The HTTP headers are correct (`ETag: W/"1"`, `Last-Modified: …`), so versioning itself works — only the resource **bodies** omit the metadata. R4 requires it at least for vread ("The returned resource SHALL have … a `meta.versionId` for the version") and clients routinely rely on `meta.lastUpdated` inside Bundles, where per-resource headers don't exist.

Visible symptom: the Resources workspace's **UPDATED column is blank for every row** on every backend — the UI asks for `meta.lastUpdated` and there is nothing there.

Fix shape: inject `meta.versionId`/`meta.lastUpdated` into the returned body from the row's version and timestamp on read/vread/history/search (merging with any stored `meta` like `profile`, which Postgres already preserves).

angela-helios:
Closes #871
Closes #873

Both found by the #448 Postgres sanity leg; base is `main` (independent of the bulk-progress stack).

## #871 — `_lastUpdated` 500s on Postgres

`build_last_updated_condition` bound the raw query string as a **text** parameter against the `TIMESTAMPTZ` `last_updated` column — tokio-postgres refuses to serialize it ("error serializing parameter 2"), so every precision and prefix returned an opaque 500. The path also predated the precision-range semantics dates got in #463.

Now it binds real timestamps and mirrors `build_date_condition` exactly: `eq` at day precision means `[day, day+1)`, `ne` its complement, `gt`/`sa` exclude the named period, `lt`/`le`/`ge` use the correct boundary, and a full-precision instant falls back to scalar comparison. Unit tests pin the SQL shape, the bound param types, and the period-end arithmetic.

## #873 — response bodies carried no server-populated `meta`

`versionId`/`lastUpdated` lived only in the `ETag`/`Last-Modified` headers: search entries, reads, vreads, history bundles, and batch/transaction responses all returned resources with no server metadata — R4 requires it at least on vread, Bundles have no per-resource headers, and the Resources workspace's UPDATED column rendered blank on every backend.

Fix is read-side and central: `StoredResource::content_with_meta()` merges the row's version and timestamp into the returned body — client-supplied `meta` members (`profile`, `tag`, `security`) survive, `versionId`/`lastUpdated` are overwritten because the row is authoritative — and every echo point uses it: read, vread, history entries, search `to_bundle` (match + include), create/update/patch responses, batch/transaction `BundleEntryResult`, and subscription event payloads. Read-side means **pre-existing rows are fixed immediately**, no migration.

An integration test walks the whole lifecycle: create (v1 meta in the response, profile preserved) → update → read (v2) → vread (v1 keeps *its* version) → history (per-version meta) → search (entries carry meta).

## Verified

- Live against postgres:16 with the #448 dataset: the leg's battery goes **20/22 → 22/22**, and a Patient read returns `meta.profile + versionId + lastUpdated` together.
- Suites: persistence lib (805), full helios-rest (all integration files, 0 failures), postgres query-builder units (22).


angela-helios:
CI fix pushed: `query_builder_tests::test_prefix_operators` asserted the scalar operators the pre-#871 code emitted for day-precision `_lastUpdated` — the exact broken semantics this PR replaces (my local run had filtered on `postgres_` and missed that module). The test now pins the range shapes (`eq` → `[day, day+1)`, `ne` → its complement, `gt` → period end, etc.), the bound counts, and the scalar fallback for full-precision instants. Full `postgres_tests` query-builder module: 26/26 locally.

angela-helios:
CI here was failing on `postgres_bulk_submit_import_directives_round_trip` — the known flake #911 fixes (another test's worker claims the manifest before this test reads it back; it has been failing `main` too since #880). Nothing in this branch touches that path, and a rerun failed identically, so I merged `main` and cherry-picked #911's commit (d4a376d) onto this branch to unblock it. Both bulk-submit tests pass locally against a fresh testcontainer. When #911 merges, the identical change resolves cleanly.

smunini (approved):

PR #877, opened 2026-09-02
fix(persistence,rest): pre-size bulk-submit byte progress via HEAD

### LINKED ISSUE #874: ui/rest: byte-based bulk progress sawtooths on multi-file manifests — the denominator grows as each file opens
Observed live during the #448 Postgres leg, on the first multi-file manifest run since #849: with 20 output files, `X-Progress` climbed to 99%, dropped to 54%, climbed again, dropped to 34%, and so on:

```
processing 99% complete (1216 entries ingested)
processing 54% complete (1618 entries ingested)
processing 95% complete (2418 entries ingested)
processing 34% complete (2716 entries ingested)   ← next file opened
```

Cause: `bytes_total` accumulates lazily — each file's size is learned from `Content-Length` only when its stream opens, so early percentages are computed against a partial denominator and every newly opened file yanks the bar backwards. Single-file manifests (the #848 case) are unaffected.

Options, cheapest first:
1. **Monotonic display**: never report a percentage lower than the previous one for a non-terminal manifest (server- or UI-side clamp). Cheap, but the bar stalls instead of regressing.
2. **Pre-size the manifest**: HEAD every output file at manifest-fetch time and set `bytes_total` once before ingesting; fall back to lazy accumulation when HEAD is unsupported. Honest percentages the whole run.

Given #849's goal was "the bar fills and never lies", option 2 with the fallback looks right.

angela-helios:
Closes #874

Stacked on #849 (which now carries #851's merge); merge that one first.

## What

On a multi-file manifest the byte-progress denominator was learned lazily — each file's size arrived only when its stream opened — so the percentage kept recomputing against a partial total and every newly opened file yanked the bar backwards. Observed live on the #448 Postgres leg's 20-file run: `99% → 54% → 95% → 34% → …`.

`SubmitInputFetcher` gains `file_size()` — a best-effort HEAD returning the advertised `Content-Length`, with a default implementation returning `None` so every existing mock keeps compiling — and the worker sums all output files up front, storing the complete denominator before ingestion starts. Fallbacks stay honest:

- any file with an unknown size → lazy accumulation exactly as before;
- encrypted (JWE) submissions skip pre-sizing entirely (wire length ≠ decrypted length);
- HEAD requests deliberately skip `Accept-Encoding: gzip` so the size matches the decompressed bytes the ingest counts.

New worker test: a two-file manifest lands `bytes_total == bytes_processed == sum` of both files through the pre-sized path. Suites: worker units (6), rest `bulk_submit` (22) + `bulk_submit_jwe` (6).

### DIFF
--- a/crates/persistence/src/core/bulk_submit_input.rs
+++ b/crates/persistence/src/core/bulk_submit_input.rs
@@ -93,6 +93,26 @@ pub trait SubmitInputFetcher: Send + Sync {
         oauth_metadata_urls: &[String],
         encryption_key: Option<&Value>,
     ) -> StorageResult<(Box<dyn AsyncBufRead + Send + Unpin>, Option<u64>)>;
+
+    /// Returns the advertised size in bytes of the file at `url` without
+    /// opening its body, or `None` when the source cannot cheaply say (no
+    /// HEAD support, no `Content-Length`, or a JWE file whose decrypted
+    /// length differs from the wire length).
+    ///
+    /// The worker calls this for every manifest file up front so the byte
+    /// progress denominator is complete before ingestion starts — learned
+    /// lazily per file, each newly opened file yanks the percentage
+    /// backwards (#874). Best-effort: any `None` (the default) falls back
+    /// to lazy accumulation.
+    async fn file_size(
+        &self,
+        _url: &str,
+        _request_headers: &[(String, String)],
+        _requires_access_token: bool,
+        _oauth_metadata_urls: &[String],
+    ) -> StorageResult<Option<u64>> {
+        Ok(None)
+    }
 }
 
 /// Maps a submission to the stable [`ExportJobId`] used as the output-store key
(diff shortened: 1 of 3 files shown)


angela-helios:
Follow-up pushed (0f716d4): live testing on the #448 MongoDB leg caught a bug in the first commit. reqwest's Response::content_length() is the **body** size hint — a HEAD response has no payload, so it reported 0 for every file regardless of the Content-Length header. The presize summed to zero and, worse, the presized flag then suppressed the lazy per-file fallback: byte progress silently vanished on every backend (bytes_processed counted fine, bytes_total stuck at 0). The size now comes from the Content-Length header itself, and a zero-sum presize no longer claims the denominator. Re-verified live: single-file manifest on SQLite lands bytes_total == bytes_processed == file size, and the multi-file MongoDB run reports a monotone percentage.

smunini (approved):

Issue #814, opened 2026-08-31
Expose Elasticsearch refresh control (refresh=wait_for, configurable refresh_interval) for read-after-write search


smunini:
## Motivation

In the composite Elasticsearch storage modes (`sqlite-elasticsearch`, `postgres-elasticsearch`, `mongodb-elasticsearch`, `s3-elasticsearch`), ES is a secondary search index, so a write can return `201`/`200` before the resource is findable via search. Raised in discussion #704.

There are two independent sources of that lag, and only one of them is tunable today:

1. **Composite write -> ES indexed.** Tunable via `HFS_COMPOSITE_SYNC_MODE` (`crates/hfs/src/main.rs:130`, documented in `crates/hfs/README.md:95`). The default is `asynchronous`, which enqueues the sync event and returns; `synchronous` awaits the secondary write (`crates/persistence/src/composite/sync.rs:315`).
2. **ES indexed -> ES searchable.** Not tunable. Indexing goes through the `index` API with no `refresh` parameter, and the index-level `refresh_interval` defaults to `1s` and is never overridden from configuration.

Consequence: even with `HFS_COMPOSITE_SYNC_MODE=synchronous`, a client that writes a resource and immediately searches for it can miss it for up to the refresh interval. There is currently no setting a deployment can turn to get read-after-write search semantics.

## Proposed behavior

Expose Elasticsearch refresh control so a deployment (and ideally an individual request) can trade write latency for read-after-write search visibility:

- A configuration knob for the ES index `refresh_interval` (e.g. `HFS_ELASTICSEARCH_REFRESH_INTERVAL`), so it is no longer pinned to the `ElasticsearchConfig` default.
- A configuration knob selecting the `refresh` parameter applied to ES index/delete operations: `false` (current behavior), `wait_for`, or `true`. `wait_for` is the interesting one — it blocks the index call until the affected shards refresh, without forcing an extra refresh per write.

## Evidence

- `crates/persistence/src/backends/elasticsearch/backend.rs:62-64` — `refresh_interval` defaults to `"1s"` via `default_refresh_interval()` at line 100; applied to the index settings in `crates/persistence/src/backends/elasticsearch/schema.rs:27`.
[...]

PR #885, opened 2026-09-02
feat(persistence): expose configurable Elasticsearch write refresh 

bomanaps:
Adds HFS_ELASTICSEARCH_REFRESH_INTERVAL and HFS_ELASTICSEARCH_WRITE_REFRESH (false/wait_for/true) so ES-backed deployments can get read-after-write search visibility when combined with HFS_COMPOSITE_SYNC_MODE=synchronous, closing #814.


smunini (approved):
Reviewed the implementation and tests; wiring covers all four ES-backed startup paths and every per-document index/delete call. Moved the consistency explanation to the persistence README and paired wait_for with synchronous sync in the CI workflows on top of this branch. Approving.

PR #887, opened 2026-09-02
ui: make SQL Export list-first with per-user job cards (#833)

### LINKED ISSUE #833: ui: SQL Export list-first — Active SQL Exports page, job cards and a per-user job store
Follow-up of #756 (drafts approved in the 2026-09-01 review). Track **A — SQL Export**, step 1 of 5.

**Design reference:** [SQL on FHIR UI Drafts](https://claude.ai/code/artifact/8dbf81a5-9bea-4caf-9768-a94071dc735e) → page *Approved*, artboards **"SQL Export — Active SQL Exports (entry page)"**, **"… dark"**, **"… empty"**. Every artboard is real HFS markup rendered with `crates/ui/assets/app.css`, so the classes you see are the classes to use.

## Summary

Make `/ui/sql/export` **list-first**, exactly as Bulk Export became in `ui: make Bulk Export list-first`: entering SQL Export lands on an *Active SQL Exports* page that lists the user's `$sql-export` jobs as job cards; the create form moves to `/ui/sql/export/new` (#834). Today the page follows one job through `?job=` in the URL and forgets it on the next navigation.

## Requirements

1. **Routes.** `GET /ui/sql/export` renders the list. `GET /ui/sql/export/new` renders today's create form (restyled in #834). Keep `POST /ui/sql/export` as the kick-off; on success redirect to the list, not to `?job=`.
2. **Per-user job store.** Persist the user's jobs in the per-user settings document under `byTenant.<tenant>.sqlExport.jobs`, object-keyed by job id, mirroring `bulkExport.jobs` (`crates/ui/src/bulk_export.rs`). A record holds: name (optional), subjects `[{name, reference, kind}]` where kind ∈ ViewDefinition | SQL Query | SQL View, format, status (`in-progress` | `complete` | `failed` | `cancelled`), poll URL, progress text, error, startedAt, finishedAt, manifest outputs once finished. `$sql-export` has no server-side job list, so this store is the only list.
3. **Job card** — one `.card.job-card` per job, structure of `partials/bulk_export_card.html`: name; `.job-card__actions` with the contextual action + `.tag.tag--{status}` chip + a `…` overflow (`<details class="menu">` with `.menu__panel--right`, as #733); full-width `.progress.progress--{status}`; **one** `.job-card__meta` line that names the **subject kinds and the format** — e.g. `2 of 6 subjects (4 ViewDefinitions · 1 SQL Query · 1 SQL View) · CSV · started 09:12` — so the kind of thing being exported is visible without opening the job.
4. **Actions.** In progress: *Cancel* (plain form POST). Failed: *Retry*. Complete: *View files* (→ #835; until #835 lands, link to today's `/ui/sql/files?job=`). Overflow: *Open*, *Run again* (re-submits the stored subjects + format + parameters), *Copy job id*, *Remove from list*. Retry / Run again create a new job record.
5. **Status.** Poll the job's status URL as Bulk Export does. `202` → in progress with the `X-Progress` text; `303` → fetch the manifest and derive complete vs failed from it; `404` → unknown → treat as cancelled/reclaimed. In-progress cards poll their own fragment every 5 s over htmx (`hx-get … hx-trigger="every 5s" hx-swap="outerHTML"`); without JavaScript the page is a plain reload.
6. **Header.** `.page-head--back-link` shape with the lede `N exports · M running` and a primary **New SQL Export** button (→ `/ui/sql/export/new`). Empty state: `.card.notice` "No SQL exports yet. Start one with New SQL Export."
7. **Light and dark** both work (tokens only; no literal colors).
8. **i18n:** every new string in `locales/{en,es,de}/main.ftl` (the i18n test requires identical key sets).

## Out of scope

The create form itself (#834), the job detail page (#835), advanced parameters (#836, #837).

## Acceptance criteria

- [ ] `GET /ui/sql/export` lists the caller's jobs for the current tenant; jobs of another user or tenant never show.
- [ ] Kicking off an export adds a card in progress; when the server finishes, the card turns `complete` (or `failed` with the error text) without a manual refresh when JS is on, and on reload when it is off.
- [ ] Cancel, Retry, Run again, Remove from list work as plain forms (nojs Playwright project passes).
- [ ] The meta line shows the subject-kind breakdown and the format for every status.
- [ ] Rust tests in `crates/ui/tests/` cover the store (add/update/remove, tenant isolation) and the status transitions; Playwright covers the list, polling and the empty state; `design-system.spec.ts` passes (no undefined classes).
- [ ] `crates/ui/README.md` documents the new routes and the `sqlExport.jobs` settings key.

## Pointers

`crates/ui/src/bulk_export.rs` (store, list-first routes, card fragment, cancel/retry/delete), `crates/ui/templates/pages/bulk-export-active.html`, `crates/ui/templates/partials/bulk_export_card.html`, `crates/ui/src/conformance.rs` (`sql_export_start/status/cancel/manifest`), `crates/ui/src/sql_export.rs` (manifest parsing), `crates/ui/templates/pages/sql-export.html` (today's page).

## Design reference (screenshots)

Full-page renders of the approved artboards (branch `docs/756-design-references`, `docs/images/sql-ui-drafts/`). The canvas linked above is the editable source.

**Active SQL Exports — light**

Active SQL Exports — light

**Active SQL Exports — dark**

Active SQL Exports — dark

**Active SQL Exports — empty**

Active SQL Exports — empty

andyguz17:
## Summary

Makes `/ui/sql/export` list-first, mirroring the Bulk Export page: the entry page lists the
current user's `$sql-export` jobs as cards (status chip, real server progress, subjects and
format, actions), and the creation form moves to `/ui/sql/export/new`.

- **Caller identity on `$sql-export` self-calls.** The four `$sql-export` methods of
  `ConformanceSource` (start, status, cancel, manifest) now receive a `Caller` (tenant +
  the browser's `Authorization`). A user bearer is forwarded verbatim; without one the
  configured outbound service token is used, as before. `SqlExportStatus::Unavailable`
  distinguishes "the poll could not be answered" from "the server does not know this job".
- **Per-user job store.** Jobs live in the user's settings document under `sqlExport.jobs`
  (per tenant) with optimistic locking and bounded retries. The server's export controller
  is in-memory and has no listing, so the list is the client's own record: a running job the
  server no longer knows becomes `cancelled` with a reason, and manifest outputs are kept
  on the record so a card stays useful after the reaper. Jobs started through the API do not
  appear (documented in the crate README).
- **Honest polling.** `202` → in progress with the server's `X-Progress`; `303` → manifest
  (`complete`, or `failed` with the OperationOutcome message); `404` → `cancelled`;
  network/auth/5xx errors keep the job `in-progress` with an "unavailable" note until the
  next poll. With JS each running card polls its own fragment every 5 s; without JS the list
  page refreshes running jobs on render.
- **Per-job actions.** Cancel, Retry (failed only), Run again, Remove from list, and Copy job
  id (progressive enhancement; the overflow menu is hidden without JS when it would be
  empty).
- **SQLite fix.** `user_settings` read-modify-write opened a DEFERRED transaction; under WAL a
  concurrent writer made the write-upgrade fail with "database is locked" without honouring
  `busy_timeout`. Both transactions now open `IMMEDIATE`, like `storage.rs` already does, with
  a concurrency regression test.
- **Download links** from `$sql-export` manifests are rendered same-origin (path only), so
  they work when the server does not run on the `HFS_BASE_URL` default.
- **E2E.** Playwright coverage for the list (empty state, polling to completion, actions,
  no-JS lifecycle, design-system and a11y sweeps over the new routes) plus cleanup of the
  resources the SQL and theme specs seed on the shared e2e server.

Closes #833

## Test plan

- [x] `cargo test -p helios-ui` (343 tests) and `cargo test -p helios-persistence --features sqlite --test sqlite_user_settings_concurrency`
- [x] `cargo test -p helios-rest --test sof_export` (mixed ViewDefinition + Library subjects in one manifest)
- [x] `cargo clippy` with the CI allow-list and `cargo fmt --check`
- [x] Playwright `chromium` and `nojs` projects (full suites)
- [x] Manual walkthrough in a real browser: empty list, create → complete, failed → retry,
  overflow menu, cancel, dark mode, ES/DE, no-JS, server restart, Bulk Export regression check


crates/persistence/src/backends/sqlite/user_settings.rs
```diff
@@ -31,8 +31,16 @@ impl SqliteBackend {
...
+        // current row before it writes the new one. Under WAL a deferred
+        // transaction takes a read snapshot on that first read and then fails
+        // the read-to-write upgrade with SQLITE_BUSY_SNAPSHOT if another
+        // connection committed in between - the busy handler (and its
+        // configured busy_timeout) is never invoked for that failure. Taking
+        // the write lock up front avoids it, same reasoning as
+        // `purge_tenant_data`/`PurgableStorage::purge` in storage.rs.
         let txn = conn
-            .transaction()
+            .transaction_with_behavior(rusqlite::TransactionBehavior::Immediate)
```
smunini, line 43: What about the other backends?  This change is only for sqlite.

smunini (changes requested):

crates/persistence/src/backends/sqlite/user_settings.rs
andyguz17: This one is SQLite-only: under WAL a deferred transaction fails the read-to-write upgrade without honoring busy_timeout. The other backends already serialize this by design (Postgres FOR UPDATE, Mongo version-conditioned update, S3 If-Match on the ETag), so nothing to change there.

But your question surfaced a different bug in Postgres: FOR UPDATE has no row to lock before the document exists, so two concurrent first writes for the same user can silently overwrite each other. Pushing a fix here (pg_advisory_xact_lock on the user key) plus a concurrency test for Postgres.

smunini (approved):

PR #904, opened 2026-09-02
perf(bulk-submit): fast-load mode — defer search indexing to a post-manifest reindex

### LINKED ISSUE #883: search(sqlite-es): _sort by search parameter is silently ignored — 200 with unsorted results
Found in the #448 Elasticsearch leg. On the `sqlite-es` composite, `Patient?_sort=birthdate` returns `200` with the parameter echoed in the self link — and the results in arbitrary order. Same for `-birthdate`.

Backend parity for `_sort` by an ordinary indexed parameter now reads:

| backend | behavior |
|---|---|
| SQLite | sorts correctly |
| PostgreSQL | sorts correctly |
| MongoDB | rejects loudly with 400 (#881) |
| sqlite-es | **accepts and silently ignores** |

The silent variant is the sharp edge — same class as the `_tag`/`_profile`/`_security` silent-drop cluster from the SQLite leg: a client sorting a clinical worklist gets an unsorted list with no signal anything was wrong.

### LINKED ISSUE #903: perf(bulk-submit): SQLite ingest is ~100-300 KB/s — a full 10k-patient Synthea export (14.7M resources, 29.6 GB) needs days; add a bulk fast-load mode
## The report

"Importing 10,000 Synthea patients takes far too long — SQLite should do it in under half an hour" (review call, 2026-09-02). Reproduced against the full Synthea 10k-patient export served locally.

## First finding: the dataset is 1,400x bigger than its name

A full Synthea export of 11,403 patients is **14,709,697 resources / 29.6 GB of NDJSON** across 24 files — Observation alone is 5.75M resources (5.7 GB), Claim.ndjson is 2.9 GB. "10k patients" undersells it by three orders of magnitude; expectations should be set against ~15M resources.

## Measured (release build of current main, SQLite, local file server)

Phase timings — none of the suspects from the call are the problem:

| phase | measured |
|---|---|
| kickoff | 91 ms |
| manifest fetch + parse | instant (3.1 KB) |
| pre-size HEAD pass (24 files) | < 1 s, total known up front (#877) |
| file download | streamed, not buffered (
| ingestion | **the entire cost — see below** |

Ingestion rate over the first ~4 minutes (1s ground-truth sampling of `bulk_manifests`):

```
t=15s   4,700 entries  313/s   (AllergyIntolerance, ~950 B/resource)
t=60s  11,176 entries  125/s
t=120s 15,776 entries   67/s   (CarePlan, ~1.9 KB/resource)
t=225s 20,776 entries   40/s
```

The apparent decay is a **per-byte cost meeting bigger resources**: 313/s × 950 B ≈ 300 KB/s and 45/s × 1.9 KB ≈ 90 KB/s — the engine processes **~100–300 KB of NDJSON per second**, flat. At that rate 29.6 GB needs **~2–3 days**. Hitting the 30-minute expectation needs ~16 MB/s — roughly **100×**.

## What it is *not*

- **Not lock contention**: one worker owns the manifest; ingestion is one IMMEDIATE transaction per 100 entries (#815); status polls are WAL readers and never block the writer. During the run the WAL held steady at 7.8 MB with healthy checkpoints — no lock churn, no WAL starvation.
- **Not the UI**: the byte-based bar (#849/#877) reports honest progress against the pre-sized 29.6 GB total. (On a dataset this size even honest progress crawls: 1% ≈ 40 minutes — nothing "hangs", it is simply that slow.)
- **Not download buffering**: files stream through a counting BufReader; memory stays flat.

## Where the per-byte cost goes

Per resource, inside the batch transaction: JSON parse → resource + history rows → **search-parameter extraction and 15–30 `search_index` rows** → **FTS upsert over the full serialized content** → receipt + rollback rows. All batched (one fsync per 100), so the remaining cost is CPU + b-tree page writes — single-threaded, as SQLite wants.

Direct evidence for where the headroom is: on the `sqlite-es` composite, where the primary **skips local search indexing and FTS**, the same engine ingests at **~1,000 resources/s** (#882 verification) — about **10× faster** than with indexing inline.

## Proposed direction

1. **Bulk fast-load mode** (biggest win, ~10×): during `$bulk-submit`, write only resources + history + receipts; skip `search_index` and FTS; run the existing `$reindex` machinery per ingested type at manifest completion. Opt-in via config (e.g. `HFS_BULK_SUBMIT_DEFER_INDEXING`), with the status manifest noting the reindex phase.
2. **Parallelize parse/extract** across worker threads with a single writer thread applying batches (SQLite stays single-writer; the CPU half parallelizes) — estimated 2–4× on top.
3. **Debug-level phase instrumentation** in the worker (per-file: download ms, parse ms, index ms, commit ms) so future reports self-diagnose — *debug* level per the call, to keep production logs clean.
4. For datasets in the tens of GB, PostgreSQL post-#880 plus a COPY-based path is the honest long-term answer; SQLite at 16 MB/s of indexed FHIR ingest is not realistic.

With 1+2, the 29.6 GB export lands in the ~2–4 hour range plus a reindex pass; the 10k-patient *demo-sized* subsets (tens of MB) become minutes.

angela-helios:
Closes #903
Closes #883

## Context

From the 2026-09-02 review call: importing "10,000 Synthea patients" was taking far too long. The full export is actually **14.7M resources / 29.6 GB** of NDJSON, and phase measurement (#903) put the entire cost in per-entry search-parameter extraction, `search_index` writes, and FTS — all inline with the batch transaction, at **~100–300 KB/s**. Not lock contention (single worker, batched transactions from #815, WAL steady at 8 MB), not download buffering, not the manifest fetch.

## What

`HFS_BULK_SUBMIT_DEFER_INDEXING=true` enables **fast-load**:

- Entries ingest **without** search-index or FTS writes (`TransactionOptions.defer_search_indexing`, honored by the SQLite and PostgreSQL transaction paths — on PG it composes with the council's buffered multi-row flush by simply building no index rows).
- When each manifest finishes, the worker fires a `DeferredReindexHook` that starts the same background reindex `POST /$reindex` runs, restricted to the manifest's resource types. The hook (`ReindexOnFinish`) **shares the server's `ReindexOperation` instance**, so `$reindex-status` reports the rebuild and audit events flow as usual. All server modes wire it; if a deployment has no reindex machinery (S3 standalone) the worker logs that `$reindex` is owed.

**Nothing is omitted from the data**: resources, history, per-line receipts, and rollback records are written identically — only the *derived* search structures are rebuilt afterwards instead of inline, by the existing reindex path (which rebuilds dynamic entries, `_contained` entries, **and** the FTS row — the same extractor as the write path). Stale index rows of updated resources are still deleted during deferred ingestion, so the index can miss a resource but never mislead about one. Reads/vreads/history are complete throughout; a manifest's resources become searchable when its reindex lands.

## Measured (same 29.6 GB manifest, release build, SQLite)

| mode | rate | 29.6 GB ETA |
|---|---|---|
| inline indexing (today's default) | ~40–90 entries/s (~0.1–0.3 MB/s) | **~2–3 days** |
| **fast-load** | **~1,500–2,000 entries/s (~3.0–4.0 MB/s)** | **~2.5 hours** + reindex |

That is a **~25–30× byte-throughput improvement**, sustained across the first 360k entries. The default stays `false` — inline indexing remains the safe default for small interactive imports.

## Tests

- Worker round-trip: deferred ingest is readable but unsearchable, the hook fires with the manifest's types, and the **real** reindex machinery (`ReindexOnFinish` over `ReindexOperation`) restores searchability.
- Full persistence lib (822) and rest `bulk_submit` (22) suites green.

cc the question set sent to @mauripunzueta — his benchmark harness is the natural place to validate this against the council's write-canary metrics.


angela-helios:
Measured the reindex half on a 10k rich-Synthea sample: fast-load ingested it in **6 seconds** (~1,700/s), the auto-reindex fired on manifest completion as designed and rebuilt all 10,010 resources — at **~50–100 resources/s**, decaying. Extrapolated to the full 14.7M-resource export, the background reindex would take ~2 days: the rebuild path (`write_search_entries`) is per-resource autocommit, the exact shape #815 removed from ingestion.

So the honest ledger for the 29.6 GB manifest today: **~2–2.5 h to fully ingested and readable** (the import itself), with search filling in progressively behind a slow reindex. The natural follow-up is applying the same per-batch transaction treatment to the reindex writer — that should put the rebuild in the same ~1,000+/s class and the whole end-to-end inside a working day. Filing it as the next work item unless review prefers it folded into this PR.

angela-helios:
Second commit (71bcee5): the reindex rebuild now runs page-per-transaction (with a per-resource-outcome page method on ReindexTarget; ES/Mongo keep the per-resource default), the driver no longer runs a duplicate extraction pass just to count entries, and schema v19 adds the keyset index fetch_resources_page always needed — without it every page re-sorts the whole type, which is quadratic on a 5.7M-row Observation table.

Honest measurement on the 10k rich-Synthea sample: rebuild went 100s → **78s**. Batching removed the fsyncs but the floor is **per-resource CPU** — search-parameter extraction plus FTS tokenization — which no transaction shape can remove. The full ledger for the 29.6 GB export as of this PR:

| phase | time |
|---|---|
| ingest (fast-load) | **~2–2.5 h**, data fully readable |
| background rebuild | ~1.5–2 days at ~70–130/s, search fills in progressively |

The remaining lever is parallelizing extraction across cores (proposal 2 in #903): the writer stays single-threaded on SQLite, but the CPU half shards cleanly — estimated 4–8× on typical hardware, which would put the complete end-to-end (ingest + searchable) inside a working day. Filed as the follow-up rather than grown into this PR.

angela-helios:
Housekeeping note: the first commit also carries the **#883 fix** — Elasticsearch `_sort` by an ordinary parameter silently returned unsorted results because the sort clause assumed every parameter lived in the `search_params.string` nested group; it now targets the group matching the parameter's type (date/number/quantity/token/reference/uri), with FHIR min/max multi-value semantics and the string group as the untyped fallback. Three unit tests pin the clause shapes (28/28 in the module), and live verification on `sqlite-es`: `Patient?_sort=birthdate` and `-birthdate` both return correctly ordered results. Declared in the PR description so #883 closes on merge.

PR #917, opened 2026-09-02
fix: MongoDB ifNoneExist search correctness

### LINKED ISSUE #709: Optimize MongoDB If-None-Exist logic
## Summary

MongoDB transaction Bundle handling currently appears to implement `If-None-Exist` conditional creates by loading all active resources of the requested resource type for the tenant and evaluating the conditional search parameters in application memory.

This bypasses the existing `search_index` collection and its SearchParameter-specific indexes.

For tenants with a large number of resources of a given type, this makes each conditional create potentially O(N) in the number of resources of that type and could become particularly expensive when `If-None-Exist` is used frequently during high-volume transaction Bundle ingestion.

## Current behavior

`find_matching_resources_in_bundle_transaction` queries `resources` using only:

```rust
.find(doc! {
    "tenant_id": tenant_id,
    "resource_type": resource_type,
    "is_deleted": false,
})
.session(&mut *session)
```

It then collects the returned documents:

```rust
let docs = collect_session_documents(cursor, session).await?;
```

and evaluates the parsed conditional search parameters against those resources in application code.

For a request such as:

```http
POST Patient
If-None-Exist: identifier=https://example.com/mrn|12345
```

a tenant with 1 million active Patient resources could therefore require HFS to retrieve and inspect a very large portion of those Patient documents simply to determine whether the identifier already exists.

## Existing search infrastructure

MongoBackend already maintains the `search_index` collection, including indexes designed for the corresponding FHIR search types.

For example, token searches can use an index shaped like:

```javascript
{
    tenant_id: 1,
    resource_type: 1,
    param_name: 1,
    value_token_system: 1,
    value_token_code: 1
}
```

The search-index writes/deletes are also already performed using the same Mongo `ClientSession` during transaction Bundle processing.

It therefore seems possible for conditional matching to query `search_index` within the existing transaction rather than enumerating the underlying resources.

For the identifier example above, the lookup could conceptually target:

```javascript
{
    tenant_id: tenant_id,
    resource_type: "Patient",
    param_name: "identifier",
    value_token_system: "https://example.com/mrn",
    value_token_code: "12345"
}
```

and then resolve the resulting `resource_id` values.

## Impact

This is potentially significant for workloads that:

* contain hundreds of thousands or millions of resources of the same type per tenant;
* make heavy use of `If-None-Exist`;
* ingest transaction Bundles concurrently; or
* have high concurrent read/write activity.

Besides the scan itself, retrieving and filtering the resources in HFS can increase Mongo reads, network traffic, application memory/CPU usage, and transaction duration.

## Suggested improvement

Could `find_matching_resources_in_bundle_transaction` reuse the normal SearchParameter/search-index query machinery while executing those queries through the active `ClientSession`?

Ideally, conditional matching within a transaction would preserve the existing transaction semantics while using the same indexed access paths as normal FHIR searches.

If there is a reason the regular `search_index` cannot safely be used for conditional matching within the transaction, it would also be useful to understand that constraint.

bomanaps:
Fixes #709  replace MongoDB's broken in-Rust ifNoneExist matcher with the real session-scoped search machinery so conditional bundle entries match correctly.

### DIFF
--- a/crates/persistence/src/backends/mongodb/search_impl.rs
+++ b/crates/persistence/src/backends/mongodb/search_impl.rs
@@ -807,7 +807,7 @@ impl MongoBackend {
         Ok(Some(ids))
     }
 
-    fn build_search_index_filter(
+    pub(super) fn build_search_index_filter(
         &self,
         tenant_id: &str,
         resource_type: &str,
@@ -1171,7 +1171,7 @@ impl MongoBackend {
         }
     }
 
-    fn build_resource_filter(
+    pub(super) fn build_resource_filter(
         &self,
         tenant_id: &str,
         resource_type: &str,
@@ -1423,7 +1423,7 @@ impl MongoBackend {
         Ok(result.resources.items)
     }
 
-    fn build_search_parameters(
+    pub(super) fn build_search_parameters(
         &self,
         tenant: &TenantContext,
         resource_type: &str,
--- a/crates/persistence/src/backends/mongodb/storage.rs
+++ b/crates/persistence/src/backends/mongodb/storage.rs
@@ -18,7 +18,7 @@ use crate::core::{
     HistoryEntry, HistoryMethod, HistoryPage, HistoryParams, InstanceHistoryProvider,
     PurgableStorage, ResourceStorage, SettingsStore, SystemHistoryProvider, TypeHistoryProvider,
     VersionedStorage, bundle_if_match_gate, bundle_if_none_exist_gate, if_match_field_satisfied,
-    normalize_etag, not_supported_entry,
+    normalize_etag,
 };
 use crate::error::{
     BackendError, ConcurrencyError, QueryErrorExt, ResourceError, StorageError, StorageResult,
@@ -2686,13 +2686,6 @@ impl MongoBackend {
                     })?;
 
                 if let Some(search_params) = entry.if_none_exist.as_ref() {
-                    if self.is_search_offloaded() {
-                        return Ok(not_supported_entry(
-                            "ifNoneExist cannot be resolved inside a transaction when search \
-                             is offloaded to a secondary backend; submit the entry in a batch \
-                             Bundle instead",
-                        ));
-                    }
                     let matches = self
                         .find_matching_resources_in_bundle_transaction(
                             db,
@@ -3335,11 +3328,17 @@ impl MongoBackend {
             return Ok(Vec::new());
         }
 
+        if self.is_search_offloaded() {
+            return self
+                .if_none_exist_offloaded_scan(db, session, tenant, resource_type, &parsed_params)
+                .await;
+        }
+
         let typed_params = self.build_search_parameters(tenant, resource_type, &parsed_params);
         let query = SearchQuery {
             resource_type: resource_type.to_string(),
             parameters: typed_params,
-            count: Some(1000),
+            count: Some(2),
             ..Default::default()
         };
 
@@ -3356,6 +3355,7 @@ impl MongoBackend {
             let pipeline = vec![
                 doc! { "$match": filter },
                 doc! { "$group": { "_id": "$resource_id" } },
+                doc! { "$limit": 1001_i32 },
             ];
             let cursor = search_index
                 .aggregate(pipeline)
@@ -3368,6 +3368,16 @@ impl MongoBackend {
                     ))
                 })?;
             let docs = collect_session_documents(cursor, session).await?;
+
+            if docs.len() > 1000 {
+                tracing::warn!(
+                    resource_type,
+                    param_name = %param.name,
+                    "ifNoneExist criteria match more than 1000 search_index entries; \
+                     intersection may be incomplete for this parameter"
+                );
+            }
+
             let ids: HashSet<String> = docs
                 .into_iter()
                 .filter_map(|d| d.get_str("_id").ok().map(str::to_string))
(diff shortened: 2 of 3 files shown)


TheCaffinatedDeveloper:
1. Unbounded parameter results
collect_session_documents() still materializes every distinct matching resource ID before intersecting in Rust. For a broad filter such as Patient?active=true&identifier=ABC123, the active=true aggregation could return tens of thousands of grouped IDs and build a correspondingly large HashSet.
Could this use bounded candidate paging instead? One approach is to probe each parameter to choose a selective driver, page that filter by search-index _id, and intersect the remaining filters using resource_id: {$in: <bounded batch>}. Conditional create only needs zero, one, or more than one match.

2. Final resource fetch is also unbounded
query.count = Some(1000) does not limit this find(). build_resource_filter() only constructs predicates, and the resulting cursor is collected without calling .limit().
Since ifNoneExist only distinguishes 0, 1, and multiple matches, please limit resource materialization to two documents and stop immediately once the second live resource is found.

3. Offloaded-search regression
Returning not-supported whenever search is offloaded changes existing transaction-bundle behavior. Elasticsearch cannot see uncommitted writes, but a later bundle entry still needs to match a resource created earlier in the same transaction.
A transaction-local fallback should scan MongoDB resources through the active session, stream results, and stop after two matches rather than rejecting the operation.

4. Missing regression coverage
Could we add tests for a broad-first query such as active=true&identifier=system|value, repeated parameters with AND semantics, same-transaction read-your-writes, duplicate index rows, and the two-result cap? This change currently has no tests, and those are exactly the cases where correctness and boundedness can drift.

bomanaps:
Addressed here https://github.com/HeliosSoftware/hfs/pull/917/commits/bc07bdf4cc2961736da7fd06dfb3701bea0b9da5

TheCaffinatedDeveloper:
1. The 1,001-ID cutoff produces false negatives
The $limit: 1001 bounds client memory, but it makes the intersection incomplete. For example, if active=true matches 90,000 patients and the patient matching the identifier is not among the arbitrary first 1,001 grouped IDs, the intersection is empty and conditional create incorrectly creates a duplicate.
Logging that the result may be incomplete does not make that behavior safe. Could we replace the cutoff with bounded candidate paging, using one parameter as the driver and intersecting each candidate batch against the remaining filters?

Also, because $group precedes $limit, MongoDB still groups the entire broad match set before returning 1,001 IDs.

2. Offloaded fallback ignores criteria
if_none_exist_offloaded_scan currently ignores parameters other than _id and identifier, then applies .limit(2). For active=false, it could return arbitrary active patients and incorrectly suppress creation or report multiple matches.
The fallback needs to evaluate every criterion against streamed resources before counting the match. If a criterion cannot be evaluated correctly, returning an explicit unsupported-query error is safer than treating it as matched.

3. Missing regression test
I would add a test with more than 1,001 resources where the selective identifier match falls beyond the broad parameter’s first 1,001 IDs. That should expose the incomplete-intersection bug deterministically.

TheCaffinatedDeveloper:
Could something like this work instead?

This keeps both the application memory and every distinct() response bounded by the candidate batch size. Unlike $limit: 1001 before intersection, paging continues until the complete driver result has been examined or two live matches are found, so broad parameters cannot cause false negatives. The production version should select the driver using a small bounded probe and retain the existing direct handling for _id and _lastUpdated.

```rust
const CANDIDATE_BATCH_SIZE: i64 = 128;

// Select this using a small bounded probe in the full implementation.
let driver_index = 0;
let driver_filter =
    self.build_search_index_filter(tenant_id, resource_type, &query.parameters[driver_index])?;

let mut last_index_id: Option<Bson> = None;
let mut matches = Vec::with_capacity(2);
let mut matched_ids = HashSet::new();

loop {
    let page_filter = match &last_index_id {
        Some(last_id) => doc! {
            "$and": [
                driver_filter.clone(),
                { "_id": { "$gt": last_id.clone() } }
            ]
        },
        None => driver_filter.clone(),
    };

    let mut cursor = search_index
        .find(page_filter)
        .sort(doc! { "_id": 1 })
        .projection(doc! { "_id": 1, "resource_id": 1 })
        .limit(CANDIDATE_BATCH_SIZE)
        .session(&mut *session)
        .await?;

    let mut candidate_ids = HashSet::new();
    let mut documents_read = 0;

    while cursor.advance(&mut *session).await? {
        let document = cursor.deserialize_current()?;
        last_index_id = document.get("_id").cloned();
        documents_read += 1;

        if let Ok(resource_id) = document.get_str("resource_id") {
            candidate_ids.insert(resource_id.to_string());
        }
    }

    if documents_read == 0 {
        break;
    }

    // Intersect only this bounded candidate batch.
    for (index, parameter) in query.parameters.iter().enumerate() {
        if index == driver_index
            || matches!(parameter.name.as_str(), "_id" | "_lastUpdated")
        {
            continue;
        }

        let parameter_filter =
            self.build_search_index_filter(tenant_id, resource_type, parameter)?;

        let bounded_filter = doc! {
            "$and": [
                parameter_filter,
                {
                    "resource_id": {
                        "$in": candidate_ids.iter().cloned().collect::<Vec<_>>()
                    }
                }
            ]
        };

        let matching_ids = search_index
            .distinct("resource_id", bounded_filter)
            .session(&mut *session)
            .await?
            .into_iter()
            .filter_map(|value| value.as_str().map(str::to_string))
            .collect::<HashSet<_>>();

        candidate_ids.retain(|id| matching_ids.contains(id));

        if candidate_ids.is_empty() {
            break;
        }
    }

    if !candidate_ids.is_empty() {
        let remaining = 2 - matches.len();

        let mut cursor = resources
            .find(doc! {
                "tenant_id": tenant_id,
                "resource_type": resource_type,
                "is_deleted": false,
                "id": { "$in": candidate_ids.into_iter().collect::<Vec<_>>() }
            })
            .limit(remaining as i64)
            .session(&mut *session)
            .await?;

        while cursor.advance(&mut *session).await? {
            let document = cursor.deserialize_current()?;
            let resource =
                document_to_stored_resource(&document, tenant, resource_type)?;

            if matched_ids.insert(resource.id().to_string()) {
                matches.push(resource);
            }
        }
    }

    if matches.len() == 2 {
        break;
    }

    if documents_read < CANDIDATE_BATCH_SIZE {
        break;
    }
}

return Ok(matches);
```

bomanaps:
> Could something like this work instead?
> 
> This keeps both the application memory and every distinct() response bounded by the candidate batch size. Unlike $limit: 1001 before intersection, paging continues until the complete driver result has been examined or two live matches are found, so broad parameters cannot cause false negatives. The production version should select the driver using a small bounded probe and retain the existing direct handling for _id and _lastUpdated.
> 
> ```rust
> const CANDIDATE_BATCH_SIZE: i64 = 128;
> 
> // Select this using a small bounded probe in the full implementation.
> let driver_index = 0;
> let driver_filter =
>     self.build_search_index_filter(tenant_id, resource_type, &query.parameters[driver_index])?;
> 
> let mut last_index_id: Option<Bson> = None;
> let mut matches = Vec::with_capacity(2);
> let mut matched_ids = HashSet::new();
> 
> loop {
>     let page_filter = match &last_index_id {
>         Some(last_id) => doc! {
>             "$and": [
>                 driver_filter.clone(),
>                 { "_id": { "$gt": last_id.clone() } }
>             ]
>         },
>         None => driver_filter.clone(),
>     };
> 
>     let mut cursor = search_index
>         .find(page_filter)
>         .sort(doc! { "_id": 1 })
>         .projection(doc! { "_id": 1, "resource_id": 1 })
>         .limit(CANDIDATE_BATCH_SIZE)
>         .session(&mut *session)
>         .await?;
> 
>     let mut candidate_ids = HashSet::new();
>     let mut documents_read = 0;
> 
>     while cursor.advance(&mut *session).await? {
>         let document = cursor.deserialize_current()?;
>         last_index_id = document.get("_id").cloned();
>         documents_read += 1;
> 
>         if let Ok(resource_id) = document.get_str("resource_id") {
>             candidate_ids.insert(resource_id.to_string());
>         }
>     }
> 
>     if documents_read == 0 {
>         break;
>     }
> 
>     // Intersect only this bounded candidate batch.
>     for (index, parameter) in query.parameters.iter().enumerate() {
>         if index == driver_index
>             || matches!(parameter.name.as_str(), "_id" | "_lastUpdated")
>         {
>             continue;
>         }
> 
>         let parameter_filter =
>             self.build_search_index_filter(tenant_id, resource_type, parameter)?;
> 
>         let bounded_filter = doc! {
>             "$and": [
>                 parameter_filter,
>                 {
>                     "resource_id": {
>                         "$in": candidate_ids.iter().cloned().collect::<Vec<_>>()
>                     }
>                 }
>             ]
>         };
> 
>         let matching_ids = search_index
>             .distinct("resource_id", bounded_filter)
>             .session(&mut *session)
>             .await?
>             .into_iter()
>             .filter_map(|value| value.as_str().map(str::to_string))
>             .collect::<HashSet<_>>();
> 
>         candidate_ids.retain(|id| matching_ids.contains(id));
> 
>         if candidate_ids.is_empty() {
>             break;
>         }
>     }
> 
>     if !candidate_ids.is_empty() {
>         let remaining = 2 - matches.len();
> 
>         let mut cursor = resources
>             .find(doc! {
>                 "tenant_id": tenant_id,
>                 "resource_type": resource_type,
>                 "is_deleted": false,
>                 "id": { "$in": candidate_ids.into_iter().collect::<Vec<_>>() }
>             })
>             .limit(remaining as i64)
>             .session(&mut *session)
>             .await?;
> 
>         while cursor.advance(&mut *session).await? {
>             let document = cursor.deserialize_current()?;
>             let resource =
>                 document_to_stored_resource(&document, tenant, resource_type)?;
> 
>             if matched_ids.insert(resource.id().to_string()) {
>                 matches.push(resource);
>             }
>         }
>     }
> 
>     if matches.len() == 2 {
>         break;
>     }
> 
>     if documents_read < CANDIDATE_BATCH_SIZE {
>         break;
>     }
> }
> 
> return Ok(matches);
> ```

Fixed here https://github.com/HeliosSoftware/hfs/pull/917/commits/62afc3e78661339e46afb46161f73ef40f22b3a4

smunini (approved):

PR #932, opened 2026-09-03
perf(search): index only queried parameters and skip unused full-text

angela-helios:
## Why

Profiling the bulk import found the bottleneck is **FHIRPath evaluation of search parameters** — ~90% of per-resource indexing time (not I/O, fsync, or transfer). A handful of parameters dominate: on Patient, `phone` (a `.where()` filter) is 54%; on Observation the composite `code-value` parameters are ~80%.

## What

`HFS_SEARCH_INDEX_PARAMS` restricts search indexing to an allowlist of parameter codes, and skips full-text (`_text`/`_content`) tokenization unless the list names one of them. `_id`/`_lastUpdated` are always kept. **Unset = index every active parameter (current behavior).** Wired for SQLite and PostgreSQL through the shared extractor.

The trade-off is explicit: only listed parameters are searchable.

Also adds `profile_indexing_cost_breakdown` — an `#[ignore]` harness that splits per-resource indexing into tree-conversion / FHIRPath / FTS and ranks the per-parameter cost. Sample resources are real data and are gitignored.

## Measured

On the 31 GB benchmark manifest: **~2.7× ingest throughput** with a realistic ~9-parameter allowlist; more with a tuned list that drops the expensive-but-rarely-queried params.

## Tests

- `index_only_restricts_to_the_allowlist` (extractor unit test).
- Full persistence lib suite green; compiles with `--features postgres`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,7 @@ __pycache__/
 # purpose: they are personal working material, not project documentation, and
 # they should never reach a PR. `tmp/` is the intended home for them.
 /tmp/
+crates/persistence/benches/samples/
+data/submit/
+data/exports/
+exports/
--- a/crates/hfs/src/main.rs
+++ b/crates/hfs/src/main.rs
@@ -545,8 +545,36 @@ async fn create_postgres_backend(
     };
     backend_config.fhir_version = config.default_fhir_version;
     backend_config.data_dir = config.data_dir.clone();
+    backend_config.index_only_params = search_index_allowlist();
 
     Ok(PostgresBackend::new(backend_config).await?)
+}
+
+/// Parses `HFS_SEARCH_INDEX_PARAMS` into a search-index allowlist of parameter
+/// codes. `None` (unset or empty) indexes every active parameter and full-text
+/// content — the default. When set, only the listed parameters are indexed, and
+/// full-text (`_text`/`_content`) indexing is skipped unless the list names one
+/// of them. Evaluating every active parameter's FHIRPath and tokenizing the
+/// whole resource for FTS is the dominant per-resource ingest cost, so a
+/// deployment that queries only a known set of parameters can trade search
+/// coverage for a proportional speed-up.
+fn search_index_allowlist() -> Option<std::sync::Arc<std::collections::HashSet<String>>> {
+    let allowlist = std::env::var("HFS_SEARCH_INDEX_PARAMS")
+        .ok()
+        .map(|raw| {
+            raw.split(',')
+                .map(|c| c.trim().to_string())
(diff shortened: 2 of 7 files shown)


angela-helios:
@smunini — following up on your out-of-the-box concern before deciding what to do with this PR.

First, one clarification in case it changes the picture: **the default behavior is unchanged**. With `HFS_SEARCH_INDEX_PARAMS` unset (the default), every search parameter is indexed exactly as today — download the binary, load data, search works. The restriction only applies when an operator explicitly sets the variable.

That said, your deeper concern is real: an operator who enables this (e.g. for a fast bulk load) and forgets creates a server where searches silently return empty — and even after unsetting it, previously-loaded data stays under-indexed until a `$reindex`. Silent-wrong is worse than slow.

So before iterating, I'd rather ask which direction you want:

1. **Harden and keep**: excluded parameters return an explicit OperationOutcome error on search (never silent zeros), `/metadata` stops advertising them, and startup warns when the parameter set changed and a `$reindex` is owed. This is the Smile CDR / Azure model (`disable_patterns` / selectable search parameters) — both ship it as their primary write-performance knob, always opt-in.
2. **Restrict the scope**: keep the mechanism but only as part of bulk import (e.g. a kickoff parameter that applies to that submission's reindex), never as a server-wide setting an operator can forget.
3. **Drop the env var entirely** and keep only the measured internals (the profiling harness), taking the write-path optimization in a different direction (we've since found the dominant cost is the 15 secondary indexes on `search_index` — partial indexes may recover much of this win with no configuration at all).

Context for the trade-off: this flag measured **2.7×** on ingest because indexing cost is proportional to parameters indexed, and most deployments query only a handful of the ~1,372 spec parameters. But if the configuration risk outweighs that for you, option 3 loses nothing that a schema-level fix can't recover.

Which way do you want it?

angela-helios:
Agreed — closing this without merging.

The measurement work here still points at a configuration-free way to recover most of the win: instrumenting the write path showed **index-row INSERTs into `search_index` are 53% of total import time** (0.40 ms/row), because the table carries **15 secondary indexes and every row updates all of them** — including the ones for value columns that are NULL on that row (a token row still pays the date/number/quantity/string/uri B-trees).

Next step: convert the per-family indexes to **partial indexes** (`WHERE value_X IS NOT NULL`), so each row only maintains the ~4 structures relevant to its type. No settings, no user-facing behavior change, search plans unaffected (comparisons like `value_date >= ?` imply the predicate). Expected to recover a large share of the 2.7× measured here. Will open as its own PR with EXPLAIN QUERY PLAN verification and the search suite as the gate.

Issue #947, opened 2026-09-04
bulk-submit(sqlite): per-entry pipeline dominates ingest (~84%) — instrument and cut the non-index write volume


angela-helios:
## Context

SQLite counterpart of #945, from an instrumented run of the real 31 GB manifest (single worker, the safe default per #942). Binary = #944's partial indexes **plus** per-phase timing counters on the write path (cumulative, dumped every 1,000 resources). Windows host, WAL mode.

**Host profile during ingest: the writer pegs one core at ~99% while the disk sits at ~12 MB/s.** SQLite ingest is CPU-bound on the single writer thread — the exact opposite of PostgreSQL, which is round-trip-bound with idle CPU (#945). The two backends need different levers.

## Measured phase breakdown (real data)

At 29,000 resources into the manifest (CarePlan / AllergyIntolerance segment, ~1.5 KB average resource, 13.8 index rows per resource):

| Phase | Per resource | Share of wall |
|---|---|---|
| Wall clock (52/s in this window) | 19.2 ms | 100% |
| FHIRPath extraction | 0.26 ms | 1.4% |
| search_index row INSERTs | 2.66 ms | 13.9% |
| FTS (exists check + insert) | 0.14 ms | 0.7% |
| **Everything else (unattributed)** | **~16 ms** | **~84%** |

Two conclusions:

1. **#944's partial indexes already paid off on real data**: 0.19 ms per index row, down from 0.40 ms measured with the full-width indexes. Indexing is now a ~14% item, not the dominant cost.
2. **The dominant cost (~84%) is the per-entry pipeline outside indexing**, and it is currently unmeasured. Known occupants, from a row-count audit of the same database:

```
resources                31,555     the resource row
resource_history         31,555     a FULL COPY of the data blob, written on create
bulk_entry_results       30,176     one bookkeeping row per manifest entry
bulk_submission_changes  30,176     one more bookkeeping row per manifest entry
search_index            437,723     ~14 rows per resource (instrumented above)
```

Every ingested entry writes **four table rows** (one of them duplicating the entire resource blob) plus its index rows, and the remainder of the 16 ms budget goes to the typed FHIR serde round-trip, savepoint/transaction bookkeeping, and NDJSON parsing — none of it yet individually timed.

## Proposed work, in order

[...]

PR #952, opened 2026-09-06
perf(sqlite): cut the bulk-submit write path 79% — instrumentation, cached statements, and index entries nothing can use

smunini:
> **Re-based on main after #951 landed** (merge 5640230). The numbers below were measured against `a4068bfd6`, main *before* #951. Re-measured today against current main, three interleaved rounds on the same host:
>
> | | median | database |
> |---|---|---|
> | main + #951 | 1,832 res/s | 1.252 GB |
> | + this PR | **2,654 res/s (+45%)** | **1.048 GB (−16%)** |
>
> #951 and this branch do not stack additively: with the branch applied, adding #951 measures 2,654 vs 2,686 res/s without it — no change beyond run-to-run noise. Once the redundant and full indexes are gone there are fewer index entries per row for its batched INSERT to amortize a statement over. The two are independent and both worth having; the combined figure is the +45% above.

Closes the instrumentation half of #947 and takes the SQLite bulk-submit write path from **1,555 to 2,783 resources/s (+79%)** with a **16% smaller database**, on the real 31 GB Synthea manifest's files.

Stacks cleanly under #951 — that PR batches the `search_index` INSERT statement, this one reduces how many b-tree entries each row costs.

## What's here

**1. Phase instrumentation and a benchmark** (`helios_persistence::perf`, `examples/bulk_submit_bench.rs`)

`HFS_PERF_PHASES=1` turns on per-phase counters across the ingest write path; with it off every call site is one relaxed atomic load and no clock read. The example drives the real `process_ndjson_stream` — the same call the bulk-submit worker makes per manifest file — against local NDJSON on a fresh database, so a before/after pair differs by the write path and nothing else.

#947 measured ~84% of the per-entry budget as unattributed. It is now fully attributed, and on this host the answer is that **the `search_index` INSERT is over half of everything**:

```
                                before      after
ndjson_parse                    0.005 ms    0.005 ms
entry (total)                   0.525       0.287
  entry_read                    0.008       0.008
  create                        0.476       0.243
    create_exists_probe         0.003       0.001
    serialize                   0.002       0.002
    resources_insert            0.009       0.005
    history_insert              0.007       0.004
    index (total)               0.451       0.229
      index_delete              0.013       —        (gone on the create path)
      extract                   0.074       0.069
      search_index_insert       0.346       0.149
      fts                       0.018       0.010
  bookkeeping                   0.035       0.031
commit                          0.088       0.066
batch_overhead                  0.018       0.001
WALL                            0.639       0.362
```

**2. Schema v21 and v22 — index entries no query can use**

Five `search_index` indexes were *full*: every row entered `idx_search_composite`, `idx_search_resource`, `idx_search_reference_display`, `idx_search_quantity_canonical` and `idx_search_contained` whether or not the indexed column was NULL.

- v21 drops `idx_search_resource`. It was `(tenant_id, resource_type, resource_id)` — a strict leftmost prefix of `idx_search_composite`, so it answered no lookup the composite index cannot.
- v22 rebuilds the other three as partial, the same treatment #944 gave the v20 set and skipped for these because v10/v11 added them later. In the benchmark corpus **zero** rows carry a canonical quantity or a contained flag and 150k of 1.18M carry a reference display; those three indexes held 1.18M entries each, for 134 MB.

`idx_search_string_folded` is deliberately left full, with a comment saying why: it is also the only full index leading `(tenant_id, resource_type, param_name)`, and the planner falls back to it for the LIKE-shaped modifier searches (`:text`, `:contains`) whose predicate SQLite cannot prove implies `IS NOT NULL`. Making it partial pushed those onto a `(tenant_id, resource_type)` scan; replacing it with a narrow `(tenant_id, resource_type, param_name)` index made the planner prefer *that* for token searches, and a selective `code=…` lookup went 2.9 → 22.1 ms. Both alternatives were measured and rejected.

**3. Write-path changes**

- **`prepare_cached` on the ingest statements** (+24% on its own). `Connection::execute` compiles its SQL on every call, and the 24-column `search_index` INSERT ran that way ~14 times per resource — it costs more to compile than to run. Statement-cache capacity goes 16 → 64 so the read and search paths sharing these pooled connections cannot evict the hot inserts.
- **No stale-row DELETE on the create path.** `create` cleared the resource's index rows after an existence probe that had just proved there is no resource row for the id — and no resource row means no index rows (soft delete keeps the row; every purge path deletes `search_index` in the same transaction; the foreign key cascades besides).
- **Batch bookkeeping rides the batch transaction.** Manifest status, entry counters and the submission timestamp were three autocommit statements around each batch — three write-lock acquisitions and fsyncs it did not need, and a window where the entries were durable but the counts describing them were not.
- **Time-ordered `change_id`.** `bulk_submission_changes` takes one row per entry keyed by `change_id`; a v4 UUID inserts at a random leaf, v7 at the resident right-hand edge — the reasoning `new_resource_id` already documents.

## Proof

**Write path**, 3 runs each, phase collection off, 86,453 resources (CarePlan 25k, AllergyIntolerance 11,453, Condition 25k, Encounter 25k = 124.2 MB, 13.68 index rows per resource, `batch_size=100`, fresh database):

| | before (a4068bfd6) | after |
|---|---|---|
| ingest | 1,552 / 1,556 / 1,555 res/s | 2,776 / 2,786 / 2,783 res/s — **+79%** |
| database | 1.254 GB | 1.048 GB — **−16%** |
| `HFS_BULK_SUBMIT_DEFER_INDEXING=true` | 11,569 res/s | 18,519 res/s — **+60%** |

**End to end through `hfs`**, real `$bulk-submit` against a 9-type subset of the manifest (185,616 resources, 262 MB fetched over HTTP, full 1,389-parameter registry):

| | before | after |
|---|---|---|
| wall | 121.0 s (1,534 res/s) | 79.5 s (2,336 res/s) — **+52%** |
| database | 2.44 GB | 2.04 GB |
| index rows | 2,230,759 | 2,230,759 (identical) |

Instrumentation overhead when enabled: **not resolved by this harness.** The one measurement behind an earlier "0.7%" was a single run against a three-run median, in sequential rather than interleaved blocks, and its 19 res/s delta sits inside the disabled arm's own 10 res/s spread. Analytically it should be ~17 spans per resource x (two clock reads + two atomic adds) ~= 0.25%. Treat <=0.7% as an upper bound, not a measurement. The *disabled* cost — a couple of relaxed atomic loads per call site, ~0.02% — is roughly 20x below this benchmark's noise floor and cannot be measured with it at all, which is why the instrumentation is compiled out rather than merely switched off (see below).

## Safety

- `EXPLAIN QUERY PLAN` over 15 representative search shapes — token, token `:text`, reference, reference `:text`, string prefix and exact, quantity raw and canonical, date, uri, identifier `:of-type`, `_contained`, composite, delete-by-resource — is byte-identical before and after, on a database carrying 1.18M real rows.
- New tests pin the index layout (`search_index_carries_no_redundant_or_full_value_indexes`), that an upgraded database ends with exactly a fresh one's indexes, that a create writes only its own index rows while an update still clears the previous version's, and that the batch's manifest counters land with the entries they describe.
- `helios-persistence` lib (841), `sqlite_tests`, `search_suite`, `crud_suite`, `transactions_suite`, `versioning_suite`, `multitenancy_suite`, `sqlite_bulk_submit_concurrency`, `backend_capability_contract`, `storage_resolve_tests`, and `helios-rest` lib + `bulk_submit` all pass.

## Measured and rejected

`PRAGMA foreign_keys = OFF` (no measurable difference), a page cache above the 64 MiB from #950 (marginally slower at this size), raising the default `batch_size` (+17% *before* the autocommit statements were folded into the batch, noise after — and leaving it alone avoids lengthening the write-lock hold, #942), and #947's item 5 (the SQLite ingest path is `serde_json::Value` end to end; `serialize` is 0.3%).

#947's item 2, stop double-writing the blob on create, measures **+3.4% and −15% of database** but is not here: every history read path would have to synthesize v1 from the `resources` row, and the same question applies to PostgreSQL and MongoDB. Numbers and the `dbstat` breakdown are in https://github.com/HeliosSoftware/hfs/issues/947#issuecomment-5559709784 for a follow-up issue.

https://claude.ai/code/session_01Cs99Fk1eg7474hYQGrcwqh

### DIFF
--- a/.claude/skills/bulk-data-submit/SKILL.md
+++ b/.claude/skills/bulk-data-submit/SKILL.md
@@ -109,3 +109,17 @@ The backend capability splits into `BulkSubmitIngest` (the synchronous `BulkSubm
   malformed is `400`. Page size `0` disables pagination and yields one manifest with an empty `link`.
 - Status-poll pacing: the `202` advertises `HFS_BULK_SUBMIT_RETRY_AFTER`, and a client that polls past `HFS_BULK_SUBMIT_POLL_RATE_LIMIT` within the window gets `429` plus a `Retry-After` pointing at the end of that window. Buckets are keyed by poll token plus principal, falling back to peer address; the check runs before any job-store work, so throttled polls stay cheap.
 - Cleanup periodically removes status artifacts for submissions whose `updated_at` exceeds `HFS_BULK_SUBMIT_OUTPUT_TTL`.
+
+## Ingest performance
+
+`HFS_BULK_SUBMIT_DEFER_INDEXING=true` relocates search indexing to a post-ingest
+reindex and is by far the biggest lever (~6.7x on SQLite); the stored resources,
+history, receipts and rollback records are identical either way.
+
+To find out where the rest of the time goes, profile the write path with the
+phase counters in `helios_persistence::perf` and the `bulk_submit_bench`
+example — both gated behind `--cfg perf_phases`, which keeps them out of
+released binaries (`ci.yml` builds those with `--all-features`, so a cargo
+feature could not have). See `/test-hfs` for the invocation and for the two
+traps that have produced wrong numbers here (the search-parameter data
+directory, and comparing runs across sessions instead of interleaving arms).
--- a/.claude/skills/test-hfs/SKILL.md
+++ b/.claude/skills/test-hfs/SKILL.md
@@ -58,6 +58,57 @@ uv run pytest python-tests/ -v
 cargo test
 ```
 
+## Profiling the ingest write path
+
+`helios_persistence::perf` carries per-phase timers across the SQLite ingest
+path — NDJSON parse, read-before-write, resource and history INSERT, search
+extraction, index INSERT, FTS, bulk bookkeeping, commit. They answer "where did
+the time go", which a wall-clock rate cannot.
+
+**They are compiled out unless you ask for them, with `--cfg perf_phases`.**
+That is a cfg and not a cargo feature on purpose: `ci.yml`'s `build` job runs
+`cargo build --workspace --all-features --release` and the `release` job
+publishes exactly those artifacts (the Docker images copy them in), so a
+feature would have shipped the instrumentation in every binary. `--all-features`
+cannot turn a cfg on. Same reasoning as `tokio_unstable`.
+
+```bash
+# Rate only — no phase table, nothing compiled in:
+cargo run --release -p helios-persistence --example bulk_submit_bench -- \
+    --limit 25000 /path/to/CarePlan.ndjson
+
+# With the phase table:
+RUSTFLAGS='--cfg perf_phases' \
+  cargo run --release -p helios-persistence --example bulk_submit_bench -- \
+    --limit 25000 /path/to/CarePlan.ndjson
+```
+
+The example prints a reminder if you ask for phases from a binary built without
+the flag, rather than showing a table of zeros. Within a `--cfg perf_phases`
+build, collection is still off until `HFS_PERF_PHASES=1` or
+`perf::set_enabled(true)` — the benchmark does the latter.
+
+`bulk_submit_bench` drives the real `process_ndjson_stream`, the same call the
+bulk-submit worker makes per manifest file, against local NDJSON on a fresh
+database. Useful flags: `--limit` (resources per file), `--batch`, `--defer-index`
+(the `HFS_BULK_SUBMIT_DEFER_INDEXING` path), `--no-phases`, `--keep` (leave the
+database for `dbstat`), `--data-dir`.
+
+### Benchmarking discipline
+
+Two traps, both of which have produced wrong numbers here:
+
+- **Run it from the repo root, or pass `--data-dir`.** The default `data`
+  directory holds `search-parameters-r4.json`; without it the registry falls
+  back to five embedded parameters, index volume drops from ~14 rows per
+  resource to ~2, and the run is meaninglessly fast. The same applies to a real
+  `hfs` benchmark: set `HFS_DATA_DIR`.
+- **Interleave arms; never compare across sessions.** This workload drifts
+  several percent between runs of a machine on different days. Build each arm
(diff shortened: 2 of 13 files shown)


smunini:
Merged main (#951) into the branch — one conflict, in `index_resource_dynamic`, where #951's eight-row INSERT lands in the same function this branch instruments. Kept #951's batched loop verbatim, wrapped it in the `IndexInsert` phase, and moved the `IndexMarshal` sub-phase onto the batch's own value normalisation and parameter building so it still answers the same question (is the index cost ours or SQLite's?). `write_index_entry` keeps its copy of that span: the reindex path still writes a row at a time. Both sides had independently reached for `prepare_cached` on these statements.

Re-measured against current main, three interleaved rounds, same host, same 86,453-resource corpus:

| arm | rounds | median | database |
|---|---|---|---|
| main + #951 | 1,787 / 1,832 / 1,833 | 1,832 res/s | 1.252 GB |
| this PR (branch + #951) | 2,642 / 2,694 / 2,654 | **2,654 res/s** | **1.048 GB** |
| branch without #951 | 2,633 / 2,689 / 2,686 | 2,686 res/s | 1.049 GB |

**+45% over current main, database 16% smaller.**

The third row is the interesting one: with this branch applied, #951 measures no better than its absence (2,654 vs 2,686 — inside run-to-run noise). That is expected rather than a conflict between them. #951 amortizes per-statement overhead across the rows of one resource; this branch removes three of the five b-tree entries each of those rows was paying for, so there is less statement overhead left per unit of real work to amortize. Both are correct changes; their gains overlap, and the combined number is the +45%.

Tests green after the merge: persistence lib (841), `sqlite_tests`, `search_suite`, `crud_suite`, `transactions_suite`, `versioning_suite`, `multitenancy_suite`, `sqlite_bulk_submit_concurrency`, plus clippy with the CI flag set.

https://claude.ai/code/session_01Cs99Fk1eg7474hYQGrcwqh

Issue #882, opened 2026-09-02
bulk-submit(composite): $bulk-submit resources are never indexed into Elasticsearch — bulk-loaded data invisible to search


angela-helios:
Found in the #448 Elasticsearch leg (`sqlite-es`, `HFS_COMPOSITE_SYNC_MODE=synchronous`, ES 8.13.4).

A 20-file / 15,514-line `$bulk-submit` completes cleanly (status manifest: 20 outputs, 0 errors; SQLite primary holds all 30 patients and reads them back fine) — but **search sees none of it**:

```
GET /Patient/{bulk-id}                 → 200          (primary read)
GET /Patient?_id={bulk-id}            → total: 0     (search, via ES)
GET /Patient?_summary=count           → 10           (only the transaction-bundle cohort)
ES: hfs_default_patient resource_id:{bulk-id} → 0 hits
```

Transaction-Bundle writes index into ES correctly (the 10 tx patients are searchable), so the gap is specific to the bulk ingestion path: the SQLite engine skips its local search indexing when search is offloaded to the composite, but the composite's `process_ndjson_stream` delegates straight to the primary and nothing ever tells Elasticsearch. `synchronous` sync mode makes no difference.

Two ironies worth recording:
- the skipped local indexing makes bulk ingest **very fast** on this backend (~1,000 lines/s — the 15.5k cohort landed in ~15s) while producing search-invisible data;
- "bulk loads that immediately search" is the workload `HFS_COMPOSITE_SYNC_MODE=synchronous` exists for, per its own docs.

Workaround until fixed: `POST /$reindex` after a bulk load. Fix shape: the composite's bulk ingestion needs to feed the ES writer with the ingested resources (batch-friendly — the ES bulk API fits the per-batch commit structure from #815/#872).

mauripunzueta:
Same failure class as this one — a resource that is stored but not searchable — reached by a new route in #1050: `$reindex` silently drops resources that exceed Elasticsearch's 10,000 nested-object limit (458 `Provenance` in the Synthea corpus). I'm working on it in `fix/1050-search-index-integrity`.

PR #962, opened 2026-09-06
fix: Inferno nightly — batch the ES conformance seed under wait_for, ms-precision SQLite dates, ES _count body

smunini:
## Summary

The scheduled Inferno US Core run [34020931692](https://github.com/HeliosSoftware/hfs/actions/runs/34020931692) went red for two unrelated reasons. This PR fixes both (each needed a second round after the first branch run, [34056011998](https://github.com/HeliosSoftware/hfs/actions/runs/34056011998), exposed the next layer), plus a third bug found on the way.

### 1. Every `*-elasticsearch` leg: HFS never became ready, then transaction Bundles timed out

Since 2913df484 those legs run with `HFS_ELASTICSEARCH_WRITE_REFRESH=wait_for`. That policy blocks each Elasticsearch write until the next scheduled refresh (1s), and the composite synced resources to ES one synchronous event at a time. Two places paid for it:

- **Startup seed** — ~1.4k conformance resources at concurrency 1 ⇒ ~1.4k refresh waits; `/health` never answered within 60s.
- **Transaction Bundles** — one sync event per entry ⇒ any bundle over ~30 entries hit the 30s request timeout (HTTP 408 on the Inferno data load).

Fix: a batch write on the storage trait, and every multi-resource sync path routed through it.

- `ResourceStorage::create_many` — one result per input, in order. Default fans `create` out at `bulk_write_concurrency`, so no backend changes behaviour unless it overrides.
- Elasticsearch overrides it with one `_bulk` request per 500 operations (contained documents included), applying the write-refresh policy once per request.
- The composite forwards a batch to the primary, then to each secondary as a batch (`SyncManager::sync_creates`); rejected items fall back to the existing per-event retry path; asynchronous mode still queues one event per resource.
- The seeder writes in batches of 256; a transaction's entries are synced one batch per resource type; the bulk-submit manifest sync (#882) batches each page of entry results.

Local reproduction of CI's environment: HFS ready in **11s** (CI run: 8s), 1,374 SearchParameters + 5 CompartmentDefinitions seeded.

The one path that cannot batch is the REST batch handler (heterogeneous entries, per-entry auth and audit): the `s3-elasticsearch` legs load a 473-entry Bundle as a *batch* because S3 has no transactions, and at 16-way concurrency that is ~30 refresh waits. CI now pairs `wait_for` with `HFS_ELASTICSEARCH_REFRESH_INTERVAL=200ms` (Inferno and UI-matrix workflows), which is the per-write price of `wait_for`.

### 2. `sqlite` v7.0.0 / v8.0.0 legs: `_lastUpdated` search failures

The SQLite date handler compared millisecond-precision values at second precision (a documented #456 trade-off). Latent until the recent SQLite write-path speedups let a transaction Bundle land all its resources inside one second; Inferno's exact-timestamp search then returned neighbouring resources. Values with fractional seconds now compare through `strftime('%f')`.

The first branch run then failed by exactly 1ms: `last_updated` holds nanoseconds, which SQLite *rounds* to the millisecond, while `meta.lastUpdated` truncates. Both sides are now cut to three fractional digits in SQL before the comparison. (Writing the column at millisecond precision was tried and rejected — a `since` captured in the same millisecond would then skip that resource in `modified_since`.)

### 3. ES `_total=accurate` returned 500 (since February)

`build_count_query` left `track_total_hits`/`size` in the `_count` body, which the API rejects. The body is now just the query clause.

## Tests

- Seeder unit tests for the batching core.
- ES integration: a 40-document `create_many` under `wait_for` with a 5s refresh interval must finish in seconds and be fully visible on return; contained docs and server-assigned ids through the bulk path.
- Composite (SQLite→SQLite, synchronous): seeding complete on the secondary on return, re-seed writes nothing, per-item `AlreadyExists` in its own slot, a 6-entry transaction fully synced.
- SQLite: the millisecond condition and the SQL truncation evaluated in a real connection; an end-to-end test that creates several resources in one second and searches each one's own `meta.lastUpdated` with `eq` plus its neighbouring milliseconds.

`cargo test -p helios-persistence` (lib, sqlite, composite, bulk-submit, ES via testcontainers), clippy with CI flags, `cargo fmt --check`, and `cargo check -p helios-hfs -p helios-rest --all-features --tests` all pass.

https://claude.ai/code/session_01Nu2HKgueYRegxFYEvV7XKS

### DIFF
--- a/.github/workflows/inferno-us-core.yml
+++ b/.github/workflows/inferno-us-core.yml
@@ -431,12 +431,19 @@ jobs:
           echo "HFS_LOG=$HFS_LOG" >> $GITHUB_ENV
           echo "HFS_BASE_URL=$HFS_BASE_URL" >> $GITHUB_ENV
 
+          # The Elasticsearch legs pair synchronous sync with wait_for so a
+          # write is searchable when the API responds. wait_for blocks each
+          # write until the next refresh, so the refresh interval is the price
+          # of every synchronous write: at the 1s default a 473-entry batch
+          # Bundle (entries synced one at a time, 16-wide) ran into the 30s
+          # request timeout; 200ms keeps it to a few seconds.
           if [ "${{ matrix.backend }}" = "sqlite-elasticsearch" ]; then
             HFS_BASE_URL="$HFS_BASE_URL" \
             HFS_STORAGE_BACKEND=sqlite-elasticsearch \
             HFS_ELASTICSEARCH_NODES=http://$DOCKER_HOST_IP:$ES_PORT \
             HFS_COMPOSITE_SYNC_MODE=synchronous \
             HFS_ELASTICSEARCH_WRITE_REFRESH=wait_for \
+            HFS_ELASTICSEARCH_REFRESH_INTERVAL=200ms \
             ./target/debug/hfs --database-url :memory: --log-level info --port $HFS_PORT --host 0.0.0.0 > "$HFS_LOG" 2>&1 &
           elif [ "${{ matrix.backend }}" = "postgres" ]; then
             HFS_BASE_URL="$HFS_BASE_URL" \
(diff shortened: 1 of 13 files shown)


smunini:
Inferno US Core on this branch (`e69c7f554`): **all 32 legs green** — https://github.com/HeliosSoftware/hfs/actions/runs/34062557249

The scheduled run 34020931692 had 13 red legs (8 `*-elasticsearch`, sqlite v7.0.0/v8.0.0, plus the first branch run's 408s on transaction Bundles). All pass now, including every s3-elasticsearch leg with the 473-entry batch Bundle.

PR #986, opened 2026-09-08
perf(bulk-submit): replace SQL receipt OFFSET pagination with keyset cursors

### LINKED ISSUE #973: perf(postgres): replace deep OFFSET bulk-submit receipt pagination with a stable cursor
## Implementation scope after profiling

Profiling was completed on our end on September 7, 2026, using the real asynchronous bulk-submit worker in an isolated development environment. The diagnostic instrumentation has been preserved externally and removed from the working tree. No performance fix was applied. This issue is now scoped to stable receipt pagination in SQL-backed bulk-submit, including Composite's second traversal.

The existing full-corpus incident remains relevant: the import had processed 18,955,865 resources but was still finalizing at 25h 30m. The original report is preserved below as historical evidence. The newer synthetic measurements isolate the mechanism without operating on that QA job.

This issue is the first part of a split:

- This issue: keyset receipt traversal, backend compatibility, exact-result tests, and controlled before/after.
- #979: local output buffering.
- #980: idempotent publication and lease ownership across retries.
- #982: bounded receipt construction, after pagination and the publication contract.
- #981: focused ingestion diagnosis under the broader #945 roadmap.

Retain the current whole-manifest receipt accumulation and artifact writer in this issue so that the pagination effect is measurable. Preserve FHIR bulk-submit kickoff, polling, terminal manifests, receipt/error content, counts, tenant isolation, and existing ingestion error isolation. Keep non-SQL supported backends working. Do not silently narrow the feature to PostgreSQL.

## What was measured

Base source: `acb9284a75f076ca866be3989d93dbc2041716ed`, plus an opt-in diagnostic patch. These source links are pinned to the unmodified base, not to whatever main contains when this is implemented.

| Resources | Mode | Worker wall | Ingestion | Finalization | Receipt read | Artifact write/register |
|---:|---|---:|---:|---:|---:|---:|
| 12,000 | Fresh | 10.086 s | 9.761 s | 0.314 s | 0.050 s | 0.258 s |
| 12,000 | Reimport | 12.966 s | 12.654 s | 0.304 s | 0.045 s | 0.253 s |
| 48,000 | Fresh | 46.720 s | 44.883 s | 1.827 s | 0.672 s | 1.134 s |
| 48,000 | Reimport | 53.107 s | 51.400 s | 1.699 s | 0.599 s | 1.075 s |
| 192,000 | Fresh | 209.814 s | 194.596 s | 15.211 s | 10.012 s | 5.098 s |

All five runs reached terminal HTTP 200, completed manifests, zero failed resources, and exact stored-resource/receipt reconciliation. Tied line numbers did not produce missing receipts in these five real-worker runs; ordering correctness still requires an exact regression test because `line_number` alone is not unique across files.

Each fourfold fresh-corpus increase grew receipt-read time by roughly 13.4x and 14.9x. At 192k the read window used only 0.201 HFS CPU-seconds over 10.012 wall seconds. PostgreSQL ran in a separate process, so HFS CPU profiling cannot divide the remaining time among PostgreSQL execution, transport, or scheduling. The SQL plans provide the additional evidence:

| Post-capture query, same 192k manifest | Execution | Plan behavior |
|---|---:|---|
| Current OFFSET 0 | 37.983 ms | Scan matching 192k rows, parallel top-N sort, return 1,000 |
| Current OFFSET 191,000 | 82.781 ms | Scan/sort 192k rows, external merge, 1,272 temp blocks read and 1,276 written |
| Hypothetical cursor at the tail | 0.424 ms | `bulk_entry_results_pkey` index scan, 1,000 rows, no sort/temp I/O |

The cursor row is a read-only SQL experiment after capture, not a worker implementation or an end-to-end speedup. One observation per original case, growing database state, cache state, and diagnostic overhead mean that a controlled baseline must be rerun before claiming the after improvement.

## Code map

All links use the measured base revision:

- [Result type and provider contract](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/core/bulk_submit.rs#L374). `BulkEntryResult` has no `file_url`; the current offset method is in the same file at `get_entry_results`.
- [Worker readback and accumulation](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/core/bulk_submit_worker.rs#L918). `write_result_artifacts` reads pages of 1,000 with a growing offset, then serializes/writes artifacts.
- [PostgreSQL query](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/postgres/bulk_submit.rs#L802).
- [SQLite query](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/sqlite/bulk_submit.rs#L818).
- [Composite second traversal](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/composite/bulk_submit.rs#L81). `sync_ingested` reads success-filtered pages before completion and has its own OFFSET loop. The same module delegates provider reads at `get_entry_results` and invokes synchronization from `finish_manifest`.
- [PostgreSQL schema](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/postgres/schema.rs#L576) and [SQLite schema](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/sqlite/schema.rs#L762). Their receipt primary key is `(tenant_id, submitter, submission_id, manifest_id, file_url, line_number)`.
- [MongoDB implementation](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/mongodb/bulk_submit.rs) and [S3 implementation](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/s3/bulk_submit.rs). Both participate in the shared provider contract.

## Proposed internal contract and algorithm

Use an additive internal API rather than changing the meaning of ingest results or assuming every provider can recover a file URL. This is a design sketch; adapt naming to existing conventions:

```rust
struct EntryResultCursor {
    file_url: String,
    line_number: u64,
}

struct StoredEntryResult {
    cursor: EntryResultCursor,
    result: BulkEntryResult,
}

// Inputs also include tenant, SubmissionId, manifest, outcome filter and limit.
// Return stored identities with results, or an equivalent page + next-cursor type.
async fn get_entry_results_page(..., after: Option<&EntryResultCursor>, limit: u32)
    -> StorageResult<Vec<StoredEntryResult>>;
```

For SQL providers, select `file_url` along with the existing six columns. For a noninitial page, use a bound continuation predicate with the same order:

```sql
SELECT file_url, line_number, resource_type, resource_id,
       created, outcome, operation_outcome
FROM bulk_entry_results
WHERE tenant_id = $1 AND submitter = $2
  AND submission_id = $3 AND manifest_id = $4
  AND (file_url, line_number) > ($5, $6)
ORDER BY file_url, line_number
LIMIT $7;
```

The first page omits the cursor predicate. Use supported equivalent SQL for each backend and inspect the actual plan; an equivalent-looking OR predicate must not accidentally lose the intended access path. Bind all cursor/filter values. Respect the stored line-number representation and reject invalid limits/conversions rather than truncating.

Within a fixed manifest and outcome filter:

1. Fetch at most the configured page limit.
2. Consume the page, preserving existing receipt semantics.
3. Continue from the last returned stored identity, strictly greater than that pair.
4. Finish on an empty/short page according to the chosen page contract, including an exact multiple of the limit. Never advance from a resource ID or line number alone.

Keep query scope unchanged across pages. Success filtering must not skip a valid later page because a page was filtered only after limiting. Test and plan the actual success-filtered query used by Composite. Its outcome index has a different order from the cursor primary key; do not add a wide index speculatively. Measure first and explain any migration separately.

Expose capability/dispatch explicitly. PostgreSQL and SQLite must use native keyset traversal in both the worker and Composite. Preserve the existing offset method for compatibility where needed. Other backends may retain an explicitly selected legacy path in this scoped change. S3 stores file identity as a hash and cannot reconstruct every original URL from existing receipt objects; do not fabricate a cursor or break old objects. If the new API affects trait implementers, update and verify their capability/delegation behavior. The keyset guarantee in this issue applies to the verified SQL primaries.

## Regression tests and acceptance

Add tests at the actual paging and worker boundaries before fixing the code. A count-only test is insufficient because omitted and duplicated identities can cancel out.

- [ ] SQL page-contract tests cover empty input, one page, exact multiples, more than two pages, many files with repeated line numbers, mixed outcomes, and boundaries that split tied line numbers.
- [ ] Assert the full ordered stored identity sequence `(file_url, line_number)` for the new contract and the exact receipt reference multiset. Include repeated resource IDs on distinct input lines so receipt identity is not confused with resource identity.
- [ ] Test tenant, submitter, submission and manifest isolation, including identical file/line identities in another scope.
- [ ] Repeat coverage with `outcome=success`; other outcomes cannot cause false EOF or gaps.
- [ ] Extend [the worker integration tests](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/core/bulk_submit_worker.rs#L1336) with multiple files and more than 1,000 results. Download/parse artifacts and check references, error content, severity counts and terminal status. Existing success/skipped/error semantics remain unchanged.
- [ ] Extend [Composite tests](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/composite/bulk_submit.rs#L804) with several pages, repeated resources and mixed outcomes. Each successful resource is synchronized according to the existing deduplication contract. A fake primary that fails on legacy OFFSET calls should prove both SQL-backed traversal paths migrated.
- [ ] Verify PostgreSQL and SQLite behavior and compile/test affected MongoDB/S3 capability paths. Preserve bulk operations for those backends.
- [ ] Run controlled before/after at 12k, 48k and 192k and inspect early/late plans. Report page latency, scanned rows/buffers, temporary I/O, whole finalization and whole worker time. Do not place fragile millisecond thresholds in CI.
- [ ] Preserve current artifact construction/publication and ingestion settings. No bounded-memory, S3 optimization, ingestion batching or global speedup claim belongs to this change.

Starting test commands, with one build job and Docker available for container-backed tests:

```bash
CARGO_BUILD_JOBS=1 cargo test --locked -p helios-persistence \
  --no-default-features --features R4,sqlite,postgres --lib bulk_submit -- --nocapture
CARGO_BUILD_JOBS=1 cargo test --locked -p helios-persistence \
  --no-default-features --features R4,sqlite --test sqlite_bulk_submit_concurrency -- --nocapture
```

Add/run the new PostgreSQL regression through the existing `postgres_tests` fixture or the appropriate new integration target. Confirm that database tests actually executed; an environment-gated skip is not a pass. Build affected provider combinations separately rather than using an expensive blanket `--all-features` build. Recheck repository test guidance and current feature gates when implementing.

## Reproduction and before/after evidence

The original real-worker corpus is deterministic: four homogeneous NDJSON files, two Patient and two Observation files, equal file sizes by row count, distinct resource IDs, and line numbers restarting at 1 in every file. Total sizes were 12,000, 48,000, and 192,000 resources. Patient IDs/names/identifiers and referenced Observation values are synthetic. The exact generator is included below so implementing agents are not dependent on a local archive path.

Protocol used:

1. Serve the four files and a Bulk Data manifest from an isolated loopback HTTP provider. The manifest has `transactionTime`, `request`, `requiresAccessToken: false`, an `output` list of `{type, url, count}`, and empty `error` and `deleted` arrays.
2. POST `/$bulk-submit` with the Parameters shown below and a unique `X-Tenant-ID`. Expect HTTP 200 from kickoff.
3. POST `/$bulk-submit-status` with the same submitter/submissionId. Expect HTTP 202 and follow `Content-Location`.
4. Poll to terminal HTTP 200, follow all status-manifest pages, download all artifacts, and reconcile exact references and counts against generated IDs. Also verify persisted resources and completed manifest counters.
5. For reimport use a new submission ID with the same corpus and tenant after its fresh run. Keep fresh/reimport results separate.

Kickoff shape:

```json
{
  "resourceType": "Parameters",
  "parameter": [
    {"name":"submitter","valueIdentifier":{"system":"urn:helios:profiling:973","value":"synthetic"}},
    {"name":"submissionId","valueString":"UNIQUE_SUBMISSION_ID"},
    {"name":"manifestUrl","valueUrl":"http://127.0.0.1:PROVIDER_PORT/manifest.json"},
    {"name":"fhirBaseUrl","valueUrl":"http://127.0.0.1:PROVIDER_PORT/fhir"},
    {"name":"submissionStatus","valueCoding":{"system":"http://hl7.org/fhir/event-status","code":"completed"}}
  ]
}
```

The status Parameters contain only the first two parameters. Use `Content-Type: application/fhir+json` and the same tenant header on every HFS request. Authentication was disabled for this local synthetic capture; that does not authorize changing production authentication.

Measured configuration: optimized R4/PostgreSQL HFS, native Apple M3 Max 36 GiB, PostgreSQL 16 in Docker capped at 2 CPUs/2 GiB, one submit worker, four input-file tasks, inline indexing, local-fs output, effective worker batch 100. The REST default of 1,000 was not wired into the async worker in the measured source; do not alter it in this issue. The SQL provider port in the diagnostic environment was host 5433, HFS 8081. Keep clear of QA's ports 5432/8080 and any active user environment.

The diagnostic build used `RUSTFLAGS='--cfg perf_phases -C force-frame-pointers=yes'`, `CARGO_PROFILE_RELEASE_DEBUG=line-tables-only`, and:

```bash
CARGO_BUILD_JOBS=1 cargo build --locked --release -p helios-hfs \
  --no-default-features --features R4,postgres
```

`HFS_PERF_PHASES=1` enabled the temporary counters, with Samply 0.13.1 at 500 Hz, PostgreSQL activity sampled about every second, and RSS about every 0.2 seconds. The clean base contains the existing opt-in counter infrastructure but not the temporary PostgreSQL/worker timing call sites. Reintroduce equivalent diagnostic coverage outside the shipping patch when needed; setting the environment variable alone does not reproduce this coverage.

For controlled before/after, restore equivalent database starting states, use identical corpus hashes, schema/indexes, startup/warm-up policy, runtime options and instrumentation. Perform at least three timing trials per mandatory case: 12k and 48k fresh/reimport, plus 192k fresh. Alternate order where practical. Report all runs and median/range. Use sampler-free timing on both sides as the primary elapsed comparison and matching profiled runs to explain the change. Stage scopes and process CPU totals overlap; do not sum nested/concurrent values as wall time.

Keep logs, profiles, plots, benchmarks' generated data, binaries and detailed reports outside the source repository. Attach or link the dated before/after evidence in the PR. The code and tests remain the source of truth for behavior. Essential results are embedded here; raw archives are preserved locally and have not been uploaded as issue attachments. Their integrity identifiers are:

- `hfs-973-before-evidence.tar.gz`: SHA-256 `857777ca5d14c87097d5924afa762db19c1fcc65b7685909241a519001b6b07b`.
- `hfs-973-before-symbols.tar.gz`: SHA-256 `252d67071d5483c657708d6ad58f097b8485902fb30dfa6146b1bba6a6a991bb`.
- Diagnostic patch: SHA-256 `08eef7c25b9838d5c5950d62fb94970b60d3ee9f7c2af094b94ba45706767d0c`.

## Scope boundaries and related work

Progress/status presentation is related to #953, #954 and #955; preserve existing reporting but do not fold a UI redesign into this pagination fix. Streaming and stronger publication semantics are explicitly delegated to the linked issues. Existing ingestion improvements in #880 and the SQLite work in #949/#952/#970 do not replace the receipt-pagination fix. Do not carry forward old per-resource SQL-count claims without inspecting current code.

<details>
<summary>Exact synthetic corpus generator used for the diagnostic capture</summary>

Save outside the repository as `generate_corpus.py` and run with `--output /external/evidence/corpus-12000 --total 12000`, then repeat for 48000 and 192000. It refuses to overwrite existing data. This is the already-used input generator, not an untested performance fix.

```python
#!/usr/bin/env python3
"""Generate a deterministic, multi-file FHIR R4 NDJSON profiling corpus."""

from __future__ import annotations

import argparse
import hashlib
import json
from pathlib import Path


DEFAULT_TOTAL = 12_000
LARGE_TOTAL = 48_000


def compact_json(value: object) -> str:
    return json.dumps(value, separators=(",", ":"), sort_keys=True)


def write_ndjson(path: Path, resources: list[dict[str, object]]) -> dict[str, object]:
    digest = hashlib.sha256()
    byte_count = 0
    with path.open("wb") as stream:
        for resource in resources:
            line = (compact_json(resource) + "\n").encode("utf-8")
            stream.write(line)
            digest.update(line)
            byte_count += len(line)
    return {
        "name": path.name,
        "resource_type": resources[0]["resourceType"],
        "count": len(resources),
        "bytes": byte_count,
        "sha256": digest.hexdigest(),
    }


def patient(dataset_tag: str, index: int) -> dict[str, object]:
    return {
        "resourceType": "Patient",
        "id": f"p973-{dataset_tag}-p-{index:06d}",
        "active": True,
        "identifier": [
            {
                "system": "urn:helios:profiling:973:patient",
                "value": f"{dataset_tag}-{index:06d}",
            }
        ],
        "name": [{"family": f"Profile{index % 997:03d}", "given": [f"Synthetic{index:06d}"]}],
        "gender": "female" if index % 2 == 0 else "male",
        "birthDate": f"{1940 + index % 65:04d}-{1 + index % 12:02d}-{1 + index % 28:02d}",
    }


def observation(dataset_tag: str, index: int, patient_count: int) -> dict[str, object]:
    patient_index = index % patient_count
    return {
        "resourceType": "Observation",
        "id": f"p973-{dataset_tag}-o-{index:06d}",
        "status": "final",
        "category": [
            {
                "coding": [
                    {
                        "system": "http://terminology.hl7.org/CodeSystem/observation-category",
                        "code": "vital-signs",
                    }
                ]
            }
        ],
        "code": {
            "coding": [
                {
                    "system": "http://loinc.org",
                    "code": "29463-7",
                    "display": "Body weight",
                }
            ]
        },
        "subject": {"reference": f"Patient/p973-{dataset_tag}-p-{patient_index:06d}"},
        "effectiveDateTime": f"2025-{1 + index % 12:02d}-{1 + index % 28:02d}T12:00:00Z",
        "valueQuantity": {
            "value": 50 + (index % 700) / 10,
            "unit": "kg",
            "system": "http://unitsofmeasure.org",
            "code": "kg",
        },
    }


def generate(output: Path, total: int) -> None:
    if total < 4 or total % 4:
        raise SystemExit("--total must be at least 4 and divisible by 4")
    output.mkdir(parents=True, exist_ok=True)
    unexpected = [path for path in output.iterdir() if path.name != ".gitkeep"]
    if unexpected:
        raise SystemExit(f"refusing to overwrite non-empty directory: {output}")

    dataset_tag = str(total)
    per_file = total // 4
    patient_count = total // 2
    definitions = [
        ("patients-01.ndjson", [patient(dataset_tag, i) for i in range(0, per_file)]),
        (
            "patients-02.ndjson",
            [patient(dataset_tag, i) for i in range(per_file, patient_count)],
        ),
        (
            "observations-01.ndjson",
            [observation(dataset_tag, i, patient_count) for i in range(0, per_file)],
        ),
        (
            "observations-02.ndjson",
            [observation(dataset_tag, i, patient_count) for i in range(per_file, patient_count)],
        ),
    ]

    files = [write_ndjson(output / name, rows) for name, rows in definitions]
    expected_ids = {
        "Patient": [f"p973-{dataset_tag}-p-{i:06d}" for i in range(patient_count)],
        "Observation": [f"p973-{dataset_tag}-o-{i:06d}" for i in range(patient_count)],
    }
    expected_path = output / "expected-ids.json"
    expected_path.write_text(compact_json(expected_ids) + "\n", encoding="utf-8")
    expected_digest = hashlib.sha256(expected_path.read_bytes()).hexdigest()

    metadata = {
        "format_version": 1,
        "total_resources": total,
        "dataset_tag": dataset_tag,
        "files": files,
        "expected_ids": expected_path.name,
        "expected_ids_sha256": expected_digest,
        "notes": [
            "Each NDJSON file starts at line 1, so line numbers repeat across files.",
            "The runner builds the absolute manifest URLs after binding a free provider port.",
        ],
    }
    (output / "corpus.json").write_text(
        json.dumps(metadata, indent=2, sort_keys=True) + "\n", encoding="utf-8"
    )
    print(f"generated {total} resources in {output}")


def main() -> None:
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("--output", type=Path, required=True)
    size = parser.add_mutually_exclusive_group()
    size.add_argument("--total", type=int, default=DEFAULT_TOTAL)
    size.add_argument(
        "--large",
        action="store_true",
        help=f"generate the {LARGE_TOTAL}-resource variant",
    )
    args = parser.parse_args()
    generate(args.output.resolve(), LARGE_TOTAL if args.large else args.total)


if __name__ == "__main__":
    main()
```

</details>

<details>
<summary>Corpus file SHA-256 values</summary>

| Resources | File | SHA-256 |
|---:|---|---|
| 12000 | `patients-01.ndjson` | `e52498f2e2abefb4009ad0bfd033b4e32db29691b29057c044007af284f652b6` |
| 12000 | `patients-02.ndjson` | `fa2537774d2e07ec88c0cf2303b99b0ffddbe1770da21dbdc8665fe833c449b9` |
| 12000 | `observations-01.ndjson` | `db59f85e1355dcee8d11cf2c8d3032018474269d6855aefe49f0e6e1f5191f49` |
| 12000 | `observations-02.ndjson` | `257a8b6a368bd93ba0bd8594efcd46bb928183838dfa7c9db401047234a80329` |
| 48000 | `patients-01.ndjson` | `bcb4df997567f8eec57cc3e83a00582488552a1217a0590bf65bc2cedcca62ff` |
| 48000 | `patients-02.ndjson` | `afd73aa4226ebd731547dd2d03b5787daf9ed4a98339f357ecc4a064a3106df3` |
| 48000 | `observations-01.ndjson` | `edbc5d88e99b09fcbc5be4f11a7b622beb32ac73ff64dae79a14376934461f73` |
| 48000 | `observations-02.ndjson` | `fe1e3a0d567db990e02dbd14d9af3864d5cb9fab5dc56f859cd64681dd125dae` |
| 192000 | `patients-01.ndjson` | `a1b97b3fded840fcc8c71de86de0e2f0e16374069fd77fffa81b631baca8669f` |
| 192000 | `patients-02.ndjson` | `543702b5c96f354f537748a43f63908834c6064c819d44b7a4717aeaedfbf4e2` |
| 192000 | `observations-01.ndjson` | `06dc3adf0d06ef291e968a0d9f3d91bfd197c257b9b93445d1fc4f0a4f9a6e5a` |
| 192000 | `observations-02.ndjson` | `2f36f99e328655fb153665de31725a0d63f5b5b1b3596dc8f1bb3cba36df03ce` |

</details>

<details>
<summary>Original full-corpus incident report and historical proposal</summary>

The observations below are preserved from the original issue. Its combined proposal is superseded by the split implementation scope above; in particular, streaming/publication and UI progress work are separate. Historical descriptions of the live job are not a claim about its state today.

## Problem

During the PostgreSQL manual QA pass (#938), a full Synthea bulk import has still not finished after **25h 30m**, although HFS reports **18,955,865 / 18,955,865 entries processed, 37,401,034,861 / 37,401,034,861 bytes, and 0 failed entries**.

The current bottleneck is **status-artifact finalization**, not an outstanding corpus download: the worker repeatedly reads `bulk_entry_results` using `ORDER BY line_number LIMIT 1000 OFFSET ...` at offsets above 17.8 million. The live PostgreSQL plan uses a parallel sequential scan and sort. No status artifact rows have been published for this submission yet, and the manifest remains `processing`.

This is a concrete finalization/scalability issue related to the broader ingestion roadmap #945. SQLite's recent improvements provide useful precedent and profiling infrastructure, but do not directly fix this PostgreSQL query.

The same readback/accumulation implementation is still present on current main, **`acb9284a75f076ca866be3989d93dbc2041716ed`**, after #970: [shared worker](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/core/bulk_submit_worker.rs#L918), [PostgreSQL query](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/postgres/bulk_submit.rs#L802). This is source inspection, not a second full-corpus reproduction on that commit.

## Environment and reproduction

- Apple M3 Max, 36 GiB RAM, macOS 26.6.2; Docker VM memory limit reported as 15.6 GiB.
- Native release `hfs` v0.2.1, source checkout **`a4068bfd6405c7aaa440b1d5143767bfdae049d9`**, built September 5. PostgreSQL 16 in local Docker, inline search indexing, R4, UI enabled.
- `HFS_BULK_SUBMIT_FILE_CONCURRENCY=4`; PostgreSQL container has no recorded restart or OOM kill.
- Full [manual test corpus and T3 workflow](https://github.com/HeliosSoftware/hfs/blob/a4068bfd6405c7aaa440b1d5143767bfdae049d9/MANUAL_TESTING_MATRIX.md#7-t3--import-the-synthea-corpus-from-the-import-page): 24 NDJSON files, 11,704 patients, 18,955,865 resources, 37.4 GB.
- Corpus served by local Nginx with a read-only bind mount. Before kickoff, four concurrent HTTP readers downloaded all 24 complete files to a discard sink successfully.
- Started from the Import UI as `synthea-postgres-nginx` at **2026-09-06T13:27:44.481Z**. Submission ID: `a5d4a859-e4fa-4330-b06f-5e25430ad145`.
- This was a **re-import into an already populated database** after partial attempts. Existing resources were upserted by ID; this is not a clean-database throughput benchmark. Earlier attempts had fixture HTTP failures and must not be included in the current job's elapsed time.

Reproduce by running the full T3 import on PostgreSQL, allowing ingestion to drain, and measuring time from the end of ingestion through artifact publication and terminal status. Also compare clean-database and re-import runs separately.

## Live evidence, September 7 UTC

| Observation | Result |
|---|---|
| Current job age at 14:57:40 | 25h 29m 56s |
| Manifest | `processing`, active worker lease, fencing token 1 |
| Processed entries / failed | 18,955,865 / 0 |
| Processed / total bytes | 37,401,034,861 / 37,401,034,861 |
| `bulk_submit_files` rows for this submission | 0 |
| Last corpus NDJSON GET completed | Observation, **00:03:54Z**, HTTP 200, full 7,523,591,229 bytes |
| PostgreSQL / HFS activity | PostgreSQL ~168% CPU in one snapshot; HFS near idle while awaiting query results |
| PostgreSQL database size | 257 GB, including earlier attempts |
| Database-wide temporary I/O counters | 75,158 temp files / 27,125,045,698,560 bytes; cumulative, not attributed solely to this job |

The last GET finished about 14h 54m before the diagnostic snapshot. That bounds the end of HTTP transfer, **not an instrumented ingestion-complete timestamp**.

Active result-query offsets advanced from **17,793,000** to **17,808,000** to **17,817,000** during this investigation. The worker is progressing slowly, rather than deadlocked.

Observed query:

```sql
SELECT line_number, resource_type, resource_id, created, outcome, operation_outcome
FROM bulk_entry_results
WHERE tenant_id = $1 AND submitter = $2
  AND submission_id = $3 AND manifest_id = $4
ORDER BY line_number LIMIT 1000 OFFSET 17793000;
```

A read-only `EXPLAIN` using this submission's values, **without ANALYZE**, returned:

```text
Limit
  -> Gather Merge
       Workers Planned: 2
       -> Sort
            Sort Key: line_number
            -> Parallel Seq Scan on bulk_entry_results
                 Filter: tenant_id / submitter / submission_id / manifest_id
```

The live primary key is `(tenant_id, submitter, submission_id, manifest_id, file_url, line_number)`; the other index ends in `outcome`. Neither provides the requested global `line_number` ordering after the manifest prefix.

## Code path and likely mechanism

- [Shared worker: write_result_artifacts](https://github.com/HeliosSoftware/hfs/blob/a4068bfd6405c7aaa440b1d5143767bfdae049d9/crates/persistence/src/core/bulk_submit_worker.rs#L918) reads every result in 1,000-row offset pages, accumulates them all in a `Vec`, then builds per-type serialized receipt vectors and only afterwards writes artifacts.
- [PostgreSQL: get_entry_results](https://github.com/HeliosSoftware/hfs/blob/a4068bfd6405c7aaa440b1d5143767bfdae049d9/crates/persistence/src/backends/postgres/bulk_submit.rs#L802) generates the observed `ORDER BY line_number LIMIT/OFFSET` query.
- [Worker lifecycle](https://github.com/HeliosSoftware/hfs/blob/a4068bfd6405c7aaa440b1d5143767bfdae049d9/crates/persistence/src/core/bulk_submit_worker.rs#L869) finishes the manifest only after artifact generation returns.

Repeated growing-offset reads cause superlinear work; the observed plan additionally scans/sorts the large result relation for each page. Whole-manifest accumulation adds memory growth proportional to result count, followed by a second representation of the receipts.

**Correctness risk to test:** line numbers restart per file. Ordering solely by `line_number` is not unique, so independent LIMIT/OFFSET queries may overlap or omit tied rows. This investigation has not established actual duplicate/missing receipts for the current run, which has not published them.

## Related performance work

- [#880](https://github.com/HeliosSoftware/hfs/pull/880), PostgreSQL batch transactions and pipelined writes, is **already included** in the running checkout. It improves ingestion, not this readback loop.
- [#945](https://github.com/HeliosSoftware/hfs/issues/945) tracks PostgreSQL statement round trips and index write volume. Those remain relevant to ingestion throughput; the current live query isolates a separate finalization bottleneck.
- [#949](https://github.com/HeliosSoftware/hfs/pull/949) removed SQLite's per-create FTS scan; [#970](https://github.com/HeliosSoftware/hfs/pull/970) later replaced scanning FTS deletes with rowid lookups for updates/re-imports. The transferable lesson is to eliminate repeated corpus-scale scans. SQLite FTS5 changes and their measured speedups are not PostgreSQL fixes or directly comparable benchmarks.
- [#952](https://github.com/HeliosSoftware/hfs/pull/952), merged September 6, adds phase-profiling infrastructure and SQLite write-path improvements. It is **not in this running build**. Its timers/example target SQLite; PostgreSQL finalization requires additional measurement coverage.

Profiling setup: the [#952 counter module](https://github.com/HeliosSoftware/hfs/blob/7a0878561eef8e56d607ae41186ed1fa947f6eae/crates/persistence/src/perf.rs) requires `RUSTFLAGS='--cfg perf_phases'` at build time and `HFS_PERF_PHASES=1` to enable collection at runtime. Its SQLite benchmark enables counters itself and excludes worker finalization. No automatic HFS reporting consumer or PostgreSQL phase call sites were found in current main, so setting an environment variable alone will not profile this running job. Extend both coverage and reporting in a separate diagnostic build. Nested counters overlap and are not disjoint wall-time buckets.

## Proposed investigation and acceptance criteria

- [ ] Benchmark ingestion and finalization **separately** at multiple corpus sizes, with clean-create and re-import cases. Report resources/s, finalization time, peak memory, temporary I/O and receipt counts.
- [ ] Extend the phase instrumentation from #952 to PostgreSQL result fetching, receipt serialization and artifact writing. A Rust CPU profile alone may show waiting while the work is inside PostgreSQL; correlate it with bounded query samples and plans.
- [ ] Replace growing OFFSET readback with a stable unique cursor, for example `(file_url, line_number)` within tenant/submission/manifest, with a matching access path. Do not use line number alone as the cursor.
- [ ] Stream receipt/error artifacts with bounded buffers instead of retaining all results plus all serialized lines in memory; preserve lease fencing and error semantics.
- [ ] Add multi-file regression coverage with repeated line numbers, page boundaries, success/error/skipped outcomes, isolation and exact receipt totals. Verify restart/lease-loss behavior.
- [ ] Verify on a realistically sized PostgreSQL corpus that finalization no longer performs repeated whole-result scans/sorts and scales approximately linearly.
- [ ] Report the finalization phase in progress/status so 100% input processing does not mask hours of remaining work; publish complete artifacts before declaring success.

The live run was inspected using bounded read-only database queries. It was not restarted, cancelled, re-imported, or schema-modified during this investigation.

</details>

Gordex2014:
Bulk-submit receipt finalization repeatedly scanned earlier rows through OFFSET pagination, while ordering by line number alone did not uniquely order receipts across input files. PostgreSQL and SQLite now traverse the stored `(file_url, line_number)` identity with bound keyset predicates and matching ordering.

The shared provider API returns pages with opaque continuations. SQL providers validate cursor bounds and propagate malformed stored rows. Worker and Composite consumers follow continuations through empty pages, preserving receipt references, artifact counts, resource deduplication and best-effort synchronization. MongoDB and S3 retain their existing traversal and persisted formats.

**Rust API migration:** `BulkSubmitProvider::get_entry_results` is replaced by `get_entry_results_page`. Implementations and callers must return or consume `EntryResultPage` and `EntryResultContinuation`. This is source-incompatible. The bulk-submit HTTP protocol and serialized receipt format remain unchanged.

Validation covers both SQL engines, real worker artifacts, Composite finish/fail behavior, late-page resources, secondary failures, scope isolation, filters, continuation boundaries and malformed data. The final focused suites passed: 54 SQL library tests, 5 PostgreSQL integration tests, 6 SQLite concurrency tests, 75 S3-enabled library tests and 12 live MongoDB integration tests. All four separate minimal provider builds and formatting checks passed. The suites overlap. Three independent code reviews and the performance-report review have no outstanding findings.

New before/after measurements used identical corpora and configuration, retained starting-state snapshots and exact result reconciliation. The matrix includes 36 PostgreSQL HTTP runs, 11 separate diagnostic runs and 54 SQLite traversals.

| Measurement at 192,000 resources | Before | After |
|---|---:|---:|
| PostgreSQL diagnostic receipt read | 12.727 s | 0.529 s |
| PostgreSQL diagnostic finalization | 16.725 s | 4.902 s |
| SQLite unfiltered isolated SQL read, median of three | 11.270 s | 0.101 s |

These are receipt-stage improvements, not whole-import speedup factors. SQLite uses isolated production SQL and decoding; full consumer correctness is tested separately. PostgreSQL HTTP completion at 192,000 resources had medians of 176.710 and 171.033 seconds, with overlapping ranges.

**Performance limits:** the original 48,000-resource reimport HTTP median increased 8.67%. Three additional matched-restore pairs changed by -3.47%, -0.88% and +7.76%; their after median remained 6.18% higher, with overlapping ranges. Those controls neither establish a global improvement nor exclude a regression. PostgreSQL 12k diagnostic reads and SQLite 12k sparse traversal also increased slightly in absolute time. Three-trial results are descriptive, and logical restores do not preserve physical layout or planner statistics.

Whole-manifest accumulation remains, so this change does not promise bounded worker memory. No ingestion settings, schema, indexes or dependencies change.

Closes #973

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1344,7 +1344,7 @@ jobs:
                   ? `  ${t.group}/${t.name}: FAILED - ${t.msg}`
                   : `  ${t.group}/${t.name}: DID NOT EXECUTE (no recorded result)`).join('\n');
                 const txNote = remaining.some(t => t.mode === 'tx')
-                  ? `\n\nThe HTS pre-flight probe was healthy (3/3), so terminology infrastructure is NOT the cause of the tx test(s) above -- this is the evaluator.`
+                  ? `\n\nThe HTS pre-flight probe was healthy (3/3), but this does not rule out a later terminology outage. Check the errors above for HTS/proxy failures as well as evaluator regressions.`
                   : '';
                 core.error(`[COVERAGE-LOST] ${remaining.length} test(s) declared in tests-fhir-r5.xml did not pass and are not listed in known-test-failures.json:\n${detail}${txNote}\n\nA test marked DID NOT EXECUTE was downgraded to inconclusive by the runner, or never ran at all; neither affects the runner's exit code, which is the silent skip this gate exists to catch. Fix the defect, or add an entry to known-test-failures.json with a reason and a linked issue. Do not delete the test to get green.`);
                 codes.push('COVERAGE-LOST');
--- a/crates/fhirpath/src/terminology_client.rs
+++ b/crates/fhirpath/src/terminology_client.rs
@@ -4,7 +4,7 @@
 //! It implements the standard FHIR terminology operations including expand, lookup,
 //! validate-code, subsumes, and translate.
 
-use reqwest::Client;
+use reqwest::{Client, RequestBuilder, Response, StatusCode};
 use serde_json::{Value, json};
 use std::collections::HashMap;
 use std::time::Duration;
@@ -20,6 +20,13 @@ const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
 
 /// Environment variable overriding [`DEFAULT_REQUEST_TIMEOUT`], in whole seconds.
 const REQUEST_TIMEOUT_ENV: &str = "FHIRPATH_TERMINOLOGY_TIMEOUT";
+
+/// Three retries after the initial request, with 1s, 2s and 4s backoff.
+const GATEWAY_RETRY_DELAYS: [Duration; 3] = [
+    Duration::from_secs(1),
+    Duration::from_secs(2),
+    Duration::from_secs(4),
+];
 
 /// Resolves the request timeout from [`REQUEST_TIMEOUT_ENV`].
 ///
@@ -53,6 +60,40 @@ pub struct TerminologyClient {
 }
 
 impl TerminologyClient {
+    /// These terminology operations only read data, including those sent as POST.
+    /// Retry transient gateway/service failures, but preserve transport, parsing and
+    /// other HTTP errors. Each attempt retains the configured HTTP client timeout.
+    async fn send_with_retry(
+        &self,
+        request: impl Fn() -> RequestBuilder,
+    ) -> FhirPathResult<Response> {
+        let mut delays = GATEWAY_RETRY_DELAYS.into_iter();
+        loop {
+            let response = request()
+                .send()
+                .await
+                .map_err(|e| FhirPathError::NetworkError(e.to_string()))?;
+            if !matches!(
+                response.status(),
+                StatusCode::BAD_GATEWAY
+                    | StatusCode::SERVICE_UNAVAILABLE
+                    | StatusCode::GATEWAY_TIMEOUT
+            ) {
+                return Ok(response);
+            }
+            let Some(delay) = delays.next() else {
+                return Ok(response);
+            };
+            tracing::warn!(
+                status = %response.status(),
+                delay_ms = delay.as_millis(),
+                "Retrying terminology request after a transient HTTP failure"
+            );
+            drop(response);
+            tokio::time::sleep(delay).await;
+        }
+    }
+
     /// Creates a new terminology client
     ///
     /// # Arguments
@@ -62,6 +103,7 @@ impl TerminologyClient {
     ///
     /// The request timeout defaults to 30s and can be overridden with
     /// `FHIRPATH_TERMINOLOGY_TIMEOUT` (whole seconds; `0` disables it).
+    /// HTTP 502, 503 and 504 responses are retried up to three times with backoff.
     pub fn new(base_url: String, fhir_version: FhirVersion) -> Self {
         let mut builder = Client::builder();
         if let Some(timeout) = request_timeout() {
@@ -117,18 +159,18 @@ impl TerminologyClient {
         }
 
         let response = self
-            .client
-            .get(&url)
-            .query(
-                &query_params
-                    .iter()
-                    .map(|(k, v)| (k.as_str(), v.as_str()))
-                    .collect::<Vec<_>>(),
-            )
-            .header("Accept", "application/fhir+json")
-            .send()
-            .await
-            .map_err(|e| FhirPathError::NetworkError(e.to_string()))?;
+            .send_with_retry(|| {
+                self.client
+                    .get(&url)
+                    .query(
+                        &query_params
+                            .iter()
+                            .map(|(k, v)| (k.as_str(), v.as_str()))
+                            .collect::<Vec<_>>(),
+                    )
+                    .header("Accept", "application/fhir+json")
+            })
+            .await?;
 
         if response.status().is_success() {
             response
@@ -187,14 +229,14 @@ impl TerminologyClient {
         }
 
         let response = self
-            .client
-            .post(&url)
-            .json(&body)
-            .header("Content-Type", "application/fhir+json")
-            .header("Accept", "application/fhir+json")
-            .send()
-            .await
-            .map_err(|e| FhirPathError::NetworkError(e.to_string()))?;
+            .send_with_retry(|| {
+                self.client
+                    .post(&url)
+                    .json(&body)
+                    .header("Content-Type", "application/fhir+json")
+                    .header("Accept", "application/fhir+json")
+            })
+            .await?;
 
         if response.status().is_success() {
             response
@@ -283,14 +325,14 @@ impl TerminologyClient {
         });
 
         let response = self
-            .client
-            .post(&url)
-            .json(&body)
-            .header("Content-Type", "application/fhir+json")
-            .header("Accept", "application/fhir+json")
-            .send()
-            .await
-            .map_err(|e| FhirPathError::NetworkError(e.to_string()))?;
+            .send_with_retry(|| {
+                self.client
+                    .post(&url)
+                    .json(&body)
+                    .header("Content-Type", "application/fhir+json")
+                    .header("Accept", "application/fhir+json")
+            })
+            .await?;
 
         if response.status().is_success() {
             let result: Value = response
@@ -360,14 +402,14 @@ impl TerminologyClient {
         });
 
         let response = self
-            .client
-            .post(&url)
-            .json(&body)
-            .header("Content-Type", "application/fhir+json")
-            .header("Accept", "application/fhir+json")
-            .send()
-            .await
-            .map_err(|e| FhirPathError::NetworkError(e.to_string()))?;
+            .send_with_retry(|| {
+                self.client
+                    .post(&url)
+                    .json(&body)
+                    .header("Content-Type", "application/fhir+json")
+                    .header("Accept", "application/fhir+json")
+            })
+            .await?;
 
         if response.status().is_success() {
             response
@@ -432,14 +474,14 @@ impl TerminologyClient {
         });
 
         let response = self
-            .client
-            .post(&url)
-            .json(&body)
-            .header("Content-Type", "application/fhir+json")
-            .header("Accept", "application/fhir+json")
-            .send()
-            .await
-            .map_err(|e| FhirPathError::NetworkError(e.to_string()))?;
+            .send_with_retry(|| {
+                self.client
+                    .post(&url)
+                    .json(&body)
+                    .header("Content-Type", "application/fhir+json")
+                    .header("Accept", "application/fhir+json")
+            })
+            .await?;
 
         if response.status().is_success() {
             response
@@ -478,14 +520,14 @@ impl TerminologyClient {
         let body = build_translate_body(concept_map_url, system, code, target_system, params);
 
         let response = self
-            .client
-            .post(&url)
-            .json(&body)
-            .header("Content-Type", "application/fhir+json")
-            .header("Accept", "application/fhir+json")
-            .send()
-            .await
-            .map_err(|e| FhirPathError::NetworkError(e.to_string()))?;
+            .send_with_retry(|| {
+                self.client
+                    .post(&url)
+                    .json(&body)
+                    .header("Content-Type", "application/fhir+json")
+                    .header("Accept", "application/fhir+json")
+            })
+            .await?;
 
         if response.status().is_success() {
             let result: Value = response
@@ -574,6 +616,128 @@ fn build_translate_body(
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    async fn call_operation(client: &TerminologyClient, operation: &str) -> FhirPathResult<Value> {
+        match operation {
+            "expand" => client.expand("http://example.org/vs", None).await,
+            "lookup" => client.lookup("http://example.org/cs", "x", None).await,
+            "validate_vs" => {
+                client
+                    .validate_vs("http://example.org/vs", None, "x", None, None)
+                    .await
+            }
+            "validate_cs" => {
+                client
+                    .validate_cs("http://example.org/cs", "x", None, None)
+                    .await
+            }
+            "subsumes" => {
+                client
+                    .subsumes("http://example.org/cs", "x", "y", None)
+                    .await
+            }
+            "translate" => {
+                client
+                    .translate(
+                        "http://example.org/cm",
+                        "http://example.org/cs",
+                        "x",
+                        None,
+                        None,
+                    )
+                    .await
+            }
+            _ => panic!("unknown test operation: {operation}"),
+        }
+    }
+
+    async fn stub_responses(statuses: Vec<u16>, body: &str) -> wiremock::MockServer {
+        use std::sync::atomic::{AtomicUsize, Ordering};
+        use wiremock::{Mock, MockServer, Request, ResponseTemplate};
+
+        let server = MockServer::start().await;
+        let calls = AtomicUsize::new(0);
+        let body = body.to_owned();
+        Mock::given(|_: &Request| true)
+            .respond_with(move |_: &Request| {
+                let index = calls.fetch_add(1, Ordering::SeqCst).min(statuses.len() - 1);
+                ResponseTemplate::new(statuses[index]).set_body_string(body.clone())
+            })
+            .mount(&server)
+            .await;
+        server
+    }
+
+    #[tokio::test]
+    async fn terminology_retries_transient_gateway_errors_for_all_operations() {
+        for operation in [
+            "expand",
+            "lookup",
+            "validate_vs",
+            "validate_cs",
+            "subsumes",
+            "translate",
+        ] {
+            let body = json!({"resourceType": "Parameters", "parameter": []});
+            let server = stub_responses(vec![502, 503, 504, 200], &body.to_string()).await;
+            let client = TerminologyClient::new(server.uri(), FhirVersion::R4);
+
+            assert_eq!(
+                call_operation(&client, operation).await.unwrap(),
+                body,
+                "{operation}"
+            );
+            let requests = server.received_requests().await.unwrap();
+            assert_eq!(requests.len(), 4, "{operation}");
+            for request in &requests[1..] {
+                assert_eq!(request.method, requests[0].method);
+                assert_eq!(request.url, requests[0].url);
+                assert_eq!(request.body, requests[0].body);
+                assert_eq!(request.headers, requests[0].headers);
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn terminology_retry_limit_preserves_final_error() {
+        let server = stub_responses(vec![502, 503, 502, 504], "gateway unavailable").await;
+        let client = TerminologyClient::new(server.uri(), FhirVersion::R4);
+        let error = client
+            .expand("http://example.org/vs", None)
+            .await
+            .unwrap_err();
+        assert!(matches!(&error, FhirPathError::TerminologyError(message)
+            if message.contains("504 Gateway Timeout") && message.contains("gateway unavailable")));
+        assert_eq!(server.received_requests().await.unwrap().len(), 4);
+    }
+
+    #[tokio::test]
+    async fn terminology_does_not_retry_other_http_errors() {
+        for status in [400, 401, 403, 404, 422, 429, 500] {
+            let server = stub_responses(vec![status, 200], "operation failed").await;
+            let client = TerminologyClient::new(server.uri(), FhirVersion::R4);
+            assert!(matches!(
+                client.expand("http://example.org/vs", None).await,
+                Err(FhirPathError::TerminologyError(_))
+            ));
+            assert_eq!(
+                server.received_requests().await.unwrap().len(),
+                1,
+                "HTTP {status}"
+            );
+        }
+    }
+
+    #[tokio::test]
+    async fn terminology_does_not_retry_invalid_success_body() {
+        let server = stub_responses(vec![200], "invalid JSON").await;
+        let client = TerminologyClient::new(server.uri(), FhirVersion::R4);
+        assert!(matches!(
+            client.expand("http://example.org/vs", None).await,
+            Err(FhirPathError::ParseError(_))
+        ));
+        assert_eq!(server.received_requests().await.unwrap().len(), 1);
+    }
 
     /// Names of the `Parameters.parameter` entries, in order.
     fn param_names(body: &Value) -> Vec<&str> {
--- a/crates/fhirpath/src/terminology_client.rs
+++ b/crates/fhirpath/src/terminology_client.rs
@@ -4,7 +4,7 @@
 //! It implements the standard FHIR terminology operations including expand, lookup,
 //! validate-code, subsumes, and translate.
 
-use reqwest::{Client, RequestBuilder, Response, StatusCode};
+use reqwest::{Client, RequestBuilder, Response};
 use serde_json::{Value, json};
 use std::collections::HashMap;
 use std::time::Duration;
@@ -73,12 +73,9 @@ impl TerminologyClient {
                 .send()
                 .await
                 .map_err(|e| FhirPathError::NetworkError(e.to_string()))?;
-            if !matches!(
-                response.status(),
-                StatusCode::BAD_GATEWAY
-                    | StatusCode::SERVICE_UNAVAILABLE
-                    | StatusCode::GATEWAY_TIMEOUT
-            ) {
+            // Cloudflare uses HTTP 530 for tunnel failures, including error 1033
+            // when no healthy cloudflared instance can receive the request.
+            if !matches!(response.status().as_u16(), 502 | 503 | 504 | 530) {
                 return Ok(response);
             }
             let Some(delay) = delays.next() else {
@@ -103,7 +100,7 @@ impl TerminologyClient {
     ///
     /// The request timeout defaults to 30s and can be overridden with
     /// `FHIRPATH_TERMINOLOGY_TIMEOUT` (whole seconds; `0` disables it).
-    /// HTTP 502, 503 and 504 responses are retried up to three times with backoff.
+    /// HTTP 502, 503, 504 and 530 responses are retried up to three times with backoff.
     pub fn new(base_url: String, fhir_version: FhirVersion) -> Self {
         let mut builder = Client::builder();
         if let Some(timeout) = request_timeout() {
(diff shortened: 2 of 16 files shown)


Gordex2014:
Visual walkthrough of the receipt pagination change, including a step-by-step comparison of OFFSET and keyset scans:

[PR #986: Receipt pagination](https://vanish.gordex.dev/s/golden-comet-ub/)

The temporary link expires on September 16, 2026 at 16:41 UTC.

smunini (approved):

Issue #995, opened 2026-09-08
Investigate bulk-submit memory peaks and retention across consecutive jobs


Gordex2014:
## Purpose

Investigate memory peaks and memory retained between consecutive bulk-submit jobs. This is an investigation based on observed measurements, **not a confirmed bug report or a claim of a memory leak**. No fix is prescribed.

During the controlled #979 local-output buffering experiment, high resident-memory samples occurred with both the unbuffered baseline and the buffered candidate. The unbuffered baseline had the highest absolute sample. The phenomenon therefore already existed without the buffering change in this comparison. This does not prove that buffering has no effect on memory, nor establish the cause of either distribution.

Related work: #979 and PR #996 buffer local artifact writes; #973 and PR #986 change receipt pagination; #982 proposes streaming receipt construction to limit working memory. This issue investigates measurement attribution and retention across consecutive jobs. It does not duplicate or implement #982's streaming design.

## Observed memory evidence

The 192,000-resource fresh case ran three paired trials in raw/buffered, buffered/raw, raw/buffered order. Exact values below are KiB; MiB values are rounded.

| Pair | Unbuffered maximum sample | Buffered maximum sample |
|---|---:|---:|
| 1 | 146,064 KiB / 142.6 MiB | 141,984 KiB / 138.7 MiB |
| 2 | 90,432 KiB / 88.3 MiB | 141,664 KiB / 138.3 MiB |
| 3 | 89,312 KiB / 87.2 MiB | 88,160 KiB / 86.1 MiB |

Unbuffered median: 90,432 KiB. Buffered median: 141,664 KiB. High samples occurred once unbuffered and twice buffered. The roughly 50 MiB difference between medians is not a measured per-job allocation caused by the buffer.

RSS means resident process memory, roughly the RAM occupied by HFS at a sampled instant. It is not a direct measurement of live allocations, ownership, a leak, or a guaranteed lifetime maximum.

The raw pair-1 maximum first appeared during artifact publication. Raw pair-2 and pair-3 maxima occurred during ingestion. The two high buffered maxima first appeared in the final HTTP-window sample after the named worker scope. These different peak locations complicate attribution; they do not invalidate or erase the higher buffered median.

All five cases, median [minimum, maximum] of sampled peak RSS in KiB:

| Case | Unbuffered | Buffered |
|---|---:|---:|
| 12000-fresh | 90,560 [87,712, 91,040] | 87,424 [85,872, 89,008] |
[...]

PR #996, opened 2026-09-08
Buffer local bulk output and publish submit artifacts atomically

### LINKED ISSUE #979: perf(local-fs): buffer bulk artifact writes before publication
## Problem

PostgreSQL bulk-submit profiling found a second finalization cost after receipt
readback. In the 192,000-resource fresh run, `artifact_write_and_register` took
5.098 seconds of the 15.211-second finalization stage. The preceding
`result_read` took 10.012 seconds and is tracked by #973. Fixing
pagination will not remove the local output cost.

The same capture attributed 6.223 CPU-seconds across all HFS threads to the
5.098-second artifact window. That total can exceed wall time because it sums
every process thread. Sampled stacks included 532 paths through
`tokio::fs::File`, `std::fs::File`, and file writes. This does not prove that
buffering will save 6.223 seconds, or that those leaf frames caused each
preceding CPU delta. Samply records `threadCPUDelta` since a thread's previous
sample, while the stack describes the current sample boundary.

The code makes buffering a reasonable, small experiment. The shared
[`ExportPartWriter::write_line`](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/core/bulk_export_output.rs#L104-L111)
issues two awaited `write_all` calls for every NDJSON line, one for the payload
and one for the newline. The
[`LocalFsOutputStore::open_writer`](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/local_fs/mod.rs#L69-L82)
passes a raw `tokio::fs::File` into that writer. Final publication already
follows the right order in
[`finalize_part`](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/local_fs/mod.rs#L84-L111):
flush, shutdown, drop, then rename the temporary file.

The five-case diagnostic run measured these stage times:

| Case | Result read | Artifact write/register |
|---|---:|---:|
| 12k fresh | 0.050 s | 0.258 s |
| 12k re-import | 0.045 s | 0.253 s |
| 48k fresh | 0.672 s | 1.134 s |
| 48k re-import | 0.599 s | 1.075 s |
| 192k fresh | 10.012 s | 5.098 s |

These are single observations from an instrumented release build. The local
write stage scales close to row count in this range. It grew 4.39x and 4.50x
when the fresh corpus grew by 4x. The table is a baseline for an experiment,
not a promised speedup.

## Proposed change

Wrap the local file in `tokio::io::BufWriter` inside
`LocalFsOutputStore::open_writer`. Start with a fixed 64 KiB capacity. Keep the
buffer at this backend boundary so both FHIR Bulk Data export and Bulk Data
Submit use it, while leaving S3 output behavior unchanged. Its current whole-file upload path is outside this issue.
Do not change `ExportPartWriter`, `ExportOutputStore`, receipt formats, or add a
new environment setting in the first version.

Keep `finalize_part`'s current publication contract. It must propagate flush
and shutdown errors and must not rename the `.tmp` file until both operations
succeed. `line_count` and `byte_count` must continue to count logical NDJSON
bytes, independent of buffered bytes waiting to be written. A failed line write
must not increment either counter.

64 KiB is a starting value, not a requirement to defend without data. Compare
it with the raw writer first; test another capacity if the initial result or memory budget warrants tuning. If the
difference is noise, close this issue without carrying a new layer or tuning
knob.

## Tests

- Extend the local filesystem round-trip test to compare exact output bytes,
  including one newline per record and the final trailing newline. Assert the
  reported line and byte counts.
- Add a small fake `AsyncWrite` test for short writes and injected write errors.
  Confirm `write_all` completes short writes and counters change only after the
  complete line plus newline succeeds.
- Inject flush and shutdown failures into finalization tests. Confirm the error
  reaches the caller and the final path is not published.
- Run the existing shared Bulk Data export and submit worker cases. This change
  must preserve manifests, download behavior, output/error receipts, fencing,
  and cleanup.

Use one build job and explicit backend features. Do not use `--all-features`:

```bash
CARGO_BUILD_JOBS=1 cargo test --locked -p helios-persistence --lib \
  --no-default-features --features R4,postgres \
  backends::local_fs::tests::test_write_finalize_read_delete -- --exact

CARGO_BUILD_JOBS=1 cargo test --locked -p helios-persistence --lib \
  --no-default-features --features R4,sqlite \
  core::bulk_export_worker::tests::worker_integration::test_run_job_system_export_end_to_end -- --exact

CARGO_BUILD_JOBS=1 cargo test --locked -p helios-persistence --lib \
  --no-default-features --features R4,sqlite \
  core::bulk_submit_worker::tests::test_worker_ingests_output_and_records_artifacts -- --exact
```

## Benchmark and acceptance criteria

Use the same deterministic 12k, 48k, and 192k synthetic corpus as the baseline.
Run PostgreSQL 16 with a 2-CPU, 2-GiB Docker limit, bulk batch size 100, worker
concurrency 1, file concurrency 4, local output, authentication disabled, and
deferred indexing disabled. Compare the raw writer with each buffer capacity.
Run three trials per case after a warm-up and report every trial plus median.

Measure worker, ingestion, result-read, receipt-serialization, and
artifact-write/register wall time separately. Record HFS process CPU and, where
available, write syscall counts. Include final flush and shutdown in the artifact-stage measurement so deferred work is not reported as a saving. The change succeeds only if that complete stage improves repeatably.
`result_read` should remain unchanged within run variance.

For every trial, compare terminal status, resource state, receipt line counts,
resource-ID sets, exact artifact bytes, reported byte counts, and absence of
leftover temporary files. Use generated patient data with no clinical or other
sensitive content. Keep raw profiles, logs, and generated artifacts in the
external investigation directory; publish only aggregate measurements and
sanitized reproduction details unless the raw files are reviewed separately.

This issue can land independently of #973. Testing both changes
together is still useful because pagination is expected to expose local output
as a larger share of finalization.

## Evidence provenance and repository policy

Profiling was completed on our end on 2026-09-07 using source `acb9284a75f076ca866be3989d93dbc2041716ed` plus temporary opt-in instrumentation. That instrumentation and the local summary were archived outside the repository, and the working tree was restored clean before implementation. No performance fix or after measurement exists yet.

#973 embeds the complete baseline table, exact synthetic generator, input hashes, HTTP reproduction flow, build/runtime settings, query-plan results, and archive integrity hashes. Those issue details are available to an implementing agent without access to the original conversation. Raw captures and executable/debug-symbol archives are preserved locally but have not been uploaded as attachments. Keep implementation code and tests in the repository; use issues/PRs and external attachments for plans, profiles, charts, generated fixtures and dated before/after evidence.

### LINKED ISSUE #980: bulk-submit: make artifact publication idempotent and lease-safe
## Problem

Bulk-submit finalization writes each output artifact, registers its database row, and only then marks the manifest complete. The storage contract describes registration as an idempotent fenced upsert, but the PostgreSQL and SQLite implementations first check the lease and then execute an unconditional insert. Artifact rows also identify the source manifest only by its nullable `manifest_url`, and status reads return all rows for the submission without selecting the winning fencing generation.

This leaves the publication boundary underspecified. A retry by the same lease can register duplicate rows. A worker that loses its lease after finalizing one part can leave a partial generation registered before a replacement worker publishes another generation. The fencing token protects artifact object names, but it does not by itself select one complete set for the status manifest. The worker also ignores `LeaseLost` from `finish_manifest`: only storage errors are returned, so a stale worker can continue into deferred reindexing.

These are findings from source inspection at baseline commit [`acb9284a`](https://github.com/HeliosSoftware/hfs/commit/acb9284a75f076ca866be3989d93dbc2041716ed):

- [`SubmitFileRecord` has no `manifest_id`, while the trait promises an idempotent fenced upsert](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/core/bulk_submit_worker.rs#L124-L247).
- [The worker registers parts one at a time, then completes the manifest](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/core/bulk_submit_worker.rs#L1005-L1066), and [does not handle `LeaseLost` before deferred reindex](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/core/bulk_submit_worker.rs#L870-L913).
- [PostgreSQL performs a fence read followed by a plain insert](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/postgres/bulk_submit.rs#L1667-L1724), and [its status read returns every artifact row for the submission](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/postgres/bulk_submit.rs#L1994-L2034). SQLite follows the same insert shape.
- [The SQL schema has no manifest identity or uniqueness constraint on artifact rows](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/postgres/schema.rs#L794-L824).
- [Artifact object keys already include the fencing token](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/core/bulk_export_output.rs#L16-L76).

No publication fault occurred in the five successful PostgreSQL diagnostic worker runs. This issue therefore describes a statically identified recovery gap, not an observed production incident. Add a deterministic failing test before selecting the implementation.

## Proposed contract

Define publication as one fenced state transition for one manifest generation. Finalized bytes may exist before that transition, but status must expose either the previous committed state or the complete winning set, never a partial or stale set.

Illustrative API; exact names may change:

```rust
struct SubmitFileRecord {
    manifest_id: String,
    manifest_url: Option<String>,
    file_type: String,
    resource_type: Option<String>,
    part_index: u32,
    file_path: String,
    line_count: u64,
    byte_count: u64,
    count_severity: Option<Value>,
}

async fn publish_manifest_artifacts(
    &self,
    lease: &ManifestLease,
    files: &[SubmitFileRecord],
    terminal: ManifestPublicationStatus,
) -> Result<(), LeaseError>;
```

`ManifestPublicationStatus` should carry either completed state or failed state with its error message. This lets manifest-fetch failures and other terminal error-artifact paths use the same fenced publication rule instead of retaining the current register-then-fail gap.

For PostgreSQL and SQLite, implement this operation in one transaction. Define the same-generation retry result explicitly: if an identical set is already committed for that manifest/token, return the committed result even though the first publication released the active lease. A different set for an already committed generation must be rejected without overwriting it. For a publication not yet committed:

1. Lock or conditionally select the manifest by tenant, submitter, submission, `manifest_id`, worker ID, and fencing token.
2. Return `LeaseLost` if ownership no longer matches.
3. Replace or upsert the artifact set for that manifest and generation using a declared identity such as `(tenant_id, submitter, submission_id, manifest_id, file_type, resource_type, part_index, fencing_token)`. Handle nullable resource type deliberately; do not rely on ordinary nullable uniqueness semantics.
4. Apply the requested completed or failed terminal state and release the lease in the same transaction.
5. Make status reads select only artifacts belonging to the committed winning generation.

Finalizing files cannot participate in a SQL transaction. If a stale worker writes a token-qualified object and then loses the lease, that object may remain orphaned, but it must never become successful visible output. Cleanup can be best-effort and separately observable.

`CompositeSubmitJobs` currently performs best-effort secondary synchronization inside `finish_manifest` before delegating completion. Preserve that ordering: run secondary sync while the lease is live, then invoke the primary's atomic publication transition. A secondary indexing failure must retain its existing `$reindex` recovery behavior and must not create a partial artifact publication.

## Backend scope and compatibility

The first implementation should provide the strong transactional guarantee for PostgreSQL and SQLite. Both need an explicit schema migration and migration tests. MongoDB and S3 already use generation-aware artifact identities differently; keep them compiling and preserve their current behavior, but do not claim an all-backend atomic guarantee unless equivalent takeover tests pass there. If the shared trait changes, adapters or backend-specific implementations are required rather than silently weakening the SQL contract.

It is a prerequisite for receipt streaming in #982. It does not depend on the keyset work in #973 or local writer experiment in #979. The ingestion investigation #981 is separate.

Related issues:

- #969 covers progress reset and heartbeat starvation during ingestion. Reuse its lease tests, but do not duplicate that scope here; this issue covers lease loss at completion/publication and stale-worker reindexing.
- #968 covers user-requested abort and UI feedback. Abort semantics are out of scope.

## Tests

Add failure injection at the worker/output-store and job-store seams before implementing the fix:

- fail after the first part is finalized;
- fail artifact registration or the terminal transaction;
- retry publication with the same fencing token;
- expire and reclaim the lease between finalization and publication;
- return `LeaseLost` from completion with deferred indexing enabled;
- publish completed output-only, completed error-only, mixed output/error, and failed-manifest error sets;
- verify tenant, submission, and manifest isolation.

Assertions must inspect terminal status and the exact visible artifact set: no duplicates, no partial set, no losing token, and no reindex call from the stale worker. Also assert that the live replacement can publish successfully after the injected failure.

Relevant existing tests and commands:

```bash
CARGO_BUILD_JOBS=1 cargo test --locked -p helios-persistence --no-default-features --features R4,sqlite \
  core::bulk_submit_worker::tests::test_worker_ingests_output_and_records_artifacts

CARGO_BUILD_JOBS=1 cargo test --locked -p helios-persistence --no-default-features --features R4,sqlite \
  --test sqlite_bulk_submit_concurrency

CARGO_BUILD_JOBS=1 cargo test --locked -p helios-persistence --no-default-features --features R4,postgres \
  --test postgres_tests postgres_bulk_submit_batch_commits_bookkeeping_and_contains_errors
```

Extend SQLite's `test_record_and_delete_artifacts`; it currently covers one insert, unlike the existing MongoDB and S3 same-generation idempotence tests. Add PostgreSQL-specific migration/publication coverage and retain `a_file_slower_than_the_lease_stays_leased_to_completion`, `a_ready_heavy_stream_still_renews_the_lease`, and `a_lost_lease_aborts_the_run_quietly`.

## Acceptance criteria

- [ ] A failing recovery test demonstrates the current static gap before the fix.
- [ ] PostgreSQL and SQLite publish one complete manifest artifact set atomically with completed or failed terminal state.
- [ ] Same-token retry is idempotent.
- [ ] A reclaimed/stale generation is not visible in status and cannot complete or trigger deferred reindex.
- [ ] Output, error, severity, line, and byte counts remain unchanged.
- [ ] SQL migrations are backward-compatible and tested on an existing schema.
- [ ] Supported non-SQL backends still compile and retain documented behavior.
- [ ] Raw profiles, database captures, binaries, and generated artifacts remain outside the source repository.

## Non-goals

This issue does not optimize deep OFFSET receipt reads, stream receipt construction, buffer local writes, change ingestion batch size, solve heartbeat starvation, define abort semantics, or promise removal of orphaned bytes. Performance measurements may verify that publication does not regress finalization, but no speedup is expected or claimed.

## Evidence provenance and repository policy

Profiling was completed on our end on 2026-09-07 using source `acb9284a75f076ca866be3989d93dbc2041716ed` plus temporary opt-in instrumentation. That instrumentation and the local summary were archived outside the repository, and the working tree was restored clean before implementation. No performance fix or after measurement exists yet.

#973 embeds the complete baseline table, exact synthetic generator, input hashes, HTTP reproduction flow, build/runtime settings, query-plan results, and archive integrity hashes. Those issue details are available to an implementing agent without access to the original conversation. Raw captures and executable/debug-symbol archives are preserved locally but have not been uploaded as attachments. Keep implementation code and tests in the repository; use issues/PRs and external attachments for plans, profiles, charts, generated fixtures and dated before/after evidence.

### LINKED ISSUE #981: bulk-submit(postgres): diagnose ingestion round trips and read amplification under the current batched writer
## Problem

The diagnostic PostgreSQL bulk-submit runs measured roughly 930-1,230 resources/s on a local synthetic workload, while the HFS process averages only about half a CPU core. This issue is a focused diagnostic child of #945. Its purpose is to identify the dominant current ingestion cost and demonstrate causality before proposing another optimization. Receipt finalization is tracked separately by #973; do not include that implementation here.

The current writer has changed materially since #945 was filed. Do not repeat the stale claim that every resource sends about 25 individual index-row inserts. At revision [`acb9284a`](https://github.com/HeliosSoftware/hfs/tree/acb9284a75f076ca866be3989d93dbc2041716ed), creates use a CTE plus array `UNNEST` for resource/history rows ([transaction.rs L291-L326](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/postgres/transaction.rs#L291-L326)), and search rows use a multi-resource array/`UNNEST` writer capped at 4,096 rows per statement ([writer.rs L1209-L1268](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/postgres/search/writer.rs#L1209-L1268)). The bulk path does, however, retain a savepoint and release for every entry inside one transaction per input batch ([bulk_submit.rs L656-L750](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/persistence/src/backends/postgres/bulk_submit.rs#L656-L750)). Those boundaries flush pending creates, so actual statement counts and array cardinalities must be measured rather than inferred from the batching APIs.

## Baseline evidence

A diagnostic capture on our end used PostgreSQL 16 in a 2-CPU/2-GiB container, native release HFS R4/PostgreSQL, inline indexing, one manifest worker, four concurrent files, local loopback input, and local output. The effective ingestion batch size was **100**. The REST configuration default remains **1,000** ([config.rs L671-L703](https://github.com/HeliosSoftware/hfs/blob/acb9284a75f076ca866be3989d93dbc2041716ed/crates/rest/src/config.rs#L671-L703)); changing that default is outside this issue.

| case | ingestion wall | rate | HFS CPU / avg cores | non-idle samples in `idle in transaction: ClientRead` | initial read |
|---|---:|---:|---:|---:|---:|
| 12k fresh | 9.761 s | 1,229/s | 5.362 s / 0.549 | 26/40 (65%) | 0.371 ms/entry |
| 12k reimport | 12.654 s | 948/s | 6.752 s / 0.534 | 36/48 (75%) | 0.371 ms/entry |
| 48k fresh | 44.883 s | 1,069/s | 21.546 s / 0.480 | 114/175 (65%) | 0.833 ms/entry |
| 48k reimport | 51.400 s | 934/s | 27.459 s / 0.534 | 149/201 (74%) | 0.373 ms/entry |
| 192k fresh | 194.596 s | 987/s | 92.616 s / 0.476 | 445/758 (59%) | 0.804 ms/entry |

The activity sampler ran once per second. Each row is a connection observation, not a duration or request. HFS CPU excludes PostgreSQL because it runs in another process. One run per case is insufficient for distributions.

Database-wide counters show an unexplained fresh-read amplification. The 48k fresh run added 80.4M returned and 78.1M fetched tuples, versus 4.0M and 3.6M on immediate reimport. The 192k fresh run added 273.0M returned and 261.5M fetched tuples. These counters include all connections and cannot yet be assigned to the per-entry point read. Post-capture catalog inspection found only `resources_pkey`, `idx_resources_updated`, and partial `idx_resources_search`. Forced generic and custom plans for present and missing IDs all used `resources_pkey` with fewer than ten execution buffer accesses. This excludes the obsolete `idx_resources_type` planner footgun from the current schema, but post-capture plans do not reconstruct plans used during ingestion.

Per-entry summed scopes also distinguish the paths. At 48k, fresh spent 0.104 ms in `create` and 1.436 ms in release/flush; reimport spent 2.247 ms in `update` and 0.330 ms in release. These are nested, process-global sums across four overlapping files, not additive wall time. Reimport ingestion was 14.5% slower at 48k.

## Questions to answer

1. Which normalized statements account for PostgreSQL calls, execution time, buffers, WAL, rows, and the fresh tuple amplification?
2. What are the actual resource and search-index array cardinalities? Does each savepoint reduce the multi-resource writer to one resource per call?
3. How much wall time belongs to transaction control, point reads, bookkeeping, create flushes, and update/index replacement, and is the limit PostgreSQL execution, PostgreSQL CPU/I/O, or statement feeding?

Test these hypotheses with explicit rejection conditions:

- **Per-entry round trips dominate.** Expect transaction-control, point-read, and receipt/change calls to account for most calls and waits. Reject if their cost is minor or reducing their count does not reduce ingestion time.
- **Savepoints collapse create batching.** Expect resource-array cardinality near one. Reject if flushes approach the configured batch or statement calls already scale with batches.
- **Update replacement explains reimport.** Expect update/history and index delete/reinsert to account for its extra work. Reject if their differential is too small or an experiment performing equivalent writes does not move the stage.
- **A plan transition explains fresh amplification.** Expect one statement's row/buffer deltas and changing plans to account for it. Reject if accounting locates the tuples elsewhere or row visits remain bounded.

Cache warmth and accumulated database state are confounders to control, not explanations to assume.

## Reproduction and measurement

The exact already-used generator, file hashes, kickoff JSON and runtime/build recipe are embedded in #973. Use that recipe to keep inputs identical. Generate deterministic FHIR R4 corpora of 12,000, 48,000, and 192,000 resources, each split equally into two Patient and two Observation NDJSON files. Restart line numbers in each file. Patients need deterministic IDs, identifier, name, gender, and birth date. Observations need deterministic IDs, `status=final`, vital-sign category, LOINC `29463-7`, Patient reference, effective time, and UCUM kg quantity. Serve them on loopback with `requiresAccessToken=false`.

Exercise the real asynchronous protocol: POST `/$bulk-submit`, POST `/$bulk-submit-status`, poll `Content-Location` from 202 to 200, follow terminal pagination, download all artifacts, and compare receipt references with the generated IDs. Record sanitized responses, timestamps, stage boundaries, corpus hashes, and resolved configuration.

Run this matrix one case at a time: 12k fresh then same-ID reimport; 48k fresh then same-ID reimport; 192k fresh only. Use a unique tenant per corpus. Fix worker concurrency at 1, file concurrency at 4, inline indexing, local output, PostgreSQL 16 with the same CPU/memory limits, and the observed worker batch size of 100. Verify that value from processing options or counters: the REST batch-size setting is not wired into the measured async worker, so setting its environment variable alone is insufficient. Preserve both the effective worker behavior and the REST default for this diagnosis.

Run **three trials per case from controlled database states**. Restore the same empty snapshot for each fresh trial. Clone its completed-fresh state for each reimport trial so rows and statistics start identically. Do not accumulate unrelated tenants.

Use `pg_stat_statements` deltas for calls, execution time, rows, shared/temp blocks, and WAL by normalized statement. Do not reset shared QA or production statistics. Capture PostgreSQL CPU/I/O, HFS CPU/RSS, and bounded `pg_stat_activity` classifications without bind values.

Capture actual load-time plans in a separate pass with sampled `auto_explain` or equivalent backend telemetry and parameter logging disabled. Do not run `EXPLAIN` per entry or mix this pass into performance timings. Use an uncommitted profiling-cfg patch to count flushes, statement families, array cardinalities, savepoints, and entries. Store raw artifacts outside the repository.

## Completion criteria

Publish the diagnostic report and machine-readable summaries in the issue/PR or external attachments, not as repository artifacts. They must:

- report the matrix, configuration, three-trial medians/spread, and exact receipt/resource reconciliation;
- account for statement calls and batch cardinalities without #945's old write shape;
- identify the fresh-amplification statement, or bound and rule out every tested source;
- demonstrate causality in one reversible experiment: predicted counter and stage-time reductions move together while integrity remains exact;
- separate HFS CPU, PostgreSQL CPU/I/O, awaited time, and any unattributed remainder.

This issue does not implement pagination, redesign receipt artifacts, change batch-size defaults, disable indexing, audit the broad PostgreSQL index set, tune production PostgreSQL, or compare SQLite. Use the result to open a narrowly scoped optimization issue with a before/after acceptance target.

Related: #945, #973. Dependencies: none; use the existing opt-in performance infrastructure with isolated diagnostic coverage.

## Evidence provenance and repository policy

Profiling was completed on our end on 2026-09-07 using source `acb9284a75f076ca866be3989d93dbc2041716ed` plus temporary opt-in instrumentation. That instrumentation and the local summary were archived outside the repository, and the working tree was restored clean before implementation. No performance fix or after measurement exists yet.

#973 embeds the complete baseline table, exact synthetic generator, input hashes, HTTP reproduction flow, build/runtime settings, query-plan results, and archive integrity hashes. Those issue details are available to an implementing agent without access to the original conversation. Raw captures and executable/debug-symbol archives are preserved locally but have not been uploaded as attachments. Keep implementation code and tests in the repository; use issues/PRs and external attachments for plans, profiles, charts, generated fixtures and dated before/after evidence.

Gordex2014:
This PR combines buffered local bulk-output writes with atomic bulk-submit artifact publication on SQLite and PostgreSQL. Local writers batch small NDJSON writes; submit workers publish the complete artifact set and completed or failed terminal state in one fenced transaction. This incorporates #1031, which fixes duplicate and partial artifact publication and prevents a stale worker from starting deferred reindexing.

It also incorporates the PostgreSQL ingestion diagnosis from #1035. The diagnostic report identifies repeated preparation of receipt/change statements as an avoidable cost; its temporary cache experiment is evidence for a future optimization and adds no production code.

Closes #979.
Closes #980.
Closes #981.

## Bulk-submit publication

SQLite and PostgreSQL recognize an exact same-token retry as `AlreadyPublished` and reject changed payloads or stale owners. Status queries select the committed generation. The worker renews its lease through finalization and composite synchronization, and starts deferred reindexing only after a fresh publication.

Manifest-aware artifact identities distinguish separate manifests and output/error files in download routes. SQL migrations preserve legacy records, deduplicate exact matches and exclude ambiguous or incomplete generations from visible output. Unambiguous legacy downloads remain supported. MongoDB and S3 retain their documented sequential publication behavior with manifest-aware identities.

Validation for this change covers deterministic recovery failures, exact replay, transaction rollback, lease expiry/reclaim, concurrent publication, cleanup rollback, existing-schema migrations and legacy downloads. Persistence/REST tests, real PostgreSQL and MongoDB tests, S3 mock contracts, Clippy and the HFS build passed. HTTP QA on both SQL backends verifies exact downloaded bytes and counts, distinct URLs, pagination, tenant isolation, failed-manifest publication and cleanup.

The transactional guarantee applies to SQLite and PostgreSQL. Finalized bytes may remain orphaned after lease loss. Durable exactly-once reindexing is outside scope, and composite secondary synchronization remains best-effort.

## Local output buffering

Local bulk output currently sends each NDJSON payload and newline through an unbuffered file writer. Add a private 64 KiB buffer in `LocalFsOutputStore::open_writer` so export and submit can combine small writes before publication. Finalization still flushes, shuts down, drops the writer and renames the temporary file before returning the published part. Counters advance only after a complete logical line is accepted. Buffering can defer I/O errors to finalization; those errors propagate before publication.

The receipt cursor dependency #986 has merged into `main` and is included in this branch. [Review the isolated buffering change](https://github.com/HeliosSoftware/hfs/commit/b7e4df1351b8952173d7010202146766fc748022). The buffering change preserves shared APIs and S3 behavior; the artifact-publication change from #1031 adds the shared publication API and manifest-aware identities described above.

## Buffering measurements

These measurements were collected for the buffering change before the atomic-publication change from #1031. They do not measure the combined implementation.

Thirty bulk-submit runs form 15 pairs covering 12k fresh/reimport, 48k fresh/reimport and 192k fresh resources. Each case uses three pairs in raw/buffered, buffered/raw, raw/buffered order, with the same corpus, logical starting snapshots, PostgreSQL 16.15, two CPUs, 2 GiB, one worker, four input tasks, effective batch 100 and inline indexing. Both binaries retain the dependency's receipt cursor fix and identical external stage instrumentation. The complete artifact stage includes open, write, flush, shutdown, drop, rename and file registration.

| Case | Raw artifact stage, median [min, max] ms | Buffered artifact stage, median [min, max] ms |
|---|---:|---:|
| 12k fresh | 276.314 [269.639, 291.110] | 5.304 [4.845, 5.492] |
| 12k reimport | 275.858 [265.016, 279.433] | 4.482 [4.408, 4.620] |
| 48k fresh | 1087.587 [1052.949, 1100.377] | 7.715 [6.324, 9.547] |
| 48k reimport | 1069.340 [1050.350, 1113.451] | 5.791 [5.664, 8.729] |
| 192k fresh | 4239.160 [4230.156, 5600.361] | 14.875 [10.655, 15.998] |

All 15 paired artifact deltas favor buffering. Each large case also has fully separated ranges, satisfying the predeclared adoption rule without additional trials. No small-case artifact or receipt-read regression guard triggers. These descriptive rules do not establish statistical significance or unchanged read performance.

Secondary medians below show raw → buffered. They describe the full experiment and are not attributed causally to the buffer.

| Case | Worker s | HTTP s | Receipt read ms | Process CPU s | Sampled RSS KiB |
|---|---:|---:|---:|---:|---:|
| 12k fresh | 15.120 → 14.951 | 16.827 → 16.589 | 84.356 → 88.533 | 5.99 → 5.57 | 90560 → 87424 |
| 12k reimport | 13.675 → 13.414 | 15.311 → 15.104 | 50.729 → 84.403 | 7.40 → 7.08 | 90592 → 87008 |
| 48k fresh | 49.228 → 52.011 | 50.834 → 53.687 | 93.975 → 100.085 | 23.54 → 22.25 | 99632 → 97328 |
| 48k reimport | 53.436 → 53.188 | 55.090 → 54.870 | 103.347 → 918.988 | 29.26 → 28.13 | 99664 → 96880 |
| 192k fresh | 182.006 → 169.868 | 183.596 → 171.444 | 555.351 → 596.491 | 94.73 → 89.47 | 90432 → 141664 |

Four individual worker/HTTP pairs are slower with buffering: 12k fresh pair 2, 48k fresh pairs 1 and 3, and 192k fresh pair 3. Their ingestion increases account for most of the observed wall-time increases. Receipt reads in 48k reimport take 0.919 and 0.951 s in two candidate trials versus about 0.10 s raw; the third candidate takes 0.089 s. All three 192k read deltas worsen, with overlapping ranges of 0.519–0.736 s raw and 0.588–0.741 s buffered. Causes are not established; these results do not support a general import-throughput guarantee.

Memory observations are tracked separately in investigation #995.

High memory samples also occurred in the unbuffered baseline: its maximum was 146,064 KiB, above the buffered maximum of 141,984 KiB. This observation predates the buffer within this controlled comparison. It is not a confirmed bug or memory leak, and the median difference does not establish buffer causation.

The 192k sampled RSS median increases from 90,432 to 141,664 KiB. Raw ranges from 89,312 to 146,064 KiB and buffered from 88,160 to 141,984 KiB; maxima occur in different phases. RSS sampled about every 0.2 s is not a guaranteed lifetime peak or an allocation measurement. The buffer adds 64 KiB per open local writer and does not bound whole-worker memory. Four input tasks do not imply four output buffers. Process CPU spans HTTP kickoff through terminal status and excludes database CPU. Stage spans overlap and include waits. Logical restores do not guarantee identical physical or cache state. Write syscall counts were not collected because the available probe required additional privileges.

Validation passed:

- Eleven focused contract tests, ten newly added, cover exact NDJSON, UTF-8 counts, empty output, buffer boundaries, oversized records, short writes, payload/newline errors, drain/flush/shutdown errors, publication order and cleanup.
- Export/submit worker suites, SQLite receipt and fencing checks, live PostgreSQL receipt integration, and REST export/submit suites passed. Explicit PostgreSQL and S3 library checks, formatting and Clippy with the planned lint allowances passed.
- All 30 primary runs preserve exact resources, metadata invariants, reference sets, manifests, downloaded bytes and counts, with no remaining temporary files. All 60 measured artifacts passed independent byte verification.
- Live HTTP QA covers submit output/error/deleted receipts and export download/delete. Four raw/buffered artifact comparisons are byte-identical. Preparation and QA attempts are retained separately; no primary trial failed, was excluded or retried.
- Independent code and final evidence reviews have no outstanding findings. Manifest validation and fallback-resume findings in external measurement tooling were fixed and rechecked before primary measurements. QA expectation corrections matched raw behavior and required no product changes.

The measured benefit applies to local artifact publication. Export compatibility is verified, but an export-specific throughput improvement was not measured.

## PostgreSQL ingestion diagnosis (#981)

[The diagnostic report in #1035](https://github.com/HeliosSoftware/hfs/pull/1035) contains the full analysis, all 30 accepted primary trials in CSV, and machine-readable summaries. Three controlled trials per arm cover 12k fresh/reimport, 48k fresh/reimport and 192k fresh. Eleven additional attempts were excluded solely for independently detected external load and retained in the external archive.

| 48k ingestion case | Control median | Experimental cache median | Time reduction |
|---|---:|---:|---:|
| Fresh import | 41.619 s | 35.571 s | 14.53% |
| Reimport | 53.159 s | 47.122 s | 11.36% |

All six 48k paired rounds improved. Reusing only the existing receipt and change statements reduced preparation requests from 96,000 to 12 fresh / 14 reimport, while the 96,000 execute calls and exact resource/receipt/history/change/search-index reconciliation stayed unchanged. Separate detailed passes show cumulative bookkeeping scopes falling by 44.18% / 45.47%; these overlapping scopes are not exclusive wall-time savings.

The current writer flushes one resource per nonempty create array. Historical fresh-read amplification was not reproduced. The report bounds current reads and records the separate 12k reimport version-check access path and its plan-state caveat. One non-target 12k reimport pair had a -0.08% reduction and remains in the results.

These measurements used [revision e1699e4](https://github.com/HeliosSoftware/hfs/commit/e1699e4966ba6f8bf9f4f16bd7281230bdd7c00d), before the #980 atomic-publication changes. They do not measure the combined implementation in this PR. The diagnostic patch was removed; the merge of #1035 adds no file changes and does not ship statement caching. The report proposes a follow-up optimization with at least 5% median improvement on both controlled 48k cases and unchanged integrity and SQL execution counts.

Diagnostic validation includes 5 PostgreSQL and 30 REST tests on the measured base, 19 HTTP-runner checks, 5 environment-guard checks, the diagnostic release build, and an evidence validator with zero failures.

[Interactive HTML explanation of the experimental performance improvements](https://vanish.gordex.dev/s/mellow-forest-4e/) is available until September 12, 2026 at 23:19 UTC. The report and embedded measurements remain available in #1035 after that temporary link expires.


Gordex2014:
Interactive visual explanation of this PR and its performance impact:

https://vanish.gordex.dev/s/mellow-stone-ai/

Covers the isolated buffering change, artifact publication timings, total worker time, receipt reads, and memory observations, using the reported benchmark data.

Temporary link, available for 6 days; expires September 15, 2026 at 21:58 UTC.

Gordex2014:
The atomic bulk-submit publication change from #1031 is now included in this PR.

[Visual walkthrough: atomic bulk-submit publication](https://vanish.gordex.dev/s/quiet-tempo-h9/)

The before/after comparison covers successful publication, file finalization failure, worker takeover, and identical retries. It also explains artifact identity and the SQLite/PostgreSQL scope. This walkthrough describes the publication change from #980; the buffering measurements for #979 remain in the PR description.

The temporary page expires on September 16, 2026 at 21:08 UTC.

Gordex2014:
Merged #1035 into this PR's branch and added the #981 diagnostic results to the description. This PR now targets closure of #979, #980 and #981 when merged into `main`.

[Interactive HTML explanation of the performance improvements, in English](https://vanish.gordex.dev/s/mellow-forest-4e/).

The experimental cache reduced median 48k ingestion time by 14.53% on fresh import and 11.36% on reimport. These measurements predate #980; the cache optimization is not shipped by this PR. The visual link expires on **September 12, 2026 at 23:19 UTC**. The full report and machine-readable measurements remain in #1035.

smunini (approved):

Issue #957, opened 2026-09-06
ui: "Status poll failed: error sending request for url" hides the cause, and a failed poll drops the Retry-After backoff


smunini:
## Summary

During a long bulk-submit import, the Submission Log intermittently records
`Status poll failed: error sending request for url (…)`. The message carries no cause —
`reqwest`'s `Display` stops at the URL and puts the real reason (timed out, connection
refused, connection reset) in the error's `source()`, which is dropped. The timing of
the failures points at the poll's own 10s timeout expiring while the status handler is
slow under ingest load, but the log as written cannot confirm that.

## Current behavior

Observed against a local recipient at `http://localhost:8080` while ingesting a large
Synthea corpus (log newest-first):

```
2026-09-06T15:21:24.967Z  Status poll failed: error sending request for url (http://localhost:8080/bulk-submit-status/162f5911-7b19-42f3-ae8b-76e0b0756a23)
2026-09-06T15:19:13.850Z  Status: processing 28% complete (4717935 entries ingested)
2026-09-06T15:19:01.715Z  Status poll failed: error sending request for url (http://localhost:8080/bulk-submit-status/162f5911-7b19-42f3-ae8b-76e0b0756a23)
2026-09-06T15:15:53.467Z  Status: processing 26% complete (4548835 entries ingested)
2026-09-06T15:13:37.669Z  Status: processing 25% complete (4386235 entries ingested)
...
2026-09-06T14:36:45.213Z  Status: processing 1% complete (315691 entries ingested)
2026-09-06T14:34:44.504Z  Status: processing 0% complete
2026-09-06T14:34:44.435Z  Bulk status kick-off request
2026-09-06T14:34:44.393Z  Manifest accepted by the recipient (200).
2026-09-06T14:34:44.358Z  Submitting manifest "http://localhost:8000/manifest.json"...
```

The message comes from `crates/ui/src/bulk_import.rs:897-901`:

```rust
Ok(r) => r,
Err(e) => {
    push_log(submission, format!("Status poll failed: {e}"));
    return;
}
```

`{e}` is `reqwest::Error`'s `Display`, which renders exactly
`error sending request for url (<url>)` for any transport-layer failure. The
distinguishing detail lives in `e.source()` (and in the `is_timeout()` /
`is_connect()` / `is_request()` predicates), none of which are consulted. So a timeout,
a refused connection, and a reset connection all produce a byte-identical log line.

### Why a timeout is the leading hypothesis

[...]

PR #1025, opened 2026-09-09
perf(sqlite): checkpoint the WAL at each bulk-submit file boundary

### LINKED ISSUE #978: persistence(sqlite): WAL auto-checkpoint starves under sustained bulk ingest — 7.3 GB WAL measured, reads degrade with depth
## Summary

Under sustained bulk ingest, SQLite's WAL auto-checkpoint (raised to 50,000 pages in #950) never catches up: measured live during a bulk load, the `-wal` file grew to **7.3 GB** (expected ceiling ~200 MB), and every read traversed the resulting ~1.8M-frame wal-index. Read latency scaled with ingest depth — the `$bulk-submit-status` poll went 0.1 s at 228k resources → 8 s at 605k → 13 s at 737k → 30 s+ at 3M — which is the trigger behind #957's poll timeouts, and the DELETE-cancel surface timed out outright (server 408 after 51 s).

Disk is doubled while it lasts (main file + giant WAL), and a post-crash/restart recovery of a 7 GB WAL takes ~50 s.

## Why

The automatic checkpoint runs on the committing connection after a transaction when the WAL exceeds the threshold, but it is passive: with back-to-back write transactions and concurrent readers it repeatedly yields and never reclaims. Raising the threshold in #950 made per-commit checkpoints cheaper (+32% ingest, table in that PR) but also made the backlog it can silently accumulate much larger. The two effects need separating: keep the throughput, cap the backlog.

## Proposed fix

Have the bulk-submit worker **force a checkpoint at manifest/file boundaries** (`PRAGMA wal_checkpoint(RESTART)` — or TRUNCATE — between output files, when no batch is holding the write lock), and optionally whenever the WAL exceeds a hard byte ceiling (e.g. 512 MB) checked from the worker between batches. Normal (non-bulk) operation keeps the 50k auto-checkpoint unchanged.

## Evidence

- WAL measured at 7,266 MB mid-ingest with `wal_autocheckpoint = 50000` (page cap ~200 MB).
- Poll-latency curve above, measured against a stock `main` build (the reproduction recorded on #957 / PR #972).
- `crates/persistence/src/backends/sqlite/backend.rs` — the #950 pragma pair and its doc comment, which assumed the ~200 MB ceiling holds.

Refs: #950 (merged — introduced the 50k threshold), #957/#972 (the visible symptom), #947 (SQLite ingest roadmap).

angela-helios:
Closes #978.

## The problem

Under a long bulk-submit ingest, SQLite's passive WAL auto-checkpoint keeps yielding to the back-to-back batch writers, so the WAL grows without bound — **measured at 7.3 GB mid-import on the real corpus** (the `-wal` file, vs the ~200 MB the `wal_autocheckpoint = 50000` threshold implies). A multi-gigabyte WAL slows every read — including the `$bulk-submit-status` poll, which is what surfaced as the poll timeouts in #957 — and doubles disk use for the duration.

## The fix

The submit worker folds the WAL back into the database at **each output-file boundary** — a point where no batch holds the write lock — via a new `SubmitWorkerStorage::checkpoint_after_file` hook:

- **SQLite** runs `PRAGMA wal_checkpoint(TRUNCATE)` on a `spawn_blocking` thread. A busy return (a reader still in a WAL frame) is fine — the next file boundary retries — so it is **best-effort and never fails the ingest**.
- **PostgreSQL, MongoDB, S3** keep the default no-op (no SQLite-style WAL).

No configuration — automatic and backend-aware, in the same spirit as the existing write-path pragmas. It does not change *how much* gets checkpointed over the run, only *when*: bounded pauses at file boundaries instead of an unbounded WAL.

## Tests

- New: a file-backed backend grows its WAL over 500 writes; `checkpoint_after_file` reclaims it (TRUNCATE → smaller `-wal`).
- The existing fan-out multi-file ingest suite stays green with the hook active (60 bulk_submit tests pass). clippy clean, fmt clean.

## Note on the follow-up validation

This was diagnosed live during a full 18.9M-resource import (the WAL growth and the ingest-rate decay it causes). An end-to-end before/after on the same corpus is the natural confirmation and can be posted on #978 once run.

### DIFF
--- a/crates/persistence/src/backends/sqlite/bulk_submit.rs
+++ b/crates/persistence/src/backends/sqlite/bulk_submit.rs
@@ -1967,6 +1967,25 @@ impl SubmitWorkerStorage for SqliteBackend {
         Ok(())
     }
 
+    async fn checkpoint_after_file(&self) {
+        // Fold the WAL back into the database at a file boundary (#978). The
+        // passive auto-checkpoint yields to the back-to-back batch writers and
+        // lets the WAL grow into the multi-gigabyte range over a long ingest,
+        // which slows every read and doubles disk use; a TRUNCATE checkpoint
+        // between files reclaims it while no batch holds the write lock. A
+        // busy return (a reader still in a WAL frame) is fine — the next file
+        // boundary tries again — so this is best-effort and never fails the
+        // ingest. Runs on a blocking thread so the checkpoint's I/O does not
+        // stall an async worker.
+        let backend = self.clone();
+        let _ = tokio::task::spawn_blocking(move || {
+            if let Ok(conn) = backend.get_connection() {
+                let _ = conn.execute_batch("PRAGMA wal_checkpoint(TRUNCATE);");
+            }
+        })
+        .await;
+    }
+
     async fn finish_manifest(&self, lease: &ManifestLease) -> Result<(), LeaseError> {
         // Losing this write to a busy would leave a fully ingested manifest
         // stuck in 'processing' until its lease expires and a worker redoes it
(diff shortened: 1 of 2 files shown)


angela-helios:
Live "before" evidence of the problem this fixes, captured importing the full 18.9M-resource corpus on the pre-fix binary (current main, no per-file checkpoint):

| Elapsed | Resources | Rate that hour | Disk consumed/hr |
|---|---|---|---|
| 1h | 1.54M | 429/s | ~15 GB |
| 2h | 2.73M | 330/s | ~15 GB |
| 3h | 3.47M | 205/s | ~13 GB |
| 4h | 4.08M | 165/s | ~20 GB |

The ingest rate decays monotonically (429 → 165/s) as the WAL grows unchecked and every commit does more work, and disk fills faster than the data warrants (the WAL doubling the footprint). Projected full import: >25h and climbing. With this PR's per-file `wal_checkpoint(TRUNCATE)`, the WAL is folded back between the 24 files, so the footprint stays bounded and the rate should hold. An end-to-end re-run on the same corpus with the fix is the clean before/after.

smunini (approved):

PR #1026, opened 2026-09-09
fix distinct-too-big paged intersection

### LINKED ISSUE #999: search(mongodb): multi-parameter searches 500 with 'distinct too big' (16 MB cap) when one parameter is high-cardinality
## Summary

On the `mongodb` backend, a search that combines a selective parameter with a high-cardinality one returns **500** after several minutes. The MongoDB `distinct` command that resolves each parameter exceeds its 16 MB BSON limit, and the whole search fails even when the final result is a handful of rows.

## Current behavior

Corpus of 11.2 M resources (827,968 Encounters, 7.7 M Observations). The query below is row 4.3 of `MANUAL_TESTING_MATRIX.md` verbatim; its expected result is "between 1 and 24".

```console
$ curl -s -o /dev/null -w 'HTTP %{http_code} time=%{time_total}s\n' \
    "localhost:8080/Encounter?patient=7d24f7a0-6f2e-ce3b-5568-db7b14695583&date=ge2016"
HTTP 500 time=413.4s
```

Server log:

```
ERROR helios_rest::error: internal error while processing request
  error.detail=internal error in mongodb: Failed to query search_index:
  Kind: Command failed: Error code 17217 (Location17217):
  Executor error during distinct command :: caused by :: distinct too big
```

Five of the matrix's own T4 queries fail this way, taking 28–109 s each to do so:

| row | query |
|---|---|
| 4.3 | `Encounter?patient=<PID>&date=ge2016` |
| 4.5 | `Observation?code=8302-2&value-quantity=gt150` |
| 4.24 | `Encounter?patient=<PID>&class=AMB` |
| 4.24 | `Encounter?patient=<PID>&date=ge2020` |
| 4.27 | `Observation?patient=<PID>&code=8302-2&value-quantity=gt120` |

Selective single-parameter searches are fine: `Observation?subject=Patient/<PID>` returns 165 in 0.8 s.

## Expected behavior

The search returns its result set. A parameter that matches many resources should not cap what the query can express — the intersection is what matters, and here it is 24 rows.

## Evidence

- `crates/persistence/src/backends/mongodb/search_impl.rs:832-842` — `distinct_resource_ids()` calls `search_index.distinct("resource_id", filter)` and collects into a `HashSet<String>`
- The multi-parameter path runs one such `distinct` per parameter and intersects the sets in memory, so a broad parameter (`date=ge2016` across 827,968 Encounters) returns hundreds of thousands of UUID strings and blows MongoDB's 16 MB command limit before the intersection ever happens
- MongoDB error 17217 is the documented `distinct` size cap

Related but distinct: #943 is the SQLite analogue (expression-tree depth on wide intermediate sets).

## Suggested approach

Resolve the intersection server-side instead of materializing each parameter's id set in the client:

- an aggregation pipeline over `search_index` that groups by `resource_id` and requires a match per parameter, or
- `$lookup`/`$in` staging that starts from the most selective parameter and filters the rest, or
- if the current shape is kept, page the `distinct` (cursor or bucketed ranges) and intersect incrementally, ordering by estimated selectivity so the smallest set is materialized first.

Any of these also removes the several-minute wait before the failure.

## Scope

MongoDB search only. The same corpus on this backend also shows plain slowness on very large result sets (`Observation?code=8302-2`, 175,355 results, 433 s) — that is throughput rather than a hard failure and is worth tracking separately.

bomanaps:
Replaces the single distinct call per search parameter with a selectivity-probed, batch-paged intersection strategy to eliminate the 16 MB BSON cap error (HTTP 500) on high-cardinality multi-parameter MongoDB searches (fixes #999).


dougc95 (changes requested):
Thanks for taking this on — and apologies for the slow review.

Before the findings, credit where it's due, because two things here are subtle and you got them right:

- **The set logic is correct.** I traced `:missing` true/false, `:not`, the compartment OR and the `Ok(None)` / `Ok(Some(empty))` boundaries against the code you replaced. The per-batch `retain` forms are equivalent intersections, and the bounded complement is exact rather than approximate — `candidates \ positive(candidates)` equals `(all \ positive) ∩ candidates` because `candidates` is already a subset of the type.
- **Your probe pipeline is `[$match, $limit, $group, $count]` — `$limit` before `$group`.** That's the right order, and it is *more* correct than the older copy of the same idea at `storage.rs:3458`, which is `[$match, $group, $limit, $count]` and therefore bounds its output rather than its input. Worth extracting the shared helper from yours and fixing that one.

I then measured the branch against the 11.2M-resource corpus rather than reasoning about it. That changed my view of one finding substantially, so the numbers lead.

---

## 1. The keyset page walks the global `_id` index — 235 s per page

`explain("executionStats")` on the driver page filter, against the real corpus (827,968 Encounters, `date=ge2016` matching 551,015 distinct ids):

```
find({tenant_id, resource_type:'Encounter', param_name:'date', value_date:{$gte:2016}})
  .sort({_id: 1}).limit(512)

  index         : _id_
  keysExamined  : 8,185,346
  docsExamined  : 8,185,346
  nReturned     : 512
  millis        : 235,166
```

There's no blocking sort — the planner satisfies `sort({_id:1})` from the default `_id` index and applies the real predicate as a **residual**. But `_id_` is global across the whole `search_index` collection, so the walk passes through Observation rows too (185.4M of them). 8.2 million keys examined to return 512 rows, and enumerating all 551,015 matches means walking toward the full ~204M-row collection.

Root cause: `_id` appears in none of the eleven `search_index` indexes — they're all `(tenant_id, resource_type, param_name, value_*)` plus two leading `(tenant_id, resource_type, resource_id, …)`. So `sort({_id:1})` can only be served by `_id_`, and `_id_` can't serve the predicate.

For comparison, the same predicate with **no sort**, projected and limited, uses `idx_search_date`:

```
  index         : idx_search_date
  keysExamined  : 2,049
  docsExamined  : 2,049
  millis        : 212
```

So dropping the sort and paging off the cursor's own end-of-stream — rather than a keyset on `_id` — is worth roughly three orders of magnitude here. The cursor is then the only state, which also removes findings 4 and 5 below for free.

## 2. The 16 MB wall is relocated, not removed

`confirmed` accumulates every matching id with no bound, and `build_resource_filter` still splices the whole set into `{id: {$in: […]}}` — a query document with its own 16 MB limit. Row 4.5 of the matrix, `Observation?code=8302-2&value-quantity=gt150`, intersects to roughly 400k ids; at ~45 bytes each that's ~18 MB and the `find` is rejected.

Measured cardinalities behind that, for sizing:

| parameter | distinct ids | vs the ~372k cap |
|---|---|---|
| `Observation value-quantity=gt150` | 423,382 | over |
| `Observation code=8302-2` | 175,355 | **under** |
| `Encounter date=ge2016` | 551,015 | over |
| `Encounter class=AMB` | 778,776 | over |

Note `code=8302-2` is *under* the cap at 175,355 — so in row 4.5 the parameter that raises 17217 is `value-quantity`, not `code`. Useful when writing a regression test for it.

Worth knowing: `Observation?code=8302-2` alone is recorded in the manual-test baseline as a client timeout, not a 500. Its `distinct` succeeds (175k ids ≈ 7.9 MB) and it then dies on the `$in` splice and resource fetch — i.e. the relocated wall is already reachable on `main` today, not just after this change.

Some bound is needed here. `SearchError::TooManyResults` already exists in `error.rs` and maps to 422, and nothing constructs it yet.

## 3. A silent wrong answer in the probe — this one I'd fix regardless

```rust
let count = docs.first().and_then(|d| d.get_i32("n").ok()).map(|n| n as i64).unwrap_or(0);
if count == 0 { return Ok(Some(HashSet::new())); }
```

`unwrap_or(0)` conflates "couldn't read the count" with "matches nothing", and the next line returns an empty restriction — so **every multi-parameter search answers 200 with zero resources.** `$count` is sugar for `$sum: 1`, whose accumulator type isn't contractually Int32, so an Int64 reply is enough to trigger it. Suggest distinguishing `docs.is_empty()` (genuinely no matches) from a present-but-unreadable `n`, which should propagate an error.

## 4. The probe measures distinct ids per *row* window, not rows to page

The window is 513 rows, but the cost the loop pays is rows to page. A parameter writing *k* index rows per resource probes to ~513/k, so a **broader** multi-row parameter looks *more* selective and wins the driver slot. This isn't marginal on real data: Observations carry **24 index rows per resource** (185,396,617 / 7,699,978) and Encounters **15.6** (12,960,211 / 827,985). A Period `date` alone writes two rows per resource (`converters.rs:467-474` pushes start and end).

Also, `best` only updates on strict `<`, so when two parameters both saturate at 513 the driver silently falls back to query order.

`count_documents(filter).limit(N)` counts index rows — the actual cost unit — and the driver's implementation puts `$limit` before `$group`, so it stays bounded.

## 5. No single-parameter fast path

With one normal parameter there's no intersection work to do, yet the query still pays a probe plus `ceil(rows/512)` sorted finds where one command sufficed. That's the commonest FHIR search shape, and it regresses relative to `main`.

## 6. `last_index_id = doc.get("_id").cloned()` can stall the loop

If `_id` is ever absent the keyset resets to page 1, the same 512 docs come back, and `docs_read < CANDIDATE_BATCH_SIZE` never fires — the request spins holding a connection. Unlikely given the projection always includes `_id`, but it's a hang rather than an error. Moot if the sort goes away per finding 1.

## 7. The complement-only path keeps the bug

`matching_resource_ids_complement_only` still calls `all_resource_ids` → `resources.distinct("id", …)`, which is the same 16 MB wall on the `resources` collection. So `Patient?gender:not=male` on a large type still fails, and the universe is re-fetched once per `:missing=true` and per `:not` parameter. Your comment acknowledges this; it's worth saying so in the PR body too so it isn't mistaken for fixed.

`idx_resources_type_scan` is `(tenant_id, resource_type, is_deleted, last_updated, id)`, so a streamed projected `find` there yields `id` straight from the index key.

## 8. Test coverage — the rewritten paths aren't covered

`mongodb_tests.rs:2598-2640` exercises `gender:missing=true`, `gender:missing=false` and `gender:not=male` each as the **only** parameter of its query. So `normal` is empty and all three route to `matching_resource_ids_complement_only` — the path this PR barely changes. The genuinely rewritten per-batch `:missing` / `:not` / compartment logic ships untested, and the two new tests use two plain token parameters with no modifier and no compartment.

Worth adding: a positive parameter combined with `:missing=true` and with `:missing=false`; one combined with `:not`; a compartment search combined with a positive parameter; and each sized past 512 driver rows so more than one batch actually runs. A multi-valued driver (identifier with two entries, or a Period date) would also cover one resource's rows straddling a batch boundary.

---

## Two things that affect any fix here, including this one

**The bare-id reference filter is a 10,000× penalty, and the matrix hits it.** `build_reference_filter` branches on whether the value contains `/`. Qualified is an index equality; bare emits an unanchored `$regex: "/X$"` from which no index bounds can be derived:

```
patient=<bare id>          keys: 827,986   docs: 41   millis: 123,293
patient=Patient/<bare id>  keys: 41        docs: 41   millis: 12
```

`docsExamined: 41` against `keysExamined: 827,986` shows the regex is evaluated in-index, so it's 828k sequential key reads rather than random fetches — but it's still the largest single constant in this issue. Matrix rows 4.3 and 4.12 both send the bare form, so any probe here measures `patient` as broad and may hand the driver to the wrong parameter. Separate fix, additive (keep today's unanchored arm as an extra `$or` branch so absolute-URL references still match), and I'll open it separately.

**No design gets a covered scan yet.** Every plan above shows `docsExamined > 0`, because `resource_id` is in none of the value-typed indexes. Appending it to their tails is the prerequisite for a genuinely covered walk — but `ensure_search_indexes` is awaited inline at startup with no timeout, so on a collection this size that build blocks the process from serving. It needs its own issue and a measured out-of-band window; I'll file it.

---

Requesting changes on 1, 2 and 3 — finding 1 is the one that decides whether this is a speedup, and finding 3 is a silent wrong answer independent of performance. Happy to talk through any of it, and the `explain` script I used is reusable if you want to measure variants yourself.

bomanaps:
Fixed here https://github.com/HeliosSoftware/hfs/pull/1026/commits/ab910c107c7fb92474383e38619acb5e05f6e0e9

dougc95:
Re-reviewed at `ab910c107`. **All three blocking findings are fixed**, and two of them are fixed better than I suggested. Thanks for turning this around quickly.

- **The `_id_` walk is gone.** `sort({_id:1})` and `last_index_id` are both removed; the driver is now one unsorted `find` cursor drained through `read_cursor_batch`. That lets the planner use `idx_search_date` instead of falling back to the global `_id` index, which was the 8.2M-keys-per-page problem. It also makes the keyset-stall finding moot by construction.
- **The relocated 16MB wall is capped**, and correctly in *both* places — the accumulation at `:1190` and the `{id: {$in: …}}` splice at `:1713`. Catching the second site matters; capping only the first would still let a large `matched_ids` arrive from elsewhere and blow the query document.
- **The probe count no longer fails open.** `get_i64` with a `get_i32` fallback, and an unreadable `n` now raises `internal_error(...)` rather than being read as zero matches. Erroring is the right call — better than the `docs.is_empty()` distinction I suggested.

Two things I'd still like to see, and one nit. None of them block.

---

## F4 — the probe still measures the wrong quantity

`search_impl.rs:1015-1048`. The pipeline at `:1021-1024` is bounded correctly, but what it *measures* is distinct resource-ids inside the first 513 **rows** — and the cost the loop actually pays is **rows to drain**.

A parameter writing *k* index rows per resource probes to ~`513/k`, so a **broader** multi-row parameter scores as *more* selective and wins the driver slot. This is not marginal on real data — measured on the 11.2M corpus:

| type | resources | search_index rows | rows/resource |
|---|---|---|---|
| Observation | 7,699,978 | 185,396,617 | **24.1** |
| Encounter | 827,985 | 12,960,211 | **15.6** |

A Period `date` alone writes two rows per resource (`search/converters.rs:467-474` pushes `start` and `end`). Separately, `best.is_none_or(|(_, prev)| count < prev)` at `:1044` only updates on a strict `<`, so when two parameters both saturate at 513 the driver silently falls back to query order.

`count_documents` with a bounded limit measures index **rows** — the real cost unit — and the driver's own implementation puts `$limit` before `$group`, so it stays bounded:

```rust
// search_impl.rs, replacing the aggregate probe at :1019-1043
use mongodb::options::CountOptions;

const PROBE_ROW_LIMIT: u64 = 100_000;

let n = search_index
    .count_documents(filter)
    .with_options(CountOptions::builder().limit(PROBE_ROW_LIMIT).build())
    .await
    .or_query_error("Failed to probe search_index for driver selection")?;
```

Two parameters that both reach `PROBE_ROW_LIMIT` still tie, but at a 200× higher threshold than 513 — and a deliberate tie-break (query order, or preferring the one whose filter has no unanchored `$regex`) is worth a line of comment either way.

## F5 — no single-parameter fast path

Much smaller than when I first raised it, because fixing the `_id_` walk removed most of its cost. What's left is one wasted aggregate round trip on the commonest FHIR query shape: with a single positive predicate there is nothing to rank and nothing to intersect against, so the probe is pure overhead — the inner verify loop at `:1076` no-ops for every batch anyway.

One line, at `:1015`:

```rust
// Nothing to rank when there is only one positive predicate: it is the
// driver by definition, and the per-batch verify loop below no-ops.
let driver_idx = if normal.len() == 1 {
    0
} else {
    // ... existing probe block unchanged ...
};
```

## F7 (nit) — the two probes now disagree

This PR's probe is `[$match, $limit, $group, $count]` (`search_impl.rs:1021-1024`) — `$limit` before `$group`, which bounds the *input*. Correct.

Its twin in `find_matching_resources_in_bundle_transaction` is `[$match, $group, $limit, $count]` (`storage.rs:3441-3444`) — `$limit` after `$group`. Since `$group` is blocking and there is no `$group`/`$limit` coalescence, that one bounds its *output* and consumes the parameter's entire slice.

It doesn't bite today because `ifNoneExist` parameters are identifiers, and it's pre-existing rather than introduced here. But this PR is what makes the two copies diverge, so it's worth a two-line swap while it's fresh:

```rust
// storage.rs:3440-3445
let pipeline = vec![
    doc! { "$match": filter },
    doc! { "$limit": PROBE_LIMIT },          // moved up
    doc! { "$group": { "_id": "$resource_id" } },
    doc! { "$count": "n" },
];
```

Extracting one shared helper would be better still, but that needs the `storage.rs` call sites to thread a session parameter, which is a bigger change than belongs here.

---

## One question on the cap

`MAX_RESULT_ID_SET = 300_000` (`:133`) sits **below** a cardinality this corpus actually has: `Encounter?date=ge2016` matches **551,015** distinct ids, measured. So a driver that legitimately matches that many now returns 422 rather than a result.

At ~45 bytes per id, 300k is roughly 13MB — inside the 16MB limit with about 70k ids of headroom. The question is whether any real query is expected to return more than 300k resources. Everything *measured* as working on this corpus is far below it (largest: 11,705), so this may well be the right trade — a fast, honest 422 where the server previously hung. Worth a comment recording the reasoning, so the next person doesn't tune it blind.

---

The remaining items are suggestions, not blockers — my standing "changes requested" is about the three findings above it, all of which are now addressed. Happy to re-approve once you've had a chance to look at F4.

bomanaps:
Fixed here https://github.com/HeliosSoftware/hfs/pull/1026/commits/c5f22c32bb231d3984e22eb3f6a9b88cb7d75667

dougc95 (changes requested):
Re-reviewed at `36d1b395b`. F4, F5 and F7 are all addressed, and the full `mongodb_tests` suite passes locally (137/137) including the new paged-intersection test. Thanks.

Only one thing left before I approve: CI's clippy runs with `-D warnings` and this branch introduces two lints that are not on its allow-list:

```
error: this expression creates a reference which is immediately dereferenced by the compiler
    --> crates/persistence/src/backends/mongodb/search_impl.rs:1383:85
error: this function has too many arguments (9/7)
    --> crates/persistence/src/backends/mongodb/search_impl.rs:1416:5
```

- `:1383` — `base` is already a `&str`, so `regex_escape(base)` instead of `regex_escape(&base)`.
- `:1416` — add `#[allow(clippy::too_many_arguments)]` to `matching_resource_ids_complement_only`, or drop the `has_compartment` parameter since it's derivable from `query`.

To check locally:

```
cargo clippy -p helios-persistence --all-targets --all-features -- -D warnings -A clippy::items_after_test_module -A clippy::large_enum_variant -A clippy::question_mark -A clippy::collapsible_match -A clippy::collapsible_if -A clippy::field_reassign_with_default -A clippy::doc-overindented-list-items -A clippy::doc-lazy-continuation
```

Will approve as soon as that lands.

dougc95 (approved):

smunini (approved):

PR #1031, opened 2026-09-09
Make bulk-submit artifact publication idempotent and lease-safe

### LINKED ISSUE #980: bulk-submit: make artifact publication idempotent and lease-safe

Gordex2014:
Bulk-submit could register duplicate or partial artifact sets, and a worker that lost its lease during completion could still start deferred reindexing. SQLite and PostgreSQL now publish the complete artifact set and its completed or failed state in one fenced transaction. An identical committed retry returns `AlreadyPublished`; changed retries and stale owners cannot overwrite the publication. Status reads select the committed generation, and deferred reindexing starts only after a fresh successful publication.

Artifacts now carry manifest identity and use scoped versioned locators, so separate manifests and output/error files cannot share a download route accidentally. The worker keeps renewing its lease through finalization and composite synchronization. SQL migrations retain legacy records, deduplicate exact matches and exclude ambiguous or incomplete generations. Unambiguous legacy downloads remain supported; ambiguous routes fail closed. MongoDB and S3 adapters retain their documented sequential behavior with manifest-aware identities.

Validation covers the original failing recovery scenarios, injected finalization and transaction failures, exact replay, lease expiry/reclaim, concurrent publication, cleanup rollback, existing-schema migrations, receipt counts and bytes, and legacy downloads. Focused and broader persistence/REST tests pass, including real PostgreSQL and MongoDB tests, S3 mock contracts and the existing lease-concurrency suite. Clippy and the HFS build pass. HTTP QA against both SQL backends verifies distinct artifact URLs, exact downloads, pagination, tenant isolation, failed-manifest publication and cleanup. Review found no unresolved blocking defect.

The atomic guarantee applies to SQLite and PostgreSQL. Finalized bytes can remain orphaned after lease loss; durable exactly-once reindexing is outside scope. Composite secondary synchronization retains its existing best-effort behavior.

Depends on #996 for the sequential branch stack.

Closes #980.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3678,6 +3678,7 @@ dependencies = [
  "rand 0.8.6",
  "regex",
  "reqwest",
+ "rusqlite",
  "serde",
  "serde_json",
  "serde_urlencoded",
--- a/crates/persistence/Cargo.toml
+++ b/crates/persistence/Cargo.toml
@@ -26,7 +26,7 @@ cassandra = ["dep:cdrs-tokio", "dep:cdrs-tokio-helpers-derive"]
 mongodb = ["dep:mongodb"]
 neo4j = ["dep:neo4rs"]
 elasticsearch = ["dep:elasticsearch"]
-s3 = ["dep:aws-sdk-s3", "dep:aws-config", "dep:aws-credential-types", "dep:sha2"]
+s3 = ["dep:aws-sdk-s3", "dep:aws-config", "dep:aws-credential-types"]
 
 # Configuration advisor binary
 advisor = ["dep:axum", "dep:tower-http", "dep:tracing-subscriber"]
@@ -96,8 +96,8 @@ elasticsearch = { version = "8.15.0-alpha.1", optional = true }
 aws-sdk-s3 = { version = "1", optional = true, default-features = false, features = ["sigv4a", "default-https-client", "rt-tokio"] }
 aws-config = { version = "1", optional = true }
 aws-credential-types = { version = "1", optional = true }
-# Hashes the per-user settings key into an opaque, fixed-length S3 object name.
-sha2 = { version = "0.10", optional = true }
+# Hashes opaque keys and submit artifact scopes into fixed-length safe names.
+sha2 = { version = "0.10" }
 
 # Audit event integration.
 #
--- a/crates/persistence/src/backends/mongodb/bulk_submit.rs
+++ b/crates/persistence/src/backends/mongodb/bulk_submit.rs
@@ -51,6 +51,9 @@ use crate::core::bulk_submit::{
     StreamProcessingResult, StreamingBulkSubmitProvider, SubmissionChange, SubmissionId,
     SubmissionManifest, SubmissionStatus, SubmissionSummary, invalid_entry_result_page,
 };
+use crate::core::bulk_submit_publication::{
+    ManifestPublicationResult, ManifestPublicationStatus, canonical_publication_files,
+};
 use crate::core::bulk_submit_worker::{
     ManifestFetchParams, ManifestLease, ManifestWorkerView, PollTokenTarget, SubmitClaimStrategy,
     SubmitFileRecord, SubmitFileRow, SubmitWorkerStorage,
@@ -1469,6 +1472,7 @@ impl SubmitWorkerStorage for MongoBackend {
         // token, so a retried write replaces its own row rather than adding a
         // duplicate entry to the status manifest.
         let mut key = submission_filter(&lease.tenant, &lease.submission_id);
+        key.insert("manifest_id", &lease.manifest_id);
         key.insert("file_type", &file.file_type);
         key.insert("resource_type", file.resource_type.as_deref());
         key.insert("part_index", file.part_index as i64);
@@ -1493,6 +1497,28 @@ impl SubmitWorkerStorage for MongoBackend {
             .await
             .map_err(|e| LeaseError::Storage(internal_error(format!("record submit file: {e}"))))?;
         Ok(())
+    }
+
+    /// Publishes by canonical validation, then fenced per-row writes and the
+    /// fenced terminal update. Mongo provides no atomic full-set publication;
+    /// this intentionally preserves the existing non-atomic backend behavior.
+    async fn publish_manifest_artifacts(
+        &self,
+        lease: &ManifestLease,
+        files: &[SubmitFileRecord],
+        terminal: ManifestPublicationStatus,
+    ) -> Result<ManifestPublicationResult, LeaseError> {
+        let canonical = canonical_publication_files(files).map_err(LeaseError::Storage)?;
+        for file in &canonical {
+            self.record_submit_file(lease, file).await?;
+        }
+        match terminal {
+            ManifestPublicationStatus::Completed => self.finish_manifest(lease).await?,
+            ManifestPublicationStatus::Failed { error_message } => {
+                self.fail_manifest(lease, &error_message).await?
+            }
+        }
+        Ok(ManifestPublicationResult::Published)
     }
 
     async fn finish_manifest(&self, lease: &ManifestLease) -> Result<(), LeaseError> {
@@ -1725,6 +1751,8 @@ impl SubmitWorkerStorage for MongoBackend {
                 resource_type: opt_str(d, "resource_type"),
                 part_index: d.get_i64("part_index").unwrap_or(0).max(0) as u32,
                 fencing_token: d.get_i64("fencing_token").unwrap_or(0).max(0) as u64,
+                manifest_id: opt_str(d, "manifest_id"),
+                legacy_locator: opt_str(d, "manifest_id").is_none(),
                 file_path: d.get_str("file_path").unwrap_or_default().to_string(),
                 line_count: d.get_i64("line_count").unwrap_or(0).max(0) as u64,
                 byte_count: d.get_i64("byte_count").unwrap_or(0).max(0) as u64,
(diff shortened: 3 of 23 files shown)


Gordex2014:
Visual walkthrough of this PR: [Atomic bulk-submit publication](https://vanish.gordex.dev/s/quiet-tempo-h9/).

The before/after comparison covers successful publication, file finalization failure, worker takeover, and identical retries. It also explains artifact identity and the SQLite/PostgreSQL scope. Built with HTML and CSS only.

This temporary page is available for 7 days, until September 16, 2026 at 21:08 UTC.

PR #1034, opened 2026-09-09
fix(search): wide two-level chained search 500s with 'Expression tree is too large' (#943)

### LINKED ISSUE #943: search(sqlite): two-level chained search 500s with 'Expression tree is too large (maximum depth 1000)' on wide intermediate sets
`GET /Observation?encounter.patient.gender=female` on the #448 benchmark dataset (~2,300 matching encounters) fails with HTTP 500:

```
Chained search resolution failed error=internal error in sqlite: Failed to prepare search query: Expression tree is too large (maximum depth 1000)
```

The chained-search resolver materializes the intermediate id set into the outer query; past ~1,000 terms SQLite refuses to prepare it. Reproduced identically on current `main`-equivalent code and on the partial-index branch — **pre-existing**, unrelated to index layout. Single-level chains and `_has` are unaffected on this dataset.

Likely fix: batch the id list (chunked OR/IN with UNION), or stage ids into a temp table and join.

angela-helios:
Fixes #943.

## Problem

A two-level chained search over a wide intermediate set — e.g. `GET /Observation?encounter.patient.gender=female` on a dataset with a few thousand matching encounters — returns **HTTP 500**:

```
Chained search resolution failed error=internal error in sqlite: Failed to prepare search query: Expression tree is too large (maximum depth 1000)
```

Two spots materialize the whole intermediate id/reference set into a single query as an **OR of one condition per value**. The SQLite fragment builder parenthesizes both operands of every OR (`(X) OR (Y)`), and SQLite counts each parenthesized expression as its own parse-tree node, so every term adds **~2** to the tree depth. Past ~500 terms the tree reaches SQLite's fixed 1000-deep parse limit and the statement fails to prepare.

## Fix

- **`chain_resolver` (backend-agnostic):** the intermediate reference hops now search the refs in **chunks** (`CHAIN_VALUE_CHUNK = 250`) and union the matches, so a hop's OR chain stays near ~500 deep no matter how wide the set is. Matches are de-duplicated between chunks so a parent referencing refs in two chunks does not inflate the next hop. Every backend benefits.
- **`sqlite` `_id` handler:** the resolver's final rewrite injects the resolved ids as an `_id` filter, which the backend built as `id = ? OR id = ? OR ...` (depth N). It is now a flat `id IN (?, ?, ...)` — a single node of any length. This also fixes any direct wide `_id=` search and the nested `_has` inner `_id` selection.

## Reproduction / test

Added `two_level_chain_over_wide_intermediate_set`: it seeds **1200** patient/encounter/observation triples (comfortably past the depth limit) and asserts `Observation?encounter.subject.gender=female` resolves and returns every match. The test **fails with the exact depth-1000 error before either change** and passes after.

## Verification

- `cargo test -p helios-persistence --lib` — **876 passed**, 0 failed (incl. the new regression test).
- `cargo clippy -p helios-persistence --lib` — clean (no new warnings in the changed files).

Why chunk size 250: with the OR builder's paren-per-operand doubling, 250 terms keeps a hop's tree near ~500 deep — half the 1000 limit, leaving margin for the surrounding query wrapping and any future per-value condition growth.

### DIFF
--- a/crates/persistence/src/backends/sqlite/search/parameter_handlers/reference.rs
+++ b/crates/persistence/src/backends/sqlite/search/parameter_handlers/reference.rs
@@ -293,9 +293,9 @@ mod tests {
         assert!(versioned >= base && versioned < upper.as_str());
         // excluded (a different id sharing the "123" prefix):
         let sibling = "Patient/1234";
-        assert!(!(sibling < upper.as_str()));
+        assert!(sibling >= upper.as_str());
         let sibling0 = "Patient/1230";
-        assert!(!(sibling0 < upper.as_str()));
+        assert!(sibling0 >= upper.as_str());
     }
 
     #[test]
--- a/crates/persistence/src/backends/sqlite/search/query_builder.rs
+++ b/crates/persistence/src/backends/sqlite/search/query_builder.rs
@@ -523,30 +523,33 @@ impl QueryBuilder {
     ) -> Option<SqlFragment> {
         match param.name.as_str() {
             "_id" => {
-                // _id searches directly on the resources table
-                let mut conditions = Vec::new();
-                for (i, value) in param.values.iter().enumerate() {
-                    conditions.push(SqlFragment::with_params(
-                        format!("id = ?{}", param_offset + i + 1),
-                        vec![SqlParam::string(&value.value)],
-                    ));
-                }
-
-                if conditions.is_empty() {
+                // _id searches directly on the resources table. Build a flat
+                // `id IN (?, ?, ...)` rather than a chain of `id = ? OR id = ?`:
+                // a nested OR of N terms parses to a tree of depth N, and SQLite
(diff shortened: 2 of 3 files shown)


smunini (approved):

angela-helios:
Manual-test note for review: verified this fix on the sqlite MANUAL_TESTING_MATRIX pass (T4) against a ~512k-resource corpus. Correctness is good — 2-level chains return the right results and no longer crash. But on a **wide** intermediate set the chunked resolution is slow and can saturate the server (e.g. `Observation?encounter.patient.gender=female`, ~5,800 patients → ~49k intermediate encounters, did not return in 240s and made `/health` time out). Root cause is that multi-value reference search is O(N) and bypasses the index, which this fix's 250-value chunks inherit — filed separately as #1052 with measurements and a suggested `value_reference IN (...)` fix. This PR's correctness stands on its own; flagging so the perf follow-up is on the radar.

PR #1114, opened 2026-09-14
perf(sqlite): synchronous=NORMAL under WAL for bulk-ingest commit cost (#947)

### LINKED ISSUE #947: bulk-submit(sqlite): per-entry pipeline dominates ingest (~84%) — instrument and cut the non-index write volume

angela-helios:
## What

Sets `PRAGMA synchronous = NORMAL` on file-backed SQLite connections when WAL is enabled, cutting per-batch commit fsync cost on the bulk-submit ingest path.

## Why

Phase instrumentation on the real 31 GB manifest (branch `diag/947-phase-timers`) attributed **~45% of the ingest wall to per-batch commit fsync** and ~26.5% to `search_index_insert`. The multi-row search-index insert (lever #1) is already on `main`; this PR lands the remaining commit-cost lever.

Under WAL, `synchronous=NORMAL` fsyncs only at checkpoint rather than on every commit, and stays corruption-safe: a crash can lose transactions committed since the last checkpoint but never the database itself (SQLite's documented, recommended level for WAL).

## Scope / safety

- Applied **only when WAL is enabled** — `NORMAL` under a rollback journal is not crash-safe.
- Skipped for **in-memory** databases (they never fsync).
- Global default; no new configuration knob.

## Tests

- New `file_backend_uses_synchronous_normal_under_wal` asserts a real file-backed backend reports `journal_mode=wal` and `synchronous=1` (NORMAL).
- Full `helios-persistence` lib suite green (1027 passed), fmt + clippy clean.

## Follow-up

Before/after re-measurement of the commit phase with this change is being run on the 31 GB manifest host and will be posted to #947.

Closes #947


angela-helios:
Measured before/after on the 31 GB manifest host (NVMe, two independent pairs, N=100,091, <1% run-to-run noise):

| | res/s | WALL | commit ms/res |
|---|---|---|---|
| FULL (before) | 2,125 | 47.1s | 0.192 |
| NORMAL (after) | 2,236 | 44.8s | 0.171 |

**−4.5% wall, −10% on `commit`, +5% throughput.** Other phases flat.

Full tables and analysis in #947 (https://github.com/HeliosSoftware/hfs/issues/947#issuecomment-5674517675). Note from that run: on fast NVMe only ~0.02 ms of the ~0.19 ms commit is the fsync `NORMAL` removes — the rest is WAL page writes — so the win is larger on slow-fsync hosts (the Windows case in the issue, network disks). Safe, free, crash-safe under WAL, hence applied globally when WAL is on.

smunini (approved):

Issue #945, opened 2026-09-04
bulk-submit: performance roadmap back to ~30 min without configuration (multi-row INSERT, index write-volume audit)


angela-helios:
## Context

With #932 (search-index allowlist) closed by design decision, the out-of-the-box bulk import time regressed to the pre-campaign baseline. Measured today on the real 31 GB / 954,288-entry export, PostgreSQL 16, `HFS_BULK_SUBMIT_FILE_CONCURRENCY=8`:

| Configuration | Rate | Full import |
|---|---|---|
| PostgreSQL ×8, indexing inline (out of the box) | **95/s** | **~2h 45m** |
| PostgreSQL ×8, `HFS_BULK_SUBMIT_DEFER_INDEXING=true` | 133/s ingest | ~2h + deferred reindex |
| The former ~30 min figure | 527/s | required the #932 allowlist, now closed |

This issue tracks the remaining levers to get back to ~30 minutes **without configuration knobs**. Ordered by expected impact.

## 1. Multi-row INSERT / batched statements on the PostgreSQL write path

The strongest lever, confirmed live during today's import: sampling `pg_stat_activity` wait events while ingesting shows **11 of 19 samples in `Client:ClientRead`** — the server spends most of its time waiting for HFS to send the next statement. HFS is nearly idle too (1.8% CPU), the disk is idle (0.2%); the pipeline is round-trip-bound. Each resource issues ~25 individual index-row INSERTs, one round-trip each.

Collapsing the index rows for a resource (or a whole batch) into one multi-row `INSERT ... VALUES (...), (...)` — or `COPY` for the bulk path — turns those waits into work. This is Lever 4 / "Problem D" in the "Cutting Bulk Import to 30 Minutes" analysis: PostgreSQL currently needs 8 workers to compensate a per-worker handicap (~110/s single, round-trip-bound).

## 2. Audit the duplicate `_recent` sort-covering indexes (PostgreSQL)

Each populated index family maintains **two** partial indexes per row: the value index and a `_recent` variant covering `ORDER BY last_updated DESC`. Measured at 470k resources into today's import:

```
idx_search_token_code_recent  351 MB   (tenant, type, param, last_updated DESC, resource_id) INCLUDE (code, system)
idx_search_token_code         345 MB   (tenant, type, param, code, last_updated DESC, resource_id) INCLUDE (system)
idx_search_date_recent         91 MB   same pattern for dates
idx_search_date                67 MB
```

[...]

PR #1134, opened 2026-09-15
perf(sqlite): drop idx_search_token_display to cut ingest write volume (#945)

angela-helios:
## What

Drops `idx_search_token_display` from the SQLite `search_index` table — the first, safe write-volume win from the #945 index audit.

## Why

Index write-volume audit on a 105k-resource NVMe load (measured with the `--cfg perf_phases` instrumentation, on `main` after `synchronous=NORMAL`):

- `value_token_display` is populated on **~12.6% of `search_index` rows** (163k of 1.30M) and its index is **13.7 MiB**.
- Its **only reader** is the token `:text` / `:code-text` modifier, which is a `COLLATE NOCASE` partition scan — **not an index seek**. `:text-advanced` uses the FTS table, not this index.

So the b-tree paid an insertion on most Coding-with-display rows during every bulk ingest, for a read path that rarely runs and never seeks it. Removing it cuts **~5.6% of index-write entries per resource** with no change to query results.

## Change

- Removed `idx_search_token_display` from `SEARCH_VALUE_INDEXES`, so a fresh schema and the bulk-rebuild self-heal (`ensure_search_value_indexes`) no longer create it.
- Added migration **v28 → v29** (`SCHEMA_VERSION` 28 → 29) that `DROP INDEX IF EXISTS idx_search_token_display` on existing databases.
- `value_token_display` **column stays** and keeps feeding the FTS trigger.

## Behavior / tradeoff

- `:text` / `:code-text` on token params still return identical results, now served by a partition scan over the parameter's rows instead of the partial index (these were already NOCASE scans, not seeks). This is a deliberate read-cost-for-write-cost trade on an uncommon modifier path.
- No data migration of row contents; only an index is dropped.

## Tests

- `search_value_indexes_match_the_fresh_schema` and `upgraded_database_has_the_same_search_index_indexes_as_a_fresh_one` updated automatically hold (fresh DB creates the index at v1 then v29 drops it; upgraded DBs shed it too).
- Full `helios-persistence` lib suite green (1039 passed), fmt + clippy clean.

## Follow-up

The larger lever the audit surfaced — the 36-byte `resource_id` UUID carried on every `search_index` row and in `idx_search_composite` (~44% of index bytes, and per #947 a direct driver of both `commit` and `search_index_insert`) — is a separate, design-review-first change (integer surrogate key). Not in this PR.

Part of #945.


angela-helios:
## Measured A/B on the 105k NVMe load (`--cfg perf_phases`)

Fresh same-session pairs, main vs this branch, N=105,091, identical env/manifest (12.00 index rows/resource in every run).

| metric | pair 1 | pair 2 | mean |
|---|---|---|---|
| throughput | +1.9% | +1.1% | **+1.5%** |
| search_index_insert ms/res | −3.4% | −5.0% | **−4.2%** |
| commit ms/res | −3.8% | −1.1% | **−2.5%** |

Clean-run noise is ~1% (2,045 vs 2,023 res/s on main), so the win is real but ~1–2σ.

**Index bytes:** row-for-row (≈1.32M rows), index footprint drops **254.9 → 241.3 MiB = −13.6 MiB (−5.6%)** — exactly the dropped index's size. Schema version = 29, `idx_search_token_display` absent.

**Correctness:** `:text` returns identical results to the exact-code search, via a scan:

| query | HTTP | total | latency |
|---|---|---|---|
| `CarePlan?category:text=Respiratory therapy` (scan, no index) | 200 | 5,932 | 1.28 s |
| `CarePlan?category=http://snomed.info/sct\|53950000` (token index) | 200 | 5,932 | 0.45 s |

So `:text`/`:code-text` on token pay ~2.8× latency (scan vs the dropped partial index) for the write savings — the deliberate trade in this PR.

**Why the write win is modest:** the index covered only 12.6% of rows, so dropping it removes ~0.13 of the ~2.24 B-tree entries per row (−5.6%), which flows near-linearly into `search_index_insert` (−4%). `commit` barely moves because per-batch dirty-page volume is dominated by `idx_search_composite` (108 MiB, every row, carrying the 36-byte `resource_id` UUID). That composite/UUID is the lever that actually moves the write path — tracked separately as a design change.

smunini (approved):

PR #1158, opened 2026-09-16
perf(sqlite): resource_key surrogate for search_index, Stage 1 (#945)

angela-helios:
## What

**Stage 1** of the SQLite `resource_key` surrogate (#945, signed off in [this comment](https://github.com/HeliosSoftware/hfs/issues/945#issuecomment-5687120954)). Introduces an integer `resource_key` (= `resources.rowid`) on `search_index` and rekeys `idx_search_composite` from the 36-byte `resource_id` UUID to that 3–4 byte varint.

The composite index was **44% of the table's index bytes and the dominant per-batch write during bulk ingest** — a 36-byte UUID repeated on every row. Measured on hfs-50 at 1M resources: `idx_search_composite` **716 MB → 185 MB (−74%)**.

## Changes

- **Schema v30 migration** — adds `resource_key`, backfills it with a **single-pass `UPDATE … FROM resources`** (a JOIN, not a correlated subquery per row — the shape Steve flagged), and rekeys the composite index. The FTS triggers are captured from `sqlite_master`, dropped across the backfill, and restored verbatim.
- **Writer** emits `resource_key` (resolved in `write_prepared_index`, which holds the connection; the connection-free prepare step leaves a placeholder). Minimal-fallback inserters set it inline.
- **Delete-by-resource** (`delete_search_index`, soft-delete) and the **sort** correlated subquery route through `resource_key` so they keep using the composite index after the rekey. The purge path (which removes the `resources` row first) stays on `resource_id`.
- `resource_id` **stays on the table** — the chained-search / `_has` / `:identifier` paths concatenate `Type/resource_id` against `value_reference` and can't use the key.

## Why not a full table rebuild

`search_index_fts` is an FTS5 **external-content** index keyed on `search_index.rowid`. A `CREATE TABLE AS SELECT` rebuild renumbers rowids and would **silently** leave the FTS pointing at the wrong rows. This migration is in-place (rowids preserved) and only sets a new column, so FTS content stays valid — the triggers are dropped only to keep the backfill from re-running the full-text delete+reinsert on every row.

## Validation (hfs-50, 1M-resource NVMe base)

- **Migration strategy** measured: copy+JOIN vs `$reindex`-from-blobs — A is 9.5× faster ([comment](https://github.com/HeliosSoftware/hfs/issues/945#issuecomment-5690554885)); the in-place `UPDATE…FROM` used here avoids the FTS-rebuild pitfall above.
- **Read baseline** on `main` shows the pre-existing composite-served queries (SORT 25s, `:missing` 18s) are **not regressed** by this change — the sort swap *prevents* a catastrophic regression (without it the rekeyed index couldn't serve the correlation).
- **FTS baseline** captured; the `:text` queries that exercise the restored triggers (`category:text=Respiratory` → 5,932, `activity-code:text=Diabetic diet` → 4,764, `code:text=allergy` → 2,095) are the A/B correctness targets. hfs-50 is running the end-to-end A/B against this branch.

## Tests

Schema/ladder (22) — including replay — and the full `helios-persistence` lib suite (1042) green; fmt + clippy clean.

## Design note for review

The surrogate is `resources.rowid` (HFS never `VACUUM`s, so it's stable) rather than a declared `INTEGER PRIMARY KEY`, to avoid rebuilding the `resources` core table (evolved over 30 migrations) in this PR. If you want the VACUUM-safe declared key, that's a `resources` rebuild we can do as its own step.

## Follow-ups (separate PRs, not urgent per the read baseline)

1. Full `build()`/count read swap (base filter, per-param, composite GROUP BY, `:missing`, compartment, `_filter` + the outer `id IN` → `rowid IN`). Verified safe — chained search is resolved separately (`SearchQuery.chain`), so it stays on `resource_id`.
2. The composite-prefix extension you suggested (drop `tenant_id`/`resource_type` from the composite index once `resource_key` carries identity).

Part of #945.


angela-helios:
## End-to-end A/B validated (hfs-50, 1M-resource NVMe base, @560e67c6a)

The surrogate is a net **win on both write and read**, with identical results and FTS intact. (An earlier revision regressed filtered reads by dropping the composite's covering property; the `route the build() wrapper through resource_key` commit fixes it and the plans are back to a covering seek.)

### Write (ingest A/B, fresh pair, N=105,091)
| | main | surrogate | delta |
|---|---|---|---|
| res/s | 2,404 | 2,629 | **+9.4%** |
| search_index_insert ms/res | 0.146 | 0.135 | −7.5% |
| commit ms/res | 0.158 | 0.133 | **−16%** |

### Index size (dbstat, 1M)
`idx_search_composite` 716 → **317 MB (−56%)**; all `idx_search%` 1,611 → 1,211 MB (−25%). (−56% not −74% because Stage 1 keeps `tenant_id`/`resource_type` in the composite; only the 36-byte UUID left.)

### Read (median of 10, 1M base)
| query | main | surrogate+fix | vs main |
|---|---|---|---|
| SORT `Claim?_sort=created` | 25,520 ms | 5,019 ms | **5.1× better** |
| :missing `Claim?facility:missing=true` | 17,767 ms | 6,523 ms | **2.7× better** |
| chained `CarePlan?care-team.status=active` | 6,593 ms | 2,371 ms | **2.8× better** |
| token exact `AllergyIntolerance?code=…` | 5.6 ms | 4.2 ms | = |
| base `Claim?_count=50` | 1.8 ms | 1.6 ms | = |
| `Claim?patient=…` (sqlite, 21/907k) | 0.58 s | 0.51 s | = |

**EQP** back to `USING COVERING INDEX idx_search_composite (…resource_key=?)` for token and reference — same plan as main with `resource_key` in place of the UUID.

### Migration & correctness
- v29→v30 on the 1M base: **41.9 s**; `schema_version=30`, 0 `resource_key IS NULL`, `search_index_fts_docsize` unchanged (168,106).
- **Results identical to main** — full-set `diff` (no LIMIT) empty for token and `patient`; HTTP page ids/totals identical.
- **FTS intact** — `diff -r` of all 16 FTS query id-sets empty; the 3 `:text` totals match (5,932 / 4,764 / 2,095).

No chain follow-up needed — chained search flows through the same `build()` wrapper and is now faster than main.

smunini (approved):

Issue #1125, opened 2026-09-15
sqlite-es: the automatic rebuild after an import fails on oversized `_bulk` requests, leaves all 11,704 Provenance resources unsearchable (1 % cut), spends 1,551 s on its first attempt alone before retrying all 24 types, and writes a SQLite index nothing reads


eterceros:
*Every figure below is labelled **measured**,
**estimated** or **inferred**. Code references are `file:line` against `origin/main` at `05aeba427`
(includes #1093, #1081, #1097, #1113 and PR #1109, merged 2026-09-15 as `2be460f50`). Run M1e was
measured on #1109's head `2cfde9b94`, whose reindex and Elasticsearch code is identical to what merged
(only `Cargo.lock` and a test `#[allow(deprecated)]` differ), so M1e describes current `main` (inferred
from identical code; not re-run on `05aeba427`). Runs M1c (`main` before #1109) and M3b (prototype)
were measured on `7967a483e`, in the same environment; the database-size comparison was measured
earlier, with the same method, on `eae3d6fef` (includes #1093, predates #1097).
The prototype referred to lives on branch `exp/937-perf`, in the commits
`perf(reindex): write only to Elasticsearch when SQLite search is offloaded` and
`perf(reindex): pipeline the Elasticsearch reindex writes` (plus, for indexing during ingest,
`feat(bulk-submit): index into Elasticsearch during ingest`, tracked separately as a companion change that indexes resources into Elasticsearch as they are
committed during ingest instead of only after the manifest closes).*

## Why this matters

| | |
|---|---|
[...]

PR #1159, opened 2026-09-16
fix: bulk-submit ingest: a mid-body stream error completes the manifest with rows missing; the lease dies behind a WAL checkpoint and forces a full re-walk; Elasticsearch idles through the whole import

### LINKED ISSUE #1127: bulk-submit ingest: a mid-body stream error completes the manifest with rows missing; the lease dies behind a WAL checkpoint and forces a full re-walk; Elasticsearch idles through the whole import
## Why this matters

| | |
|---|---|
| **A "successful" import can be wrong** | When an input file's body breaks mid-stream, HFS abandons the rest of that file, writes one file-level `error` artifact, and still publishes the manifest as **`completed`** — with **no WARN in the log**. Measured: 4–8 files per run lose their last 20–130 KB. At corpus scale that is **thousands of clinical resources missing from a run the operator was told succeeded**. |
| **A real, but much smaller, time win** | Indexing into Elasticsearch **during** ingest instead of rebuilding after the manifest. #1156 (merged 2026-09-16, closed #1125) fixed the rebuild this comparison used to be built on: it now completes in **145 s, one generation, 0 errors, 11,704 / 11,704 Provenance indexed** (measured on #937 with `HFS_ELASTICSEARCH_WRITE_REFRESH=wait_for` and `HFS_ELASTICSEARCH_REINDEX_REFRESH=false`). So the honest comparison on the 1 % cut today is **ingest ~151 s + rebuild ~145 s ≈ 296 s** against the prototype's **225 s** with indexing folded into ingest — roughly **1.3×**, not the 4.4–6.9× this issue claimed while the rebuild was still failing. The structural win survives: the separate rebuild phase disappears altogether, and at 19M a serial rebuild may scale worse than ingest does (**hypothesis** — the full corpus has not been re-run since #1156). See §5. |
| **One slow secondary costs a full re-walk** | The real run lost its lease after ~7 h and re-read all 24 files: **+11 h**, 18 h 31 min total (measured, older code). Reproduced independently with a saturated Elasticsearch. With a non-terminal submission this can loop indefinitely (#998). |
| **The status report cannot be trusted** | Replay rewrites byte-identical resources as new versions (measured: DB **3.9 → 7.8 GB**, history ×2) and `total_entries` accumulates across passes (**37,911,730** reported for **18,955,865** receipts). The submission status poll is also O(N) over a table that reaches one row per ingested resource. |
| **Cost of not doing it** | Every corpus-scale `sqlite-es` import risks silent data loss, an 11 h penalty and a multi-hour rebuild, and produces numbers nobody can validate. |
| **Effort / risk** | Prototyped, compiled and tested on branch `exp/937-perf`, in the commits `feat(bulk-submit): index into Elasticsearch during ingest`, `fix(bulk-submit): keep the lease alive across WAL checkpoints`, `perf(bulk-submit): cut the per-entry ingest work` and `fix(bulk-submit): report the cause of a mid-body file stream failure` (rebased on #1093). Index-during-ingest is **opt-in** (default off); the lease fix keeps today's defaults. Only the fetcher `Range` retry is still to write. Medium-large effort, low risk. |

All line references are against `origin/main` at `05aeba427` (`git show origin/main:<path>`; includes #1093, #1081, #1097, #1113 and #1109, merged 2026-09-15 as `2be460f50`). Runs M1c (`main` before #1109) and M3b (prototype) were measured on `7967a483e` (1 % cut = 228,580 resources, release R4-only build, isolated Elasticsearch 8.15.0 with a 4 GB heap, corpus served by a real static server, idle machine, no stall cutoff); database sizes come from earlier same-method runs on `eae3d6fef`; run M1e was measured on PR #1109's head `2cfde9b94` in the same environment. **M1c and M1e are pre-#1156 history**: #1156 (merged 2026-09-16, closed #1125) rewrote the `_bulk` path, and the rebuild now completes in 145 s with 0 errors, so neither run describes current `main`'s rebuild any more (§5). Line references were re-checked against `origin/main` after #1156. Every figure is labelled **measured**, **estimated** or **inferred**.

---

## 1. Silent tail loss (the headline)

**What the operator sees.** The submission card reaches **Completed**. The manifest's `error` artifact contains one line per broken file, and the message it carries today is the useless `error decoding response body` — because `reqwest`'s `Display` drops the source chain. The log says nothing at `warn` or above. The counts look plausible, because nobody knows what the exact count should be. The database is simply missing the tail of 4–8 files.

**The actual error**, only visible via `{e:?}`:

```
hyper::Error(Body, Os { code: 10054, kind: ConnectionReset,
             message: "An existing connection was forcibly closed by the remote host." })
```

**Mechanism.**

- `crates/rest/src/bulk_submit_fetcher.rs:49` builds the client as `reqwest::Client::new()` — **no request timeout, no read timeout, no retry, no `Range` resume**.
- `:346-348` maps a body-stream failure straight into an `io::Error` using `Display`:
  `std::io::Error::other(format!("reading file {owned_url}: {e}"))`.
- The SQLite ingest engine turns that into a storage error at
  `crates/persistence/src/backends/sqlite/bulk_submit.rs:1393`:
  `internal_error(format!("Failed to read line: {}", e))`. Because `read_line` returned `Err`, `process_ndjson_stream` (`:1365`) returns immediately — **every remaining line of that file is abandoned**, while the batches already committed stay.
- The worker catches it at `crates/persistence/src/core/bulk_submit_worker.rs:1395-1411`: it writes **one** file-level `error` artifact (`failed to ingest file {url}: …`, `:1401`), bumps `failed` by 1 (`:1405`), adds `add_manifest_progress(lease_ref, 0, 1, 0)` (`:1408`) — and moves on to the next file.
- Nothing consults that counter when the manifest goes terminal: `:1593-1602` unconditionally publishes `ManifestPublicationStatus::Completed` (`fail_manifest` exists, `:371`, but is not reached from here).

**The shape is already pinned by a test — the outcome is not.** PR #1081 added `test_worker_reports_batches_committed_before_a_file_fails` (`bulk_submit_worker.rs:4240-4283`): a file whose stream breaks after two full batches reports both batches to the write observer and leaves **exactly 200 of its 250 Patients** in storage. It asserts that the committed batches survive a broken file; nothing asserts what the manifest's terminal status is in that situation, so "200 of 250, manifest `completed`" is known-shaped but unguarded.

**Measured.** Corpus served by `python -m http.server`, both the default HTTP/1.0 and `-p HTTP/1.1`: 4–8 files per run truncated, 20–130 KB each, manifest `completed` every time. Same corpus from a ~20-line Node static server: **0 losses**, and ~10× faster. The original 19M run did not hit it only because the debug binary read slowly enough.

**Precedent on MongoDB.** #1096 (merged 2026-09-13) added transient-retry and per-batch containment to the MongoDB bulk-submit path and left as follow-up "a 'completed with N unrecoverable entries' status signal". This is the same class of gap on SQLite, where there is no signal at all: the manifest ends `completed` with rows missing.

**Fix.** Three parts, in increasing order of work:

1. Diagnostic (already in the prototype commit `fix(bulk-submit): report the cause of a mid-body file stream failure`) — make the cause reachable at all:

```rust
let stream = resp.bytes_stream().map_err(move |e| {
    tracing::warn!(url = %owned_url, error = ?e, "bulk-submit file stream failed mid-body");
    std::io::Error::other(format!("reading file {owned_url}: {e:?}"))
});
```

   The URL must be redacted (query string stripped) before it goes anywhere — **in error messages, not only in the log**. On `main` a presigned input URL, signature included, already reaches the manifest's `error` artifact: the fetcher embeds `{url}` in its errors (`bulk_submit_fetcher.rs:313`, `:328`, and `owned_url` at `:348`), and the worker writes `failed to ingest file {url}: {e}` into the artifact (`bulk_submit_worker.rs:1401`). The prototype's `{e:?}` would carry the same URL into one more place.

2. Bounded retries with `Range: bytes=<consumed>-` on a body error, tracking bytes consumed per file, plus a request/read timeout on the client.
3. **A file that cannot be completed fails the manifest** (`fail_manifest`) instead of publishing `Completed`. A loud failure is strictly better than a quiet partial import.

---

## 2. Lease starvation → a second full pass over the corpus

**What the operator sees.** After ~7 h of a clean-looking ingest, one line:

```
WARN bulk-submit run abandoned while writing result receipts: its lease is no longer held
```

(`bulk_submit_worker.rs:1581-1590`) — and then the sibling worker starts again at file 1. **Measured (older code):** all 24 files re-read, **+11 h**, 18 h 31 min of ingest in total, `total_entries` 37,911,730 for 18,955,865 receipts, and the `bulk_submissions` row never reached a terminal state.

**Mechanism.**

- `checkpoint_after_file` (`sqlite/bulk_submit.rs:2226-2243`, called at every file boundary from `bulk_submit_worker.rs:1414-1416`) runs a bare `PRAGMA wal_checkpoint(TRUNCATE);` (`:2239`), introduced per file by #978/#1025. It runs inside `spawn_blocking` (`:2237`), so the checkpoint itself does not block a runtime thread; the problem is the lock it holds. `TRUNCATE` holds the WAL write lock for the whole copy; by that point in the run the WAL was multi-gigabyte, so the copy takes minutes — and every other writer, the heartbeat included, waits behind it.
- The heartbeat (`sqlite/bulk_submit.rs:1762-1796`) is wrapped in `retry_bookkeeping_on_busy("heartbeat", lease_retry_budget(lease), …)` (`:1770`; helper at `:124-152`), and that helper calls each attempt **synchronously** (`attempt()` at `:135`; only the back-off between attempts awaits, `:146`). Each attempt is a blocking rusqlite `UPDATE` executed directly on a runtime thread — this, not the checkpoint, is the piece that blocks the runtime. With `busy_timeout` at its 30 s default (`sqlite/backend.rs:206-215`), a single attempt parks that runtime thread for up to 30 s **with no await point**.
- The keeper guards it with `tokio::time::timeout(left, jobs.heartbeat(&lease))` (`bulk_submit_worker.rs:810`) — a timeout that **cannot fire while an attempt is blocked**, because the future it wraps is blocking rather than pending.
- When the `UPDATE` finally lands, `lease_expiry` has passed, a sibling worker has re-claimed the manifest, and the `status`/`worker_id`/`fencing_token` fence no longer matches: 0 rows affected → `lease_lost`. And that is swallowed (`bulk_submit_worker.rs:815-821`):

```rust
// Already reclaimed by another worker — expected, and the run aborts quietly.
Ok(Err(LeaseError::LeaseLost { .. })) => { let _ = flag.send(true); return; }
```

  (A `warn` does exist for the other exit — "bulk-submit lease could not be renewed before it expired", `:835-845` — but only when the loop runs out of time without an answer; a `LeaseLost` answer takes the silent path above.)

**It is not only the WAL.** Reproduced independently with a saturated Elasticsearch (the 1,000-shard wall): ES rejected writes → the **unbounded** secondary `flush` waited **40 min** → the heartbeat could not get the SQLite write lock for **4.5 min** (`database is locked`, 8 retries logged) → the lease expired (all measured, older code). **Any slow secondary can cost a full re-walk.**

**Fix** (prototype commit `fix(bulk-submit): keep the lease alive across WAL checkpoints`, 18 tests): `PASSIVE` checkpoint first, then `TRUNCATE` under a 1 s `busy_timeout`, logging WAL bytes/frames and duration at `info` (`warn` past 5 s); heartbeat the lease immediately **before** the checkpoint so the copy starts with a full lease in hand; run each heartbeat attempt on a blocking thread (`retry_bookkeeping_on_busy_async`) so the keeper's timeout works; and `warn!` on `LeaseLost`, heartbeat timeout and reclaim — a lost lease must never be silent.

---

## 3. Replaying a manifest is not idempotent

Nothing compares content, so re-submitting the same manifest rewrites byte-identical resources as new versions (the update path, `sqlite/bulk_submit.rs:1283-1302`). **Measured** (same database, same manifest, twice, 1 % cut, older code): **211 s → 321 s**, `fhir.db` **3.9 → 7.8 GB**, `resource_history` doubled to **458,539** rows. Combined with §2 this is what made the 19M run cost 18 h 31 min for ~7 h of real work.

The counters compound it: `sqlite/bulk_submit.rs:938` does `total_entries = total_entries + ?1`, so every pass adds the whole file again — hence 37,911,730 for 18,955,865 receipts, which is what made the status report unreadable.

**Contrast with #1081.** Its new dashboard counters already handle a re-walk correctly: a re-ingested existing resource is reported with `created = false` (`sqlite/bulk_submit.rs:1296-1301`), `BatchCountsReporter` tallies it as an update (`bulk_submit_worker.rs:559-584`), and `DashboardCountsObserver` records only `+created`/`-deleted` — "updates change no count" (`crates/rest/src/handlers/dashboard_counts.rs:13-15`, `:60-81`). So after a replay the Home dashboard shows the right resource count while the submission's `total_entries` shows double. Two views of the same run now disagree.

#1096 (merged 2026-09-13) made replay idempotent on MongoDB (content-identical resources are skipped); on SQLite it is not, as measured above. SQLite and Postgres want the same treatment (opt-in), plus a `total_entries` that reports the manifest's entries rather than the sum over passes.

---

## 4. `HFS_BULK_SUBMIT_BATCH_SIZE` is a dead knob

It is parsed at `crates/rest/src/config.rs:844` (`batch_size: env_u32("HFS_BULK_SUBMIT_BATCH_SIZE", d.batch_size)`, documented default 1000) and then **never reaches the ingest**: the worker builds its options at `core/bulk_submit_worker.rs:1144-1148` as `BulkProcessingOptions::new().with_import_mode(…).with_defer_indexing(…).with_byte_progress(…).with_cancel(…)` (plus #1081's `with_batch_observer`, `:1151-1156`) — no batch size — so every run uses `default_submit_batch_size() = 100` (`core/bulk_submit.rs:964-966`). The setter exists — `BulkProcessingOptions::with_batch_size` (`core/bulk_submit.rs:1064-1067`) — but the server worker never calls it; outside tests only the benchmark does (`crates/persistence/examples/bulk_submit_bench.rs:308`).

This also qualifies #947, which describes its measurements as "batch 1,000": through the server worker the real batch is 100 unless the worker was patched locally (inferred).

Note the measurement before "fixing" it by raising the default: with the ingest sink enabled, batch 1000 was **worse** — **382 s vs 252 s** for batch 100 (measured, 1 % cut, older code). **The fix is to honour the knob, not to change the default.**

---

## 5. Index into Elasticsearch during ingest instead of rebuilding afterwards (a smaller win since #1156)

**Current design.** With `HFS_BULK_SUBMIT_DEFER_INDEXING=true` (default since #946) on `sqlite-es`, `composite_submit_jobs` (`crates/hfs/src/main.rs:1851-1864`, called at `:2295-2300`) picks the **raw primary** as the job store:

```rust
if defer_indexing && has_reindex_hook { primary } else { /* CompositeSubmitJobs wrapper */ }
```

(`:1857`), so the composite wrapper is bypassed and `sync_ingested` falls back to its default no-op (`core/bulk_submit_worker.rs:381-384`, `Ok(IngestSyncReport::default())`). **Elasticsearch receives nothing at all during ingest**; it is filled only by the deferred reindex hook fired after the manifest is terminal (`bulk_submit_worker.rs:1604-1610`).

**The rebuild this section used to measure against is now fixed (#1156).** Pre-#1156 history: on `main` before #1109 (`7967a483e`, 1 % cut) that rebuild **failed twice** — 403 s, then an automatic retry of all 24 types for 441 s, both ending with `2500 resource errors` — and gave up **995 s** after the import started, logging `deferred reindex failed twice; run $reindex manually`, with **226,080 of 228,580** documents in Elasticsearch: 2,500 Provenance resources stayed unindexed (measured, M1c). On #1109's head `2cfde9b94` the higher `nested_objects.limit` made each Provenance document carry far more nested objects, and **0 of 11,704** Provenance were indexed in a 1,551 s generation, every error reading `backend unavailable: elasticsearch`, `retryable: true`, with the whole 24-type rebuild repeating (measured, M1e). The cause was a defect in the Elasticsearch bulk-indexing path: `_bulk` was chunked by operation count with no byte cap, so a 500-operation Provenance request exceeded the client's 30 s request timeout and was failed chunk-wide. **#1156 (merged 2026-09-16, closed #1125) fixed exactly that**: byte-capped `_bulk` chunking (`chunk_ranges`, 10 MiB default via `HFS_ELASTICSEARCH_BULK_MAX_BYTES`), recursive half-split of a chunk on timeout/413/408/504 down to a single document, and 429 exponential backoff, all inside `send_bulk_index` (`crates/persistence/src/backends/elasticsearch/storage.rs`). The rebuild now completes in **145 s, one generation, 0 errors, 11,704 / 11,704 Provenance** (228,590 / 228,590 root documents) with `HFS_ELASTICSEARCH_WRITE_REFRESH=wait_for` + `HFS_ELASTICSEARCH_REINDEX_REFRESH=false` — measured on the same 1 % cut and reported on #937, 2026-09-16.

**What that leaves of this issue — the headline claim is withdrawn.** The comparison this section used to make (995 s of a failing rebuild against 225 s of the prototype, "**4.4×**", and "**≥ 6.9×**" against M1e's failed generation) no longer describes anything on `main`, because the rebuild it counted is no longer broken or slow. Against a fixed rebuild the arithmetic on the 1 % cut is **~151 s ingest + ~145 s rebuild ≈ 296 s** versus **225 s** for the prototype with indexing folded in: about **1.3×**. Two things still argue for the design. First, it removes the rebuild phase rather than making it fast — no second serial pass over everything just written, no window in which the manifest reads `completed` while search is still incomplete, no `$reindex` to babysit. Second (**hypothesis, not measured**): that second pass is a full re-read and re-index of the whole corpus, so at 19M resources it may scale worse than ingest does even after #1156 — 145 s at 228k says little about 19M, and nobody has re-run the full corpus since #1156. If that holds, the benefit at corpus scale is larger than the 1 % cut's 1.3× suggests; if it does not, this is a modest optimisation and should be scheduled as one. History, older code and pre-#1156: the real 19M run rebuilt at 93 res/s, ~53 h projected (measured rate, estimated projection) — that projection is now obsolete.

**Proposed design** (prototype commit `feat(bulk-submit): index into Elasticsearch during ingest`, 7 tests, opt-in): the engine calls `batch_committed` **right after** a batch transaction commits — never before, so the sink only sees what the primary durably holds — handing the batch to N writer tasks behind **bounded** queues, FIFO per resource by `hash(type, id)` so a late duplicate cannot overwrite a newer version. The worker drains the sink **before writing receipts** (#1007), so a receipt never reads `success` for a resource search cannot find; what the sink rejects is marked `unindexed` and only those types get the deferred reindex. (#1081 has since added an upstream post-commit hook at the same point — `notify_batch_committed`, `sqlite/bulk_submit.rs:978` — which the sink can share.)

**The bulk-writing half of this design no longer has to be built.** The prototype shipped its own 10 MiB byte chunker because `main` had none; `origin/main` now has that logic in `send_bulk_index` (`elasticsearch/storage.rs`, from #1156) — byte-capped chunking, half-split of a request on timeout/413/408/504, 429 backoff, transient reclassification. An index-during-ingest sink should hand it pages and inherit all of it rather than reimplement any of it, which removes the riskiest part of the original proposal.

**Measured** (1 % cut). The first two rows are **pre-#1156 history**, on `7967a483e` (`main` before #1109); the third is current `main` as measured and reported on #937 after #1156:

| Run | Ingest | Rebuild | Until search is final | ES root docs at end |
|---|---|---|---|---|
| M1c `main` before #1109 (pre-#1156) | 151 s | 403 s + 441 s, both failed | **995 s**, then gives up | 226,080 / 228,580 |
| M3b `main` before #1109 + prototype | 225 s, indexing included | none needed | **225 s** | **228,580 / 228,580** |
| current `main`, after #1156 (2026-09-16) | ~151 s | **145 s, 1 generation, 0 errors** | **~296 s** | **228,590 / 228,590** |

M3b's ingest is slower than M1c's because it includes the indexing; Elasticsearch is complete when those 225 s end, with no warnings or errors — which is still the best number on the table, by roughly 1.3×. Database size was part of the old argument too (measured earlier on `eae3d6fef`: 6,517 MB on `main` vs 3,941 MB with the prototype); that gap has largely closed, because #1156 also stops `sqlite-es` writing the dead SQLite `search_index` (890,350 rows → 0) and takes the same cut from 4,739 MB to 3,772 MB on `main` (measured, #1125) — the prototype's range. Opt-in via `HFS_BULK_SUBMIT_INDEX_DURING_INGEST` (default **off**), shaped by `HFS_BULK_SUBMIT_INDEX_QUEUE` (16), `_CONCURRENCY` (4–8), `_COALESCE` (4). History, older code: end-to-end **37 min → ~4 min** (measured).

**Critical detail learned by measurement (older code):** the flush must be **bounded** and the sink pages **small**. Raising coalescing from 4 to 16 took ingest **245 → 329 s (+34 %)**, and a queue of 32 on top of that took it to **548 s (+65 % more)** — and an **unbounded** flush is precisely what starved the lease in §2. A slow secondary must degrade to `unindexed`, never block the writer.

---

## 6. Cheap reporting fixes

The submission status poll is O(N) in ingested resources. `crates/rest/src/handlers/bulk_submit.rs:879` calls `jobs.get_submission`, whose SQLite implementation (`sqlite/bulk_submit.rs:219`, query at `:276-288`) runs, **per poll**:

```sql
SELECT COUNT(*), SUM(CASE WHEN outcome = 'success' …), … FROM bulk_entry_results
 WHERE tenant_id = ?1 AND submitter = ?2 AND submission_id = ?3
```

`bulk_entry_results` grows to one row per ingested resource — 19M by the end — so the poll gets steadily more expensive *and* competes with the ingest writer for the same SQLite file; in the real run the UI's 5 s refresh visibly stretched to minutes (measured, older code).

**What #1081 changed, and what it did not.** The **Home dashboard** is now memory-served: page loads never query storage, and its job counts come from a background refresh (`crates/rest/src/dashboard.rs:1325-1328`, `count_active_submissions` at `:1365`). That caveat is gone. The **submission status poll** is unchanged — it is still the query above, and the UI's bulk-import tracking still drives it through `$bulk-submit-status` (`crates/ui/src/bulk_import.rs:874-878`).

The manifest row already carries the counters: serve the summary from `bulk_manifests` and keep the scan for an explicit detail call. Alongside it: stop accumulating `total_entries` (§3) and close the `bulk_submissions` row (#998).

---

## 7. Context that bounds the ambition — do not re-do this work

On older code the ingest loop was a **single serial tokio task** at **~2,300–2,700 resources/s** (measured). A full producer/consumer rewrite (batched existence probes, moved instead of cloned JSON, multi-row bookkeeping, 1 MB read buffer, commit in `spawn_blocking`; kept on `exp/937-perf` as `perf(bulk-submit): optional producer/consumer ingest pipeline`) was implemented and A/B-measured on the same binary: **245 s reference → 248 s serial, 267 s pipelined, 235 s pipelined with 8 writers** — **±8 %** (measured, older code). #1093 has since parallelised index preparation and brought the fast-load ingest of the 1 % cut to **151–157 s** (measured: 157 s on `eae3d6fef`, 151 s in M1c on `7967a483e`). An independent measurement agrees: with the phase timers merged in #1113, angela-helios reported on #947 (Linux, inline indexing, 133k resources) **commit at 45 %**, `search_index_insert` at 26.5 %, bookkeeping at 9.3 % (`entry_results` 8.1 %) and clones at 2.3 % of ingest time — so bookkeeping and clones, the per-entry work a rewrite can remove, bound that gain to roughly 12 % (the timers still need a `--cfg perf_phases` build plus `HFS_PERF_PHASES=1`). The commit share is what #1114 (open) addresses by making `synchronous=NORMAL` the default under WAL, with no environment variable; the prototype's `HFS_SQLITE_SYNCHRONOUS` knob used in the history table below is prototype-only and superseded by #1114. Once indexing happens during ingest, **the secondary write path, not the ingest loop, is the ceiling**. The remaining lever is the Elasticsearch page writer (FHIRPath extraction and repeated serialization per page); that belongs to a separate defect in the Elasticsearch write path (filed separately), with its own measured PR.

---

## Measured results (1 % cut = 228,580 resources, release binary, isolated ES)

**Pre-#1156 history** — `main` before #1109 (`7967a483e`, R4-only release build, no stall cutoff), plus M1e on #1109's head `2cfde9b94`. **Neither M1c nor M1e describes current `main`**: since #1156 the rebuild completes in 145 s with 0 errors (last row, §5).

| Configuration | Ingest | Search ready | Total |
|---|---|---|---|
| M1c `main` before #1109, pre-#1156 (deferred rebuild) | 151 s | +403 s failed rebuild, +441 s failed retry; gives up at **226,080 / 228,580** docs | **995 s** to give up, search still incomplete |
| M1e #1109 head, pre-#1156 (deferred rebuild) | — | +1,551 s failed generation 0, Provenance **0 / 11,704**; full 24-type retry (stopped) | never (run stopped) |
| M3b prototype on `7967a483e` (index during ingest) | 225 s | **no rebuild**, 228,580 / 228,580 docs | **225 s** |
| current `main` after #1156 (deferred rebuild), measured on #937 2026-09-16 | ~151 s | +145 s rebuild, 1 generation, 0 errors, **228,590 / 228,590** docs | **~296 s** |

`HFS_BULK_SUBMIT_BULK_INDEX_REBUILD=true` is not listed: per the code (`begin/end_bulk_index_rebuild` return early on `is_search_offloaded()`), it cannot activate on `sqlite-es`.

**History, older code (before #1093; 4 GB ES heap):**

| Configuration | Ingest | Search ready | Total |
|---|---|---|---|
| Baseline (real-run env: `wait_for`, `synchronous`) | 235 s | +2,005 s rebuild (114 res/s) | **~37 min** |
| Index during ingest + `HFS_SQLITE_SYNCHRONOUS=normal` (prototype-only knob; superseded by #1114) | **211–245 s** | **0 s** | **~4 min** |
| Index during ingest, batch 1000 | 382 s | 0 s | the knob would be a regression |
| Index during ingest, coalesce 16 / + queue 32 | 329 s / 548 s | 0 s | unbounded flush starves the lease |
| Manifest replayed on the same database | 321 s | — | DB 3.9 → 7.8 GB, history ×2 |

19M projection: **~2.3 h** (estimated on older code: ~2.3k res/s ingest plus the fixed ~2.5 min Provenance cost; not re-derived from M3b), against **>70 h** measured for the original run.

## Acceptance criteria

- [ ] A file truncated mid-stream is retried with `Range`; if it still cannot be completed, the manifest goes **`failed`**, never `completed`, and the stream error is logged at `warn` with its source chain.
- [ ] Regression test: an input server that closes the body mid-file yields a non-`completed` manifest and a receipt count that is either complete or explicitly short (extending `test_worker_reports_batches_committed_before_a_file_fails` to assert the terminal status).
- [ ] Under an adversarial lease (10 s lease, 2 workers, forced multi-GB WAL) the manifest is never reclaimed; any `LeaseLost`, heartbeat timeout or reclaim appears at `warn`; checkpoint duration and WAL frames are logged.
- [ ] A secondary that stops accepting writes degrades the batch to `unindexed` within a bounded time and does **not** expire the lease.
- [ ] `HFS_BULK_SUBMIT_BATCH_SIZE=N` changes the observed batch size; the default stays 100.
- [ ] Replaying a manifest leaves database size and `resource_history` count ~unchanged, and `total_entries` equals the manifest's entry count, not a multiple of it (consistent with the dashboard counters).
- [ ] With `HFS_BULK_SUBMIT_INDEX_DURING_INGEST=true`, the Elasticsearch document count equals the receipt count at the instant ingest ends, and no deferred rebuild runs unless something was rejected.
- [ ] The submission status poll's cost does not grow with the number of ingested resources.
- [ ] The `bulk_submissions` row reaches a terminal state (#998).

Related: #969, #998, #1001, #1096, #1007, #947 (and #1113, its phase timers: independent per-phase breakdown, §7), #1114 (open: `synchronous=NORMAL` by default under WAL, supersedes the prototype's `HFS_SQLITE_SYNCHRONOUS`), #978 / #1025 (origin of the per-file `wal_checkpoint(TRUNCATE)`, §2), #1081, #1093, #1050 (closed), #1109 (merged `2be460f50`), #1156 (merged 2026-09-16, closed #1125 — fixes the deferred rebuild §5 used to measure against, and provides the `send_bulk_index` chunking an ingest-time sink should reuse), #1161 (open: rolling back or aborting a deferred bulk submit leaves orphaned documents in Elasticsearch — the same family as §1 and §6, a submission whose reported outcome does not match what storage holds), #937 (the manual test pass this came from).

eterceros:
Closes #1127

Three independent defects made a bulk-submit import lose data, repeat work, and misreport its own outcome. They are fixed together because they were found in the same 19M-resource import and they compound each other.

## What was wrong

**1. A broken body stream published a completed manifest with rows missing.** The fetcher had no connect or read timeout, no retries and no resume. When a server hung up mid-body, the tail of the file was silently dropped and the worker published the manifest as `Completed` anyway. Against `python -m http.server`, 4-8 files per run lost 20-130 KB each, every run reporting success.

Even when the failure did reach the database, it did not reach the operator: one broken file produced **two** OperationOutcome artifacts — one inventing a parse error that never happened, one disguising the cause as `internal error in sqlite` and leaking the URL signature — while the UI list still showed `Completed / 0 error files`.

**2. The lease died behind a WAL checkpoint and forced a full re-walk.** `checkpoint_after_file` ran a bare `PRAGMA wal_checkpoint(TRUNCATE)`, which holds the write lock for the whole copy; on a multi-gigabyte WAL the heartbeat queued behind it for minutes. Heartbeat attempts also blocked a runtime thread with no await point, so the keeper's timeout could not fire, and `LeaseLost` was swallowed in four places. A sibling worker then reclaimed the manifest and **started over at file 1**: on the real run, 24 files re-read, **+11 hours** (18h31m total), and `total_entries` reporting 37,911,730 for 18,955,865 receipts.

**3. Elasticsearch idled through the whole import.** With deferred indexing on `sqlite-es`, the worker received the raw primary and `sync_ingested` fell through to its no-op, so ES received nothing during ingest and was filled only by the post-manifest rebuild — which itself failed wholesale on an unbounded `_bulk` body, so one page of large `Provenance` resources blew up the request as `backend unavailable`.

That second half is no longer part of this PR: #1125 fixed the unbounded `_bulk` first and more thoroughly, and this branch now takes `elasticsearch/storage.rs` from main unchanged. What remains here is the idling itself.

## What changed

**Fetcher** (`crates/rest/src/bulk_submit_fetcher.rs`): connect and read timeouts, `no_gzip()` so the wire offset is known, resume via `Range`/`If-Range`, 3 attempts with 1/2/4s backoff, handling for 416, short 206, ignored `Range`, and a resent file shorter than the offset. Errors keep their cause chain through the typed `BulkSubmitError::InputStream`, and input URLs are redacted in every operator-visible message.

**Manifest and lease** (`crates/persistence/src/core/bulk_submit_worker.rs`, `backends/sqlite/bulk_submit.rs`): a file that cannot be read in full fails the manifest instead of completing it, with a message naming the first failing file, the real cause and a redacted URL, and exactly one error artifact. Checkpoints run PASSIVE then a bounded TRUNCATE with frames/bytes/duration logged; the lease is renewed immediately before each checkpoint; heartbeat attempts move off the runtime thread so the keeper's timeout can fire; `LeaseLost`, heartbeat exhaustion and rival claims are logged at `warn` instead of returning silently.

**Replay accounting**: a per-file watermark table `bulk_manifest_file_progress` (SQLite migration v30, mirrored in PostgreSQL and MongoDB) means a replayed file is counted once. Note this prevents double counting, not re-ingestion — skipping unchanged work is still gated on `HFS_BULK_SUBMIT_SKIP_UNCHANGED`. Submission status is now served from manifest counters rather than aggregating `bulk_entry_results`, so polling cost no longer grows with the import.

**Elasticsearch**: a new opt-in sink indexes committed batches during ingest through bounded per-writer queues, FIFO per resource, degrading a stalled target to `unindexed` so the deferred rebuild covers only the rejected types, and draining before receipts are written. The `_bulk` byte ceiling this branch originally carried is not here any more — #1125 landed a configurable one (`HFS_ELASTICSEARCH_BULK_MAX_BYTES`, same 10 MiB default) with retry and back-off on top, and this branch adopts it.

**Other**: `HFS_BULK_SUBMIT_BATCH_SIZE` is now honoured. The UI accepts `countSeverity` as an array or an object — HFS itself emits the array form, which previously read as "no errors" — and the submission list's Status column is covered by a test.

## Defaults, migration and operator notes

- `HFS_BULK_SUBMIT_INDEX_DURING_INGEST` defaults to **off**. Nothing about indexing behaviour changes unless an operator turns it on.
- SQLite **migration v30** adds `bulk_manifest_file_progress`.
- The sink's tuning knobs should not be raised blindly: measured on the prototype, coalesce 4 to 16 went 245s to 329s (+34%), adding queue 32 went to 548s, and `BATCH_SIZE` 100 to 1000 went 252s to 382s. That is why the fix honours the knob rather than raising the default.

## Verification

`cargo test --workspace --all-features --no-fail-fast`, on top of main as of 755b33bc: 300 suites, 8390 tests, all green. New adversarial coverage includes a pinned WAL checkpoint that no `TRUNCATE` can reset (the manifest must be claimed exactly once, with no lease lost), a stalled search target that must degrade to `unindexed` without losing the lease, and an Elasticsearch integration test asserting that indexed documents equal receipts when the worker returns, with no deferred rebuild. A scripted browser QA run covering resume, permanent failure and a 90s stall is summarised in the first comment below.


eterceros:
## Verification at 80b69cf1

### Browser QA against this code (SQLite backend)

A scripted run drove the three stream-failure shapes against a server that cuts bodies on demand. Numbers are from that run, not from the investigation on the earlier prototype.

| Scenario | Result |
|---|---|
| Body cut once at 21,023 of 35,034 bytes | Resumed with `Range` + `If-Range` to a 206, **250/250 Patients** ingested, manifest `Completed` |
| Body cut on every attempt | 1 GET + **3 retries**, then manifest **`Failed`** with 1 error file; the **100** Patients from batches that had already committed are kept |
| Body stalls for 90s | 60s read timeout fires, resumes to a 206, **250/250** ingested |

The error artifact is now a single OperationOutcome naming the real cause, with the URL redacted:

```
failed to ingest file http://127.0.0.1:18530/patients.ndjson?[redacted]: reading file ...:
error reading a body from connection: end of file before message length reached
(gave up after 21623 bytes and 3 retries)
```

Before this change the same file produced two artifacts: one claiming `1 submitted resource(s) could not be parsed or did not match the declared resource type` (no parse error ever occurred) and one reporting `internal error in sqlite` with `sig=...` still in the URL.

Resource totals reconcile exactly across the run (1220 + 10 + 250 + 100 + 250 = 1830) and stay flat after a replay. The QA suite went from 2 failures to **15/15 passing**; the two that had been red were the ones asserting operator-visible state.

Checkpoint instrumentation over the same run: 5 checkpoints, `wal_frames` 114-1413, `wal_bytes` 466,944-5,787,648, `duration_ms` 4/89/75/8/8, `truncate_busy=false`, and no lease warning anywhere. That is a small-scale sanity check, not a reproduction of the multi-gigabyte WAL case — that case is covered by the pinned-checkpoint test instead.

### Measured impact and where each number comes from

Numbers below are from the investigation in #1127, measured on the `exp/937-perf` prototype, **not re-measured on this branch**:

- Index during ingest, 1% corpus slice (228,580 resources), isolated ES 8.15.0: **995s and gives up at 226,080/228,580 documents** to **225s complete at 228,580/228,580** — 4.4x, plus 2,500 `Provenance` documents that previously never became searchable. Ingest itself is ~49% slower (151s to 225s); the saving is the rebuild, which was 844s of the 995s.
- Lease survival on the real 19M run: 24 files re-read to 0, **+11 hours removed**, `total_entries` 37,911,730 corrected to 18,955,865.

What this branch adds is the guarantee rather than a new measurement: the ES integration test asserts documents equal receipts when the worker returns with no deferred rebuild, and the concurrency tests assert exactly-once claim under a WAL that cannot be checkpointed. Since `HFS_BULK_SUBMIT_INDEX_DURING_INGEST` defaults to off, the 4.4x is available to operators who opt in, not a default change.

If a hard number on this branch would help the review, I can run the ES benchmark on the merged code and post it here.

### Unrelated finding

A local `--workspace --all-features` run surfaced R6 corpus drift: `crates/fhir/build.rs` re-downloads the examples from `build.fhir.org`, which now serves `6.0.0-snapshot1` where `NutritionOrder.enteralFormula.type` became an array, while the committed model is still `6.0.0-ballot4`. It has nothing to do with this PR and is filed as #1163. CI is unaffected because the jobs running that test use `--workspace --all-features`, which activates `helios-fhir/skip-r6-download`.

eterceros:
## Merged main at 755b33bc

Forty-seven commits, five conflicts. Two of them changed what this PR contains, so they are worth a reviewer's attention rather than being buried in the merge commit.

### Elasticsearch `_bulk`: superseded by #1125

This branch carried a byte ceiling for `_bulk` bodies. #1125 fixed the same defect first and went further — a configurable ceiling (`HFS_ELASTICSEARCH_BULK_MAX_BYTES`, defaulting to the same 10 MiB this branch hard-coded), binary halving on 413, back-off on 429, a preserialized body reused across retries, and a wiremock suite. `elasticsearch/storage.rs` is therefore taken from main unchanged, and `plan_bulk_chunks`, `BULK_BYTES_PER_REQUEST` and `bulk_index_op_bytes` are gone. No behaviour is lost: `chunk_ranges` covers the same cases, oversized single document included. The PR body has been corrected accordingly.

### SQLite migration renumbered to v32

main took v30 for `index_pending` (#1156) and v31 for the `resource_key` surrogate (#1158), so `bulk_manifest_file_progress` and `bulk_manifests.skipped_entries` move to v31→v32 at the end of the ladder. Neither of main's migrations touches what this one adds. `drop_all_tables` now drops the new table, which it never did.

### Two of main's tests pinned accounting this PR moved

Both are updated rather than the behaviour, because in both cases the old expectation is now the inconsistent one.

`postgres_bulk_submit_batched_bookkeeping_preserves_max_error_semantics` expected a manifest left at 0/0/0 after a max-errors abort. Since this PR charges the counters inside the batch transaction, the aborting entry — whose receipt is committed — is now counted: 1/0/1. Three things say that is the right value:

- main's own test asserts exactly this in its continue half (`failed_entries == 1`); the same entry was counted in one half and not the other, purely because of where the statement sat.
- SQLite already reported 1 before this branch, and MongoDB does since #1127. Postgres was the outlier.
- With the summary now read from the counters, a submission aborted on errors would report `total_entries: 0, error_count: 0` while holding a durable `processing-error` receipt.

`postgres_bulk_submit_status_uses_the_complete_identity_and_preserves_summary` seeded four receipt rows and expected the summary to aggregate them. It now seeds the manifest counters as well. The receipts stay on purpose: they are the pagination surface, and leaving them shows the summary no longer reads them. The complete-identity half of that test, from #1154, is untouched and passes — `get_submission_status` is unchanged by this merge.

### Known outlier, not fixed here

`s3/bulk_submit.rs` still charges its counters after the max-errors return, so it reports 0 in that scenario and is now the only backend that does. Out of scope for this PR; worth its own issue.

### Verification

`cargo test --workspace --all-features --no-fail-fast` with Docker-backed suites (MinIO, Elasticsearch, PostgreSQL): **300 suites, 8390 tests, no failures**.

PR #1165, opened 2026-09-16
perf(sqlite): reorder idx_search_composite (param_name before resource_key) (#945)

angela-helios:
## What

Reorders `idx_search_composite` — same columns, **same size**, a pure column-order change (v32 migration):

`(tenant_id, resource_type, resource_key, param_name, composite_group)`
→ `(tenant_id, resource_type, param_name, resource_key, composite_group)`

Follows the merged `resource_key` surrogate (#1158).

## Why

A candidate sweep on hfs-50 (1M-row corpus, `ANALYZE` + EXPLAIN QUERY PLAN + timing of the exact server SQL) compared four column orders. This one **strictly dominates** the shipped order:

| query | shipped (v31) | this (v32) | Δ |
|---|---|---|---|
| `:missing` (`Claim?facility:missing=true`) | 6.48 s | **5.25 s** | −19% |
| sort (`Claim?_sort=created`) | 5.00 s | **3.73 s** | −25% |
| `Claim?patient=…` | 0.53 s | 0.53 s | = |
| index size | 302.2 MB | 302.4 MB | = |

Leading with `param_name` turns the `:missing` inner `(tenant, type, param_name)` subquery into a **covering seek** (467k entries) instead of a full-type scan (7.7M), and lets the sort's correlated `resource_key = rowid AND param_name = …` probe land in a `param_name`-contiguous index range. Nothing measured slower.

## Rejected alternative

Dropping `tenant_id`/`resource_type` from the composite for a smaller index (−41%) was measured and **rejected**: it regresses `:missing` (+21%) and drops any `(tenant, type, param_name)` filtered subquery that doesn't fit a partial family index to a **table scan** — a structural multi-tenancy risk, not just one slow query.

## Tests

Schema/ladder (22) — including replay and fresh-vs-upgraded — and full `helios-persistence` lib suite (1066) green; fmt + clippy clean.

## Follow-ups (measured, out of scope here)

hfs-50 flagged two query-builder items that dwarf any index order: the `:missing` outer `DISTINCT` wrapper scans the whole type (~5 s of the 5.25 s), and reference search scans `resources` rather than inverting the join to probe by `rowid` (0.5 s for 21 results). Both are query-builder changes for separate PRs.

Part of #945.


angela-helios:
Closing in favor of #1211.

That PR fixes a delete-by-resource full-scan — the root cause of #1197: `DELETE FROM search_index WHERE resource_key = (…)` without the `(tenant_id, resource_type)` prefix cannot use `idx_search_composite` and full-scans the table, which is O(search_index rows) on every resource UPDATE and every re-indexed resource (17 resources/s at 6M). The fix keeps `resource_key` immediately after `(tenant_id, resource_type)` in the composite so the delete seeks it.

This PR reorders the composite to put `param_name` ahead of `resource_key`. That re-degrades the delete to a `(tenant, type)` prefix scan — a direct conflict, and no single column order serves both the delete (wants `resource_key` next) and `:missing`/sort (want `param_name` next). The delete is on the write path and O(n) per UPDATE, so it takes priority; the `:missing`/sort gains here are secondary and already dwarfed by the surrogate itself. If `:missing` needs more later, it should get its own index rather than this reorder.

Issue #1055, opened 2026-09-11
search(mongodb): `_id:not` returns exactly the excluded resource, and `_id:missing` matches nothing


dougc95:
## Summary

On the `mongodb` backend, `_id:not=<id>` returns **exactly the resource that was excluded** — the filter is built as the positive match. `_id:missing=false` matches nothing where it should match everything. This is silent wrong data, not an error.

## Current behavior

`matching_resource_ids` skips `_id` and `_lastUpdated` **before** the modifier dispatch (`crates/persistence/src/backends/mongodb/search_impl.rs:947`):

```rust
for param in &query.parameters {
    if matches!(param.name.as_str(), "_id" | "_lastUpdated") {
        continue;                       // <-- before any modifier handling
    }
```

They are handled instead by `build_resource_id_condition` (`search_impl.rs:1487`) and `build_resource_last_updated_conditions` (`:1508`), **neither of which ever inspects `param.modifier`**:

```rust
fn build_resource_id_condition(&self, param: &SearchParameter) -> StorageResult<Document> {
    let mut ids = Vec::new();
    for value in &param.values {
        if value.prefix != SearchPrefix::Eq { return Err(...); }
        ids.push(value.value.clone());        // modifier never consulted
    }
    if ids.len() == 1 { return Ok(doc! { "id": ids.remove(0) }); }
    Ok(doc! { "id": { "$in": ... } })
}
```

So:

| Query | Filter produced | Result |
|---|---|---|
| `_id:not=abc` | `{id: "abc"}` | returns **only** `abc` — the exact inverse of the request |
| `_id:missing=false` | `{id: "false"}` | matches nothing; should match every resource |
| `_id:missing=true` | `{id: "true"}` | matches nothing |
| `_lastUpdated:missing=true` | — | at least loud: the value `"true"` fails date parsing and raises an error |

The REST layer does not stop these. `_id` resolves to `SearchParamType::Token` (`crates/persistence/src/types/search_params.rs:138`) and both `Not` and `Missing` are valid for Token, so the generic `is_valid_for` gate passes (`crates/rest/src/extractors/search_query_builder.rs:446`).

## Why it matters

An inverted `_id:not` silently returns precisely the data the caller asked to exclude. Anywhere `_id:not` is used to filter something out of a view, the excluded record becomes the only record returned.

## Expected behavior

`_id:not=abc` excludes `abc`. `_id:missing=false` matches every resource with an id; `_id:missing=true` matches none.

[...]

Issue #473, opened 2026-07-31
search: token :not is per-row, not per-resource — wrong on every nested/array-valued parameter


angela-helios:
Found by the #448 sweep v2 (SQLite, Synthea data): a coherence matrix over every positively-matched parameter shows `:not` fails on **19 of 19** token parameters whose values live in nested or array elements, while top-level scalars behave.

### Evidence (all manually re-verified)
| Query | `=` total | `:not` total | Expected `:not` | Cohort |
|---|---|---|---|---|
| `Patient?gender:not=male` (top-level scalar) | 16 | 11 | 11 | 27 — ✅ correct |
| `Patient?language:not=urn:ietf:bcp:47|en-US` | 24 | **27** | ≤3 | 27 — returns everyone, matches included |
| `Observation?combo-code:not=http://loinc.org|8302-2` | 85 | **1400** | 1315 | 1400 — returns everyone |
| `ExplanationOfBenefit?status:not=active` | 514 (all) | **1** | 0 | 514 — one phantom row |
| `Encounter?reason-code:not=sct|36971009` | 3 | **245** | ≥370 | 373 — missing-field resources excluded |
| `MedicationAdministration?reason-given:not=…` | 9 | **9** | 16 | 25 — returns the matches |

…plus 13 more in the sweep output (PractitionerRole specialty/role/telecom, Organization/Device type, Immunization vaccine-code, identifiers, DiagnosticReport category, Provenance agent-type, SearchParameter target).

### Reading the shapes
FHIR `:not` means *no value of the parameter equals X* — per resource, and resources without the element count as matches. The observed behaviors fit a per-**row** predicate on the search index join (`EXISTS(row != X)` or plain `value != X` on a multi-row join) instead of `NOT EXISTS(row = X)`:
- multi-row resources leak in whenever any of their rows differs (category arrays, codings),
- resources with no rows for the param drop out entirely (Encounter reason-code),
- the EOB oddity (all 514 active yet `:not` finds 1) suggests an extra indexed row from a contained resource satisfying the `!=`.

The gender case passes only because single top-level scalars have exactly one row.

Part of the #448 sanity pass, iteration 2. Likely home: the sqlite token modifier handling (`modifier_handlers.rs` / token handler) — same neighborhood as the #456 date fix.

mauripunzueta:
Design note from #1050 (Elasticsearch drops resources over its 10,000 nested-object limit).

[...]

PR #1192, opened 2026-09-16
fix(persistence): honour _id:not on sqlite/postgres/elasticsearch (#1092)

### LINKED ISSUE #1092: search: _id:not returns the excluded resource on sqlite, postgres and elasticsearch too
## Summary

`_id:not=<id>` returns **exactly the resource that was excluded** on SQLite, PostgreSQL and Elasticsearch. All three `_id` query builders read only `param.values` and never `param.modifier`, so the modifier is silently dropped and the value is consumed as a plain positive match.

This is the same defect as #1055, which was fixed for MongoDB in #1091. That issue asked whether the SQL backends shared it; **they do**, and this issue tracks the remaining three.

## Confirmed sites

| backend | file | line |
|---|---|---|
| SQLite | `crates/persistence/src/backends/sqlite/search/query_builder.rs` | 525 |
| PostgreSQL | `crates/persistence/src/backends/postgres/search/query_builder.rs` | 676 |
| Elasticsearch | `crates/persistence/src/backends/elasticsearch/search/query_builder.rs` | 262 |

Each builds an `_id` condition from `param.values` alone. The REST layer does not stop the query getting there: `_id` resolves to `SearchParamType::Token`, and `SearchModifier::Not` is valid for Token, so the generic `is_valid_for` gate passes it through.

## What is and isn't affected

- **`:not` is broken on all three.** `_id:not=abc` returns only `abc` — the precise inverse of the request.
- **`:missing` is correct on all three.** It resolves before the special-param dispatch, so it never reaches these builders. SQLite's behaviour is already pinned by `crates/persistence/tests/search/modifier_tests.rs:288`.

So the scope here is narrower than #1055 was for MongoDB: `:not` only, and no `_lastUpdated` component.

## Why it matters

Silent wrong data, not an error. A caller filtering something out with `_id:not` receives **only** the thing they excluded, with an HTTP 200 and a well-formed Bundle. In any view where `_id:not` is used to exclude a record, the excluded record becomes the entire result.

## Suggested fix

Mirror what #1091 did for MongoDB:

- Honour `Not` in each `_id` builder — `id != ?` / `id NOT IN (…)`, and the Elasticsearch equivalent.
- Reject any other modifier on `_id` explicitly rather than letting it degrade to a positive match.
- Check whether each backend has a path that reaches the builder without passing `validate_query_support` first. On MongoDB the `ifNoneExist` conditional-create path (`storage.rs:3465`) does exactly that, which is why the rejection was applied in both places there.

Worth doing as three separate PRs or one, but note each backend needs its own integration coverage, and PostgreSQL and Elasticsearch each require their own testcontainer — which is the reason this was not folded into #1091.

## Test hint

The unit tests in #1091 are the cheap model: assert the exact filter document/SQL the builder produces for `_id:not=a`, `_id:not=a,b`, and a no-modifier control, with no database needed. The live integration test then only has to confirm the page and the count agree.

---

Found while fixing #1055; the three sites above were confirmed by reading the code, not inferred.

mauripunzueta:
Closes #1092

## Summary

`_id:not=<id>` returned **exactly the resource that was excluded** on SQLite, PostgreSQL and Elasticsearch — the precise inverse of the request, with a 200 and a well-formed Bundle. `_id` is answered from the `resources` table (the top-level document on ES) through a dedicated builder that sits outside the generic modifier dispatch, and all three builders read only `param.values`, so the modifier was silently dropped. Same defect as #1055, fixed for MongoDB in #1091; this PR covers the remaining three backends and mirrors that fix.

Triage first: confirmed live on `main` at `a36d0decd` — none of the three builders read `param.modifier`, no PR since #1091 touched them, and the only in-flight search-index work (`docs/superpowers/plans/2026-09-15-mongodb-search-index-v2.md`) is MongoDB-only.

## Changes

**Builders honour `:not`** (positive form unchanged on every backend; PG's is byte-identical):

| backend | `_id:not=a` | `_id:not=a,b` |
|---|---|---|
| SQLite (`sqlite/search/query_builder.rs`) | `... AND id NOT IN (?3)` | `... AND id NOT IN (?3, ?4)` |
| PostgreSQL (`postgres/search/query_builder.rs`) | `id <> $3` | `id NOT IN ($3, $4)` |
| Elasticsearch (`elasticsearch/search/query_builder.rs`) | `{"bool":{"must_not":[{"term":{"resource_id":"a"}}]}}` | `{"bool":{"must_not":[{"terms":{"resource_id":["a","b"]}}]}}` |

`:not=a,b` is `NOT (a OR b)` everywhere, as FHIR defines it (#473).

**Explicit rejection of anything else.** Unlike MongoDB, these backends have no `validate_query_support`, the builders are infallible (`Option`), and the REST `is_valid_for` gate passes every token modifier on `_id` (`_id` resolves to Token). New shared `helios_persistence::search::reject_unhonoured_metadata_modifiers` (in `search/registry.rs`, next to `fallback_param_type`) is called at each backend's `search` / `search_count` entry **and** at `search_with_connection` (SQLite) / `search_with_client` (PostgreSQL), which are the in-transaction entries used by `ifNoneExist` / conditional criteria. It refuses with `SearchError::UnsupportedModifier` (→ 400 via the existing REST mapping):

- `_id`: honours `None`, `:not`, `:missing`; rejects the rest (`:text`, `:in`, `:above`, …)
- `_lastUpdated`: honours `None`, `:missing`; rejects the rest

`:missing` was already correct on all three (resolved before the special-param dispatch) and is untouched.

**Conditional paths, checked as the issue asked.** On SQLite/PostgreSQL (`build_search_parameters`) and the composite (`find_conditional_matches`), conditional criteria are parsed with `modifier: None` unconditionally — the raw `name` (including any `:modifier` suffix) becomes the parameter name — so `_id:not` cannot reach the builders through `If-None-Exist` today. The gate on the in-transaction entries is therefore defensive rather than closing a live hole. That the conditional parsers drop modifiers entirely is a separate, pre-existing limitation not addressed here. Elasticsearch has no conditional path of its own (the composite delegates to `search`, which is gated).

## Test plan

Unit (no database; all pass locally):
- `search::registry::tests::{honoured_metadata_modifiers_pass, unhonoured_metadata_modifiers_are_rejected_by_name, ordinary_parameters_are_not_gated}`
- SQLite `query_builder::tests::{id_without_modifier_is_a_positive_in_list, id_not_inverts_the_in_list}` — exact SQL + bound params
- PostgreSQL `query_builder::tests::id_modifier::{id_without_modifier_is_a_positive_match, id_not_negates_instead_of_matching, id_not_respects_the_running_offset}` — exact SQL + params, including gap-free `$n` numbering after another parameter
- Elasticsearch `query_builder::tests::id_modifier::{id_without_modifier_is_a_positive_term, id_not_negates_instead_of_matching}` — exact clause JSON

Integration (3 patients; `_id=1` control, `_id:not=1` → {2,3}, `_id:not=1,2` → {3}, `search_count` agrees with each page, `_id:text` refused by both `search` and `search_count`):
- SQLite: `search_suite::search::modifier_tests::{test_id_not_modifier_excludes_the_named_resources, test_unhonoured_id_modifier_is_rejected}` — **pass locally**
- PostgreSQL: `postgres_integration::postgres_integration_search_id_not_modifier` — testcontainer; **CI only** (no Docker in the dev session)
- Elasticsearch: `es_integration::es_integration_search_id_not_modifier` — testcontainer; **CI only**

Also run locally: `cargo fmt --all --check` (clean); full `search::query_builder` + `search::registry` unit modules (110 pass); full `modifier_tests` module (21 pass); `cargo clippy -p helios-persistence --all-features --tests --no-deps -- -D warnings` with only lint kinds that fire on untouched files allowed.

Note for reviewers: with the local toolchain (clippy 1.96.1) an unfiltered clippy run fails on pre-existing lints in files this PR does not touch (`collapsible_if` on `if let` chains, `doc_lazy_continuation`, `question_mark`, `collapsible_match`, `large_enum_variant`, `field_reassign_with_default`, `items_after_test_module`); none of the hits fall inside this PR's hunks. Toolchain drift, not part of this change — CI's `stable` clippy is the arbiter.


smunini:
Closing as a duplicate of #1190, which was merged into `main` earlier today (b42abec13). Both PRs close #1092 with the same fix on the same three backends: a `:not` branch in each `_id` builder plus a shared gate that rejects other `_id` modifiers with `UnsupportedModifier`.

The one thing this PR covers that #1190 does not is gating unhonoured modifiers on `_lastUpdated`. That will land as a small follow-up on top of `main`. Thanks for the thorough write-up.

PR #1207, opened 2026-09-16
fix(export): keep a SQLite kick-off from dying behind an index rebuild

### LINKED ISSUE #1185: sqlite: $export kick-off fails with "database is locked" (500 after 30 s) while a search-index rebuild is running; UI shows a raw "error sending request" after its 15 s timeout
## Summary
On the SQLite backend, kicking off a Bulk Data `$export` while a post-import search-index rebuild is running fails with a 500 after ~30 s: the single-row insert into `bulk_export_jobs` waits out the full `busy_timeout` behind the rebuild's write lock and then gives up with `database is locked`. The UI's kick-off request times out even earlier (15 s), so the user sees a bare `error sending request for url (http://127.0.0.1:8080/$export)` instead of any explanation.

## Current behavior
1. Run a large `$bulk-submit` import on SQLite with deferred indexing, so the automatic search-index rebuild is still running afterwards (the Resources page shows the "Search index rebuilding" notice).
2. From the Bulk Export UI page, start a system-level export (or `GET /$export` with `Prefer: respond-async` directly).

The UI job card shows:

```
error sending request for url (http://127.0.0.1:8080/$export)
```

Server log:

```
2026-09-16T14:32:18.654678Z ERROR helios_rest::error: internal error while processing request error.detail=internal error in sqlite: Failed to create export job: database is locked
2026-09-16T14:32:18.655016Z ERROR tower_http::trace::on_failure: response failed classification=Status code: 500 Internal Server Error latency=30041 ms
```

The 30041 ms latency is `busy_timeout_ms` (default 30 000) expiring, i.e. no other writer released the lock for a full 30 s.

## Expected behavior
Creating an export job is a single small insert and should not be blocked for tens of seconds by a background rebuild. Either it succeeds promptly (the rebuild yields the write lock often enough), or, if the store genuinely cannot accept the write right now, the client should get a clear 503 with `Retry-After` rather than a 500 and a client-side timeout. The UI should surface that as a readable message rather than the raw reqwest error.

## Evidence
- `crates/persistence/src/backends/sqlite/bulk_export.rs:96-132` — `create_export_job` does one autocommit `INSERT INTO bulk_export_jobs` on a pooled connection; any `rusqlite::Error` becomes `internal_error(...)` → HTTP 500. Unlike #791 this is not a DEFERRED read-then-write upgrade; the busy handler *was* invoked and simply ran out of time.
- `crates/persistence/src/backends/sqlite/backend.rs:206-215` — `default_busy_timeout_ms` is 30 000 ms, matching the observed latency.
- `crates/persistence/src/backends/sqlite/storage.rs:4447-4491` — `begin_bulk_index_rebuild` drops all thirteen `search_index` value indexes for the run and `end_bulk_index_rebuild` recreates them via `schema::ensure_search_value_indexes` on a single connection. Each `CREATE INDEX` in `crates/persistence/src/backends/sqlite/schema.rs:30-74` holds the write lock for its whole duration, and on a large corpus a single one can exceed 30 s (see #1125 for rebuild timings measured in the hundreds/thousands of seconds). The per-page rebuild writes themselves are also back-to-back IMMEDIATE transactions with no yield between them.
- `crates/ui/src/bulk_export.rs:1039-1043` — the UI kick-off request uses a 15 s timeout, so it fails before the server responds; the reqwest error is stored verbatim as `job.error` at `crates/ui/src/bulk_export.rs:1080-1084`, which is the `error sending request for url (...)` string the user sees.

## Suggested approach
Any one of these fixes the symptom; the first two address the cause.

1. **Bound the rebuild's lock hold time.** Keep the end-of-run `CREATE INDEX` statements but interleave them (they are already separate statements — make sure they are not wrapped in one transaction), and consider building them with a lower priority or in smaller batches. For the page-write phase, add a short yield between page transactions so foreground writers can interleave.
2. **Map lock exhaustion to 503, not 500.** In the SQLite backend (or a shared helper) detect `SQLITE_BUSY` / `database is locked` on the export-job insert and return a `StorageError` that `helios_rest::error` renders as `503 Service Unavailable` with a `Retry-After`, similar to what `$bulk-submit` does for its pre-ingest phases (#953). Also worth applying to the other small foreground writes that race a rebuild (bulk-submit status, user settings).
3. **UI: make the kick-off failure legible.** Either raise the kick-off timeout above `busy_timeout_ms`, or catch `reqwest::Error::is_timeout()` and render "the server did not answer the export kick-off within 15 s; a search-index rebuild may be holding the database, try again shortly" instead of the raw error string.

Also worth checking: whether `get_export_status` / `cancel_export` and the export worker's own writes hit the same wall mid-rebuild, and whether PostgreSQL is unaffected (it should be — the same insert at `crates/persistence/src/backends/postgres/bulk_export.rs:130` does not contend on a single writer lock).

## Scope / out of scope
- In scope: `$export` kick-off failing on SQLite during an index rebuild, its HTTP status, and the UI message.
- Out of scope: rebuild performance in general (#1125, #945) and the `$bulk-submit` lock-contention issues already tracked (#942, #791).

eterceros:
Closes #1185

On SQLite, `$export` kick-off returns a 500 when a background search-index rebuild holds the write lock: `SqliteBackend::start_export` stringifies every `rusqlite::Error` into `BackendError::Internal`, dropping the `SQLITE_BUSY` discriminant, and `map_storage_err` flattens the unmatched arm to `InternalError`. The persistence layer already has `classify_sqlite_error` mapping BUSY/LOCKED to `BackendError::Unavailable`, and helios-rest already renders that as 503 with `Retry-After`, so the fix wires the bulk-export paths through the existing classifier and delegates the unmatched arm to `RestError::from`. The per-page rebuild loop also re-took the write lock the instant it released it, and the UI kick-off showed reqwest's raw `Display`; both are addressed.

A second defect surfaced while testing the first: `query_map` only binds, so the first `sqlite3_step` — and any `SQLITE_BUSY` with it — lands on the per-row `Result`. Every read ran those through `filter_map(|r| r.ok())`, reporting a failed read as an empty one: a manifest with no files, a status with no progress, an export finishing "complete" having written nothing. Those ten sites now collect through their error.

**Scope:** this classifies the contention, it does not remove it. The end-of-run `CREATE INDEX` statements still hold the write lock for their whole duration and no yield can split them, so on a large corpus a kick-off can still answer 503 — correctly, and retryably, which is the point.

**Verified against a live server:** with the write lock held past `busy_timeout`, kick-off answers `503` + `retry-after: 5` + `code: transient` after 30.031 s (the issue reported 30041 ms and a 500); once released it answers 202. During a real `$reindex` of 1,382 resources the kick-off returns 202 in 0.059 s. The UI card shows `kick-off failed: timed out after 15s — a search-index rebuild may be holding the database, try again shortly` instead of `error sending request for url (...)`, and its Retry button then completes the job.


eterceros:
The `Linting` failure was not introduced by this branch.

The Windows self-hosted agent carries clippy 1.91, where `const_is_empty` const-folds `is_empty()` on anything reachable through `env!`. `build_info.rs` asked `!GIT_SHA_RAW.is_empty()` on a const derived from `env!("HFS_BUILD_GIT_SHA")`, so the question had a constant answer in every build — always `false` where a commit resolved, always `true` where none did — and `-D warnings` rejected it. It does not fire on stable 1.98, which is why local runs and the Linux jobs were green.

Fixed in 7f0ef1a62 by making the unknown-commit case the variable's absence (`build.rs` emits `HFS_BUILD_GIT_SHA` only when a commit is known, `build_info.rs` reads it with `option_env!`) instead of an empty-string sentinel, so there is no const check left to fold.

Two things worth flagging for whoever looks next:

- The same toolchain also rejects a `u64 -> u64` cast in `crates/persistence/tests/postgres_tests.rs`, introduced by #1159 and inherited here through the merge of main. CI never reported it because `helios-rest` had already failed the run. It is fixed in the same commit, but it means **main is currently red on this job too**.
- Reproduced and verified locally with `cargo +1.91.1 clippy --all-targets --all-features` and the job's exact flag list, which now passes across the whole workspace. The 1.91-vs-1.98 gap is the real hazard here: anything the agent's older clippy rejects is invisible to `cargo clippy` on stable.

smunini (approved):

PR #1209, opened 2026-09-16
fix(rest): allow clippy::const_is_empty on git_sha (#1208)

### LINKED ISSUE #1208: ci: Linting fails with clippy::const_is_empty in build_info.rs on runners with no git sha
## Summary

The Linting job fails with `clippy::const_is_empty` in `crates/rest/src/build_info.rs` whenever it lands on a runner where the build script could not determine the git commit. The `[self-hosted]` pool has both Linux and Windows runners; on the Windows one the sha comes back empty, so the emptiness check on a compile-time constant becomes a constant-false expression and clippy rejects it under `-D warnings`. On Linux the constant is non-empty and the same code lints clean, which is why the failure looks random per PR.

## Current behavior

PR #1201, a change to a Playwright spec only, went red on Linting after an earlier run of the same PR had passed:

```
error: this expression always evaluates to false
  --> crates\rest\src\build_info.rs:41:7
   |
41 |     (!GIT_SHA_RAW.is_empty()).then_some(GIT_SHA_RAW)
   |       ^^^^^^^^^^^^^^^^^^^^^^
   = note: `-D clippy::const-is-empty` implied by `-D warnings`
error: could not compile `helios-rest` (lib) due to 1 previous error
```

Job: https://github.com/HeliosSoftware/hfs/actions/runs/35141658738/job/104947388776 (`git version 2.51.0.windows.1`). The Linting jobs on #1200 and #1205 ran on Linux runners (`git version 2.48.1`) and passed.

## Expected behavior

Linting is independent of which runner picks the job, and of whether the build could read a git sha.

## Evidence

- `crates/rest/src/build_info.rs:19` — `const GIT_SHA_RAW: &str = env!("HFS_BUILD_GIT_SHA");` is documented as `""` when the sha is unknown.
- `crates/rest/src/build_info.rs:41` — `git_sha()` tests that constant for emptiness; clippy folds the test when the constant is empty.
- `crates/rest/build.rs:24-36` — the sha falls through `HFS_GIT_SHA`, then `git rev-parse --short HEAD`, then `""`; the empty case is by design (tarball builds).
- Introduced in dde1ac446 (#992, `hfs --version` / `CapabilityStatement.software`).

## Suggested approach

The emptiness of `GIT_SHA_RAW` legitimately varies per build, so the lint is a false positive here: allow `clippy::const_is_empty` on `git_sha()` with a comment saying why. No behaviour change.

Separately worth a look, but not this issue: the Windows self-hosted runner produced an empty sha at all, so `hfs --version` built there prints no commit. Likely `git rev-parse` failing in the build script's environment (`safe.directory`, or `git` off `PATH` for the toolchain).

dougc95:
Closes #1208.

The Linting job fails with `clippy::const_is_empty` in `crates/rest/src/build_info.rs` whenever it lands on a runner where the build script found no git commit. `GIT_SHA_RAW` is `env!("HFS_BUILD_GIT_SHA")`, which `build.rs` sets to `""` in that case by design (tarball builds, `git` unavailable, and in practice the Windows self-hosted runner), so on those builds the emptiness test in `git_sha()` folds to a constant and clippy rejects it under `-D warnings`. On builds with a sha the same code lints clean, which is why PR #1201 went red on a spec-only change after passing an hour earlier: it drew the Windows runner the second time, while #1200 and #1205 drew Linux ones.

## What changes

`#[allow(clippy::const_is_empty)]` on `git_sha()`, with a doc comment saying why the value varies per build and the check is intended. No behaviour change.

## Verification

fmt clean; the CI clippy command clean for `helios-rest`; the three `build_info` unit tests pass. The false positive itself could not be reproduced on the local 1.98 toolchain even with an empty sha forced through `build.rs` (the runner that failed reports clippy from Rust 1.91, whose heuristics for this lint are broader); the fix is the exact attribute the runner's own diagnostic suggests, on the one function it names.

Not in scope: why the Windows self-hosted runner cannot read the sha at all, noted on the issue.

https://claude.ai/code/session_013Uo4p2nXn3U3j54MyW4Aqd


smunini:
Closing as superseded. The underlying fix already landed on `main` via 7f0ef1a62 (PR #1207: `build.rs` now emits `HFS_BUILD_GIT_SHA` only when a commit is known and `build_info.rs` reads it with `option_env!`, so there is no const emptiness check left to fold), and PR #1201 also carried the same `#[allow]`. That leftover allow and its comment are being removed in a follow-up.

Issue #998, opened 2026-09-08
bulk-submit: a submission can never reach a terminal state — status poll and Mark completed both time out, and the import re-ingests forever


dougc95:
## Summary

At corpus scale a bulk-import submission can never reach a terminal state. The automatic status poll always times out, and the manual `Mark completed` escape hatch fails too — for a different reason, and then loses a write race on top. Combined with #968 and #969 this leaves the submission `in-progress` forever and the manifest permanently re-claimable, so the import re-runs from the beginning indefinitely.

## Current behavior

Measured against a `mongodb` primary holding 11.2 M resources.

**1. Every status poll times out.** The Import page polls with a 10 s timeout while the endpoint takes 26 s:

```console
$ curl -s -o /dev/null -w 'HTTP %{http_code} time=%{time_total}s\n' \
    localhost:8080/bulk-submit-status/71da80d2-33f9-4053-add4-e2fdd37037b8
HTTP 202 time=26.089165s
```

The Submission Log fills with one line every ~17 s:

```
2026-09-07T19:14:31.014Z Status poll failed: error sending request for url (http://localhost:8080/bulk-submit-status/71da80d2-…)
```

and the progress card freezes on the last successful poll — it read `processing 65% complete (133304 entries ingested)` for hours while 11.1 M resources had actually landed.

**2. `Mark completed` never applies.** `POST /ui/bulk-import/{id}/complete` returns 303 after **15.24 s**, exactly the `post_kickoff` timeout. In `set_status`, `s.status = status` is assigned only inside the success arm, so a timed-out kick-off leaves the status untouched and only appends `Status change failed: {e}` to the log. From the UI it looks like the button does nothing: it spins, the page reloads, still `In Progress`.

**3. That log write then loses an optimistic-lock race.** The page's own status card refreshes every 5 s and each failed poll rewrites the same submission document, so the user action loses:

```
WARN helios_ui::bulk_import: failed to persist bulk-import submission
  submission=f6b8a709-… error=optimistic lock failure:
  resource BulkProviderSubmission/f6b8a709-… has been modified
```

51 occurrences between 19:11Z and 20:37Z, roughly one every 16 s.

**4. The import therefore never stops.** Because the submission stays non-terminal, the manifest stays `processing` and remains eligible for `claim_next_manifest`; when the lease lapses a worker re-claims it and (#969) re-walks each file from the top. Observed 3 h 45 min *after* ingestion had finished:

[...]

PR #1230, opened 2026-09-17
fix(persistence): restore main compilation after the #1194/#1159 and #1203/#1204 merges

mauripunzueta:
## Summary

`main` does not compile since this morning's merges (`c7110b833` and the three commits before it are red on Linting and Test Rust). Two pairs of PRs each passed CI on their own branches but conflict once both are on `main`. This PR fixes both with the smallest change that matches what each side intended.

### `complete_submission` signature (#1194 × #1159)
#1194 changed `BulkSubmitProvider::complete_submission` to return `StorageResult<()>`, so the status transition never aggregates receipts (#998). #1159 added `IndexingSubmitJobs`, whose forwarding impl still declared `StorageResult<SubmissionSummary>`. That caused E0053 at `composite/indexing_submit_jobs.rs:382` and E0308 at `:433`. The impl now returns `StorageResult<()>`, like every other backend and composite.

### `BatchFtsError` in `index_fts_content_batch` (#1203 × #1204)
#1204 made the error mapping in `index_fts_content_batch` explicit: every failure is a `BatchFtsError`. #1203's empty-content `DELETE FROM resource_fts` in the same function still used `?` on a bare `StorageError`, causing E0277 at `postgres/storage.rs:1890`. It is now wrapped in `BatchFtsError::Other`, exactly as the upsert right below it is.

### A test #1194 made flaky (MongoDB)
Once main compiled again, `bulk_submit::test_active_submission_count_and_expiry_scan` failed in `mongodb_tests` (CI run 35226096836): `assertion failed: expired.iter().any(|(_, sub)| sub == &id)`. The test completes a submission and immediately scans with `list_expired_submissions(Utc::now(), 0s)`. The scan selects `updated_at < now - ttl`, and `updated_at` is stored at millisecond precision, so a scan in the same millisecond as the write correctly excludes it. It used to pass only because `complete_submission` re-read the whole summary after its update, which took longer than a millisecond. #1194 removed that re-read. The test now scans from one second after the write. The strict comparison in the product code is unchanged.

No behaviour change beyond restoring the build and that test.

## Test plan
- [x] `cargo check -p helios-persistence -p helios-rest -p helios-hfs --tests` (sqlite, postgres, mongodb, elasticsearch, s3, ui, subscriptions): clean, 0 warnings
- [x] `cargo test -p helios-persistence --lib --features sqlite,postgres,mongodb,elasticsearch,s3`: 1,664 passed
- [x] `cargo test -p helios-persistence --features postgres --test postgres_tests -- reindex fts`: 25 passed, including the page/FTS recovery paths from #1203 and #1204
- [x] `cargo fmt --check` clean; `cargo clippy -p helios-persistence --all-features --lib --tests` reports nothing in the changed lines
- [x] Coverage for both repaired lines (Codecov first reported 14% patch coverage):
  - `composite::indexing_submit_jobs::tests::complete_submission_is_forwarded_to_the_inner_store`: completing through the wrapper reaches the store, and a second completion is rejected, not swallowed.
  - `postgres_integration_reindex_page_replays_a_failed_empty_fts_delete`: a trigger rejects the first `DELETE` of a stale `resource_fts` row. The page ends `Ok` with the row gone after exactly two attempts: the rejected batched delete, then the per-resource replay.
- [ ] CI


mauripunzueta:
`claude-review` fails on this PR before the model is ever called, so it isn't about the change (a two-line compile fix). We can't see why, because the action hides the result text even with debug logging. Could someone with access to the self-hosted runners (@smunini?) share the `result` field from `_work/_temp/claude-execution-output.json` for any of these runs?

| Attempt | Job | Runner | Session id | Duration |
|---|---|---|---|---|
| 1 | [105217947837](https://github.com/HeliosSoftware/hfs/actions/runs/35226096612/job/105217947837) | `github-agent1` | `730410b8-064c-44aa-86c4-264459daea6d` | 634 ms |
| 2 | [105219418639](https://github.com/HeliosSoftware/hfs/actions/runs/35226096612/job/105219418639) | `github-agent1` | `12107c79-e326-4555-9523-79e158722d8f` | 728 ms |
| 3 (debug) | [105223049979](https://github.com/HeliosSoftware/hfs/actions/runs/35226096612/job/105223049979) | `github-agent7` | `6fa18247-43fc-49dc-a6fc-f1b78d3e58e9` | 481 ms |

Every attempt ends with `"subtype": "success", "is_error": true, "total_cost_usd": 0, "modelUsage": {}`. `Claude Code initialized` (model `claude-sonnet-5`) is logged, then the run returns an error within one turn without making an API call.
- Two runners fail identically, so it isn't one machine.
- The same failure happened on #1222 (`feat/1160-contained-rows-collection`, runs at 00:58 and 01:53 today), which was merged.
- Other PRs reviewed successfully in between (e.g. `fix/1058-mongo-sorted-cursor-link` at 12:50).

Meanwhile this PR is what brings `main` back to compiling, so it doesn't need to wait on the review check. Locally:
- `helios-persistence` lib tests pass with all backends: 1,664.
- The PostgreSQL reindex/FTS integration tests pass: 25.
- `cargo fmt` is clean.
- Clippy reports nothing new in the changed lines.

mauripunzueta:
Update on `claude-review`: the cause is known. The Claude account behind `CLAUDE_CODE_OAUTH_TOKEN` is out of usage, which fits a run that errors immediately with `total_cost_usd: 0` and empty `modelUsage`. Nothing to dig into on the runners. It will pass again once usage is available.

angela-helios:
Heads-up from #1237, which is stacked on this branch because `main` does not compile without it: the one failing check here, **Test Rust**, is a single test — `mongodb_tests::bulk_submit::test_active_submission_count_and_expiry_scan` (`mongodb_tests.rs:10863`, `assertion failed: expired.iter().any(|(_, sub)| sub == &id)`, 173 passed / 1 failed) — and it fails identically on #1237, so it comes with the base rather than with either diff.

One thing that may save time: the MongoDB `list_expired_submissions` is `find({updated_at: {$lt: cutoff}}).sort({updated_at: 1}).limit(10)`. With a zero TTL every submission in the shared test database qualifies, so the ten *oldest* come back and the submission the test just completed is not among them once more than ten exist — which #1194's additional bulk-submit tests would do. That reads like test isolation (or the limit) rather than a regression in `complete_submission`, but I have not run the MongoDB suite here (no Docker engine on my host), so treat it as a pointer, not a diagnosis.

smunini (approved):

Issue #1166, opened 2026-09-16
Manual test pass: MinIO backend (s3 against MinIO)


smunini:
## Summary
Run the manual testing pass in `MANUAL_TESTING_MATRIX.md` for the **MinIO (S3 backend against MinIO)** configuration (`HFS_STORAGE_BACKEND=s3`) with authentication disabled, and record the results in this issue.

The matrix is organised backend-first: every backend gets the same T0 through T9 sequence. This issue covers the `s3` row exactly as the matrix describes it, with MinIO as the S3-compatible store. The same row against AWS S3 is #1168.

## Procedure
Follow `MANUAL_TESTING_MATRIX.md` sections 2 through 14 in order, using the `s3` row of the per-backend environment table in section 5.

Infrastructure: MinIO from section 4 of the matrix (`hfs-minio` container with the `hfs`, `hfs-export`, `hfs-sql-export` buckets). Reset with `docker rm -fv hfs-minio` for a clean start.

- [ ] T0 Build with the CI selection: `cargo build --workspace --all-features --release`
- [ ] T1 Start HFS with `HFS_STORAGE_BACKEND=s3` and confirm `/health`, `/metadata`, and `/ui`; the startup log names the S3 backend, the `hfs` bucket, and the MinIO endpoint
- [ ] T2 Batch / Transaction page: batch bundles only (6.2 and 6.4). The transaction bundle is refused on the S3-family backends; confirm the refusal message matches 6.1
- [ ] T3 Import the Synthea corpus from the Import page (record elapsed time, file count, resource total, and the 7.4 count check)
- [ ] T4 Search types: N/A on standalone S3 (no search index). Confirm the Search page and `GET /Patient` fail with a clear OperationOutcome rather than an error page
- [ ] T5 Bulk export: N/A (501, no job store on S3). Confirm the Export page reports this rather than erroring; the S3 output-backend variant in 9.5 is not applicable
- [ ] T6 Create `patient_demographics` and `observation_flat` ViewDefinitions on `/ui/sql/view-definitions` and examine `$sql-run` output in json, csv, ndjson, parquet (in-process scan runner on S3)
- [ ] T7 `$sql-export` with a ViewDefinition, a SQL Query, and a SQL View subject, cycling all four formats; UI `/ui/sql/export`
- [ ] T8 Create a SubscriptionTopic and rest-hook Subscription, trigger notifications, verify delivery, `$status`, `$events`
[...]

PR #1237, opened 2026-09-17
fix(s3): bulk-submit batches rolled back the lease they ran under, so the other worker reclaimed it every 60 s (#1229)

### LINKED ISSUE #1229: s3 backend: bulk-submit lease lost / reclaimed by the other in-process worker every ~60 s during a large manifest
## What happens

`HFS_STORAGE_BACKEND=s3` (MinIO, auth off, main `25a53bd0d`), defaults: `HFS_BULK_SUBMIT_WORKER_CONCURRENCY=2`, `LEASE_DURATION=60`, `HEARTBEAT_INTERVAL=20`. Importing an 11-file, 556k-resource Synthea manifest from the Import page:

- 80 s after kick-off, and then roughly **every 60 s for the whole run** (96 times in 90 min), the log shows

```
WARN helios_persistence::core::bulk_submit_worker: bulk-submit lease lost: another worker reclaimed the manifest; abandoning this run submission=urn:helios:hfs:bulk-submit|bf1674da-… worker=… fencing_token=…
WARN helios_persistence::core::bulk_submit_worker: bulk-submit run abandoned mid-manifest: its lease is no longer held …
```

- the other worker resumes from the cursor, so the submission keeps advancing, but each hand-over abandons the current file download (the fixture HTTP server logs a `BrokenPipeError` at every hand-over) and the effective rate was ~28 resources/s (149k resources in 90 min). The submission was aborted from the Import page (`Stopped`, acknowledged cleanly).
- A second, small manifest (11 files, 16,503 resources: full Patient + reference data + one patient's clinical rows) completed in 10 min 23 s with **zero** lease losses at the same ~28 res/s, so the thrash needs the large files.

`crates/persistence/src/backends/s3/submit_worker.rs`: `claim_next_manifest` reclaims a `Processing` manifest when `lease_expiry < now`; `heartbeat` renews through `fenced_mutate` (CAS on the submission object, retrying `PreconditionFailed`). The renewal evidently does not land within the 60 s lease while a big file is streaming — whether the heartbeat is starved by the ingest, or its CAS keeps losing to the worker's own progress writes on the same object, is not determined here.

## Expected

A single process with two workers must not steal its own manifest; the heartbeat has to keep the lease alive for the duration of a file, or the claim path should refuse to reclaim a manifest whose holder is alive in the same process. Worth checking with `HFS_BULK_SUBMIT_WORKER_CONCURRENCY=1` and with a longer `HFS_BULK_SUBMIT_LEASE_DURATION` to separate the two hypotheses.

Found during the #1166 MinIO pass (T3). Log: `hfs-s3.log` lines 57–227 in the pass artifacts.

angela-helios:
Fixes #1229.

Rebased onto `main` at `6f58ff8fa` now that #1230 is merged: a single commit, `backends/s3/*` plus one log line in `core/bulk_submit_worker.rs`.

## The bug

On the S3 backend a manifest's lease (`worker_id`, `lease_expiry`, `fencing_token`) lives in the same object as its counters. `claim_next_manifest` and the lease keeper's `heartbeat` compare-and-swap that object. `process_entries` did not: it read the state before the batch and **wrote that whole copy back after it**, lease fields included.

Reproduced on MinIO (the only environment that shows it), default 2 in-process workers, 60 s lease / 20 s heartbeat, the 556k-resource manifest from #1166, with every lease event logged:

```
13:28:51.309 lease acquired  worker-1 token=1 held_until=13:29:51
13:29:11.320 lease renewed   worker-1 token=1 held_until=13:30:11
13:29:12.546 end-of-batch save: stored_expiry=13:30:11  written_expiry=13:29:51   <- renewal rolled back 1.2 s later
13:29:14.037 end-of-batch save: stored_expiry=13:29:51  written_expiry=13:29:51   <- next batch read the rolled-back value
13:29:31.332 lease renewed   worker-1 token=1 held_until=13:30:31
13:29:31.789 end-of-batch save: stored_expiry=13:30:31  written_expiry=13:29:51   <- rolled back again
13:29:53.509 lease acquired  worker-0 token=2 previous_holder=worker-1 previous_expiry=13:29:51   <- "expired" exactly 60 s after the claim
13:30:11.342 WARN lease lost worker-1 token=1 held_until=13:30:51 (what the keeper believed)
13:30:55.402 lease acquired  worker-1 token=3 previous_holder=worker-0 previous_expiry=13:30:53
13:31:35.425 lease acquired  worker-1 token=3 previous_holder=worker-0 previous_expiry=13:30:53   <- token 3 acquired a second time
```

So the answer to "why does worker B think A's lease expired" is: **because in storage it had.** The batch in flight when a renewal landed wrote the pre-renewal expiry back about a second later; the next batch read that; the stored expiry therefore never moved past the one the claim had set (14 renewals, 78 rolled-back writes in five minutes). During a long file, batches run back to back and no renewal survives, so the other worker reclaims exactly one lease duration after every claim — from a holder that is alive and heartbeating on schedule. With small files the renewals land in the gaps between files and survive, which is why the 16k-resource manifest in #1166 never lost its lease.

Each reclaim abandons the download in progress and re-walks the manifest from its first file (#1127): `Condition.ndjson` was fetched **6 times in the first five minutes**.

The last line of the log is the second half of the same bug: the whole-object write also put a fenced-out holder's `worker_id`/`fencing_token` back over the new holder's, so a fencing token was handed out twice and the fenced-out worker kept ingesting for another minute.

Renewal cadence and claim-time ownership checks were both fine; neither the keeper nor the claim path changes.

## The fix

`process_entries` on S3 makes its two manifest writes (status at the start, counter deltas at the end) through a new `mutate_manifest_state`: compare-and-swap onto the state **as stored**, touching only what a batch owns and never the lease fields. SQLite, PostgreSQL (column `UPDATE`s) and MongoDB (`$set`/`$inc`) were never affected.

Also: lease acquisition (with the previous holder and its stored expiry) and renewal are logged at `debug`, and the lease-lost warning carries `held_until`/`now`, so the next lease question can be answered from a log. That is the instrumentation used above, minus a temporary probe.

## Tests

Two regression tests in the S3 worker test module drive the interleaving deterministically through the `BatchCommitObserver`, which fires between a batch's entry writes and its manifest write:

- `a_batch_does_not_roll_back_a_lease_renewed_while_it_ran` — a heartbeat lands mid-batch; afterwards a second worker must not be able to claim, the holder's next heartbeat succeeds, and the batch's counters landed.
- `a_batch_that_outlives_its_lease_does_not_undo_the_reclaim` — a reclaim lands mid-batch; afterwards the new holder still holds and the old one is still fenced out.

Both **fail without the fix** (checked by reverting `bulk_submit.rs` alone) and pass with it. Locally: `cargo fmt --all -- --check` clean; clippy with CI's flags on `helios-persistence --all-targets --all-features` clean; the `backends::s3` + `core::bulk_submit*` test modules 210 passed.

## Before / after (same host, same manifest, empty MinIO each time, 2 workers)

| | before (`c7110b833` + logging) | after (this branch, measured before the rebase; re-checked after it, see the comments) |
|---|---|---|
| lease acquisitions, first 5 min | 5 | **1** |
| lease lost | 4 (one per ~60 s; 96 in 90 min in #1166) | **0** in 20 min |
| renewals that persisted | 0 of 14 | 59 of 59 |
| `Condition.ndjson` downloads, first 5 min | 6 | **1** |
| "resources written" at +295 s | 14,100 (48/s, counts re-walked entries again) | **19,700 (67/s)** |
| later | — | 31,000 at +10 min, 42,100 at +15 min, 49,400 at +20 min, still 1 acquisition / 0 lost while `cargo test` + clippy were saturating the host |

Honest note on throughput: removing the thrash is worth ~1.4x in the first window and stops the re-walks, but it does not make S3 ingest fast. Once the run is into the larger Encounter resources it settles around 35-40 resources/s on this MinIO, because every entry is still several sequential PUTs. That is a separate problem from the lease and is not addressed here.


angela-helios:
CI on `36aadddd2`: **Linting ✅ · Security Audit ✅ · Cleanup ✅ · Test Rust ❌ · claude-review ❌** — and both failures are inherited from the base, not from this change:

- **Test Rust** fails on exactly one test, `mongodb_tests::bulk_submit::test_active_submission_count_and_expiry_scan` (`crates/persistence/tests/mongodb_tests.rs:10863`, `assertion failed: expired.iter().any(|(_, sub)| sub == &id)`; 173 passed, 1 failed). #1230's own head (`2b3b3d3a3`, which this branch is stacked on) fails **the same test with the same result** in its run — job 105217950597. This PR touches only `backends/s3/*` and a log line in `core/bulk_submit_worker.rs`; nothing on the MongoDB path.
- **claude-review** errored inside the action (`Internal error: directory mismatch … tsconfig.json` → `Claude execution failed: result is_error:true`); #1230 shows the same.

Everything this change can influence is green locally and in CI's lint job: `cargo fmt --all -- --check`, clippy with CI's flags (`helios-persistence --all-targets --all-features`), and the `backends::s3` + `core::bulk_submit*` modules (210 passed, including the two new regression tests, which fail without the fix).

Holding off on requesting review until the base is green: I will rebase onto `main` as soon as #1230 (or whatever fixes that MongoDB test) lands and re-run CI.

angela-helios:
Rebased onto `main` `6f58ff8fa` (single commit `6424c6945`; #1230's commit dropped out on its own) and force-pushed with lease. Re-checked after the rebase:

- **Lease fix still holds** — `hfs 0.2.2 (git 6424c6945)`, 2 in-process workers, the 556k manifest, empty MinIO, 10 minutes, with the host saturated by clippy + `cargo test` for most of it: **1 acquisition, 30 renewals, all persisted (`held_until` advancing 19:40:43 → 19:50:44), 0 leases lost, 0 heartbeat timeouts**, `Condition.ndjson` downloaded once. Same picture as the pre-rebase run, so the before/after table above stands.
- Local: `cargo fmt --all -- --check` clean, clippy with CI's flags on `helios-persistence --all-targets --all-features` clean, `backends::s3` + `core::bulk_submit*` 210 passed (both regression tests included).
- The MongoDB test that failed on the old base, `bulk_submit::test_active_submission_count_and_expiry_scan`, was a test bug and is already fixed on `main` by `d6d4dc2f8` ("query the zero-TTL expiry scan strictly after the completion write"); `main`'s own Test Rust is green with it. Not related to this change, nothing left to file.

angela-helios:
**Patch coverage: 33.33% → 93.93%** (`6424c6945` → `2d19364dd`), target 85.28%. All checks green on `2d19364dd`.

- What Codecov measures here: the Code Coverage job runs `cargo llvm-cov --workspace --features postgres,mongodb`, so `backends/s3/*` — where the fix lives — is not in the report. The patch figure only counts the keeper's log lines in `core/bulk_submit_worker.rs`. The S3 regression tests (`a_batch_does_not_roll_back_a_lease_renewed_while_it_ran`, `a_batch_that_outlives_its_lease_does_not_undo_the_reclaim`) run under Test Rust with the `s3` feature.
- Added in `2d19364dd`: the keeper tests' capture subscriber takes a verbosity; the reclaim test asserts the warn's `held_until`/`now`, and a new test asserts every landed renewal is logged with the expiry it was granted.
- Not covered, deliberately: exhausting the 5 CAS attempts in `mutate_manifest_state`. The mock S3 client has no way to inject a persistent ETag conflict, and a timing-based test would be brittle.
- The earlier Test Rust failure on this head was not a test: the job landed on the Windows runner (`github-agent4`) and `helios-hts-ui` did not compile because of the `locales` symlink (#1257, fixed on main by #1266). Re-run on a Linux runner passed. This branch predates that fix, so a future re-run can hit it again until it is rebased on `main`.

smunini (approved):

PR #1255, opened 2026-09-17
fix(s3,sof): resolve canonicals and list SQL definitions by scan on standalone S3 (#1228)

### LINKED ISSUE #1228: s3 backend: $sql-run / $sql-export with a Library subject (or subjectCanonical) fail with HTTP 500 — canonical lookup needs search
## What happens

`HFS_STORAGE_BACKEND=s3` (MinIO, auth off, main `25a53bd0d`). A ViewDefinition subject referenced by id works (`$sql-run` json/csv/ndjson/parquet and `$sql-export` all complete; 11,704 rows). Anything that has to resolve a **canonical** fails with a 500:

```
POST /$sql-run  {subjectReference: Library/<sql-view id>}         → 500 "An internal error occurred while processing the request."
POST /$sql-run  {subjectReference: Library/<sql-query id>, parameters…} → 500
POST /$sql-run  {subjectResource: <inline sql-view Library>}       → 500
POST /$sql-run  {subjectCanonical: http://example.org/ViewDefinition/patient_demographics} → 500
POST /$sql-export with any Library subject                         → 500 (7 of the matrix's 11 T7 jobs)
```

Server log:

```
ERROR helios_rest::error: internal error while processing request error.detail=canonical lookup failed for ViewDefinition url=http://example.org/ViewDefinition/patient_demographics: capability 'search' not supported by S3
```

`crates/rest/src/handlers/sof/references.rs:66` (`resolve_by_canonical_url`) runs a `url=` search to resolve `relatedArtifact.depends-on` canonicals (and `subjectCanonical`), and the S3 backend has no search, so every SQL View / SQL Query Library is unusable there.

Same root cause on the UI: `/ui/sql/views`, `/ui/sql/queries` and the **Subjects** table of `/ui/sql/export/new` render **empty** (no rail entries, no name field, no error) although the Libraries and ViewDefinitions read fine by id — the listing also goes through search and the failure is swallowed.

## Expected

`MANUAL_TESTING_MATRIX.md` T7: “This step is expected to pass on all eight backends (in-DB on SQLite/Postgres/Mongo, **in-process on S3**)”, and section 1 lists `$sql-run` / `$sql-export` as “in-process scan” on s3. Either:

1. resolve canonicals on S3 by scanning the type's objects and matching `url` (the in-process runner already scans resources this way), and list ViewDefinitions/Libraries for the SQL pages the same way; or
2. at minimum answer `501`/`422` with the underlying message (*capability 'search' not supported by S3 — canonical references cannot be resolved on this backend*) instead of a 500, and mark the SQL View / SQL Query kinds N/A on the s3 row of the matrix.

Found during the #1166 MinIO pass; T7 is recorded ❌ there (ViewDefinition-subject jobs ✅, Library-subject jobs ❌).

angela-helios:
Fixes #1228 (option 1 of the issue: implement the scan).

## What was broken

`HFS_STORAGE_BACKEND=s3` (MinIO, auth off). A ViewDefinition referenced by id ran fine; anything that had to resolve a **canonical** answered 500, and the SQL pages that list definitions were empty:

| case (reproduced on `main` `6f58ff8fa`) | before | after |
|---|---|---|
| `$sql-run` `subjectReference: Library/<sql-view>` | **500** | 200 |
| `$sql-run` `subjectReference: Library/<sql-query>` + `parameters` | **500** | 200 |
| `$sql-run` `subjectResource:` inline sql-view Library | **500** | 200 |
| `$sql-run` `subjectCanonical: …/ViewDefinition/patient_demographics` | **500** | 200 |
| `$sql-export` with a Library subject | **500** | 202 → completed |
| `/ui/sql/views`, `/ui/sql/queries`, `/ui/sql/view-definitions`, Subjects table of `/ui/sql/export/new` | empty | list the 2 ViewDefinitions + 2 Libraries |

Log before: `ERROR helios_rest::error: internal error while processing request error.detail=canonical lookup failed for ViewDefinition url=http://example.org/ViewDefinition/patient_demographics: capability 'search' not supported by S3`.

Two separate paths went through search: `resolve_by_canonical_url` (a `url=` search for `subjectCanonical` and every `relatedArtifact.depends-on`), and the UI's `ConformanceSource::fetch`, which is `GET /{type}?_count=10000` — a 501 on S3 that the pages swallow.

## The change

- **`ResourceStorage::resource_scan()`**, a hook next to `sof_runner()`: a backend without a search index returns the whole-type scan its in-process SOF runner already uses. S3 implements it; every other backend keeps the default `None`.
- **Canonical resolution** (`handlers/sof/references.rs`) still asks search first, so backends that have one are untouched. On `UnsupportedCapability` it scans the type and applies the same `canonical_matches` predicate — newest `meta.lastUpdated` wins, `|version` / `@version` pins honoured. A backend with neither search nor scan now answers **501 with the reason**, never a generic 500 (the fallback the issue asked for, kept as the floor).
- **S3 `search`** serves exactly one thing by scan: the *filterless* listing of the two SQL-on-FHIR definition types, `ViewDefinition` and `Library` — newest first, `_count`/`_offset` honoured, `total` reported. That is what the three pages request, so they need no change. Any filter, and every other type, still answers `UnsupportedCapability`: `GET /Patient` stays 501 (checked), so the matrix's T4 N/A expectation for the s3 row is unchanged.

One behaviour worth knowing: `GET /ViewDefinition?url=…` on S3 now returns the unfiltered list under the default lenient handling, because the REST layer drops `url` as an unknown parameter before the backend sees it (the `self` link shows it was not applied; `Prefer: handling=strict` answers 400 as before). Previously it was a 501 either way.

## Validation on MinIO (16.5k-resource dataset: 11,704 Patients + reference data + the anchor patient's clinical rows)

All **11** T7 `$sql-export` jobs from the #1166 pass now complete; the 7 with a Library subject were the ones failing:

```
vd-ndjson              completed  patient_demographics  11704 lines
query-csv              completed  tall_female_patients  header + 1 row   (min_height bound)
view-parquet           completed  female_patients       5813 rows
all-three-json         completed  3 files (11704 / 1 / 5813)
vd-csv / vd-parquet    completed  11704
query-ndjson / query-parquet  completed  {"id":"7d24f7a0-…","city":"Everett","height":145.3}
view-ndjson            completed  5813 lines
one-patient            completed  1 line
tracked-csv-noheader   completed  5813 lines, clientTrackingId=release-check-01
```

Zero `canonical lookup failed` / 500 lines in the server log afterwards.

## Tests

- `backends::s3::tests::definition_types_list_by_scan_and_everything_else_stays_unsupported` — the listing (total, `_count`, tenant isolation) and that a filtered `ViewDefinition` search and a `Patient` search are still `UnsupportedCapability`.
- `backends::s3::tests::resource_scan_hook_returns_the_tenants_live_resources` — live resources only, tenant-scoped.
- `handlers::sof::references::tests::newest_canonical_match_follows_the_search_paths_rules` — unpinned picks the newest, `|1.0.0` / `@2.0.0` pick that version, unknown version or url → none.

Locally: `cargo fmt --all -- --check` clean; clippy with CI's flags on `helios-persistence` and `helios-rest` (`--all-targets --all-features`) clean; `backends::s3` 117 passed, `handlers::sof` 46 passed.


angela-helios:
**Patch coverage: 46.96% → 88.94%** (`bda25d62b` → `e5d82b33f`), target 85.28%. Gate still green.

What Codecov actually measures here: the Code Coverage job runs `cargo llvm-cov --workspace --features postgres,mongodb`, so `crates/persistence/src/backends/s3/*` is not in the report at all. The patch figure only counts the non-S3 lines of the diff — `crates/rest/src/handlers/sof/references.rs` and the `resource_scan()` default in `core/storage.rs`.

Added in `e5d82b33f`:
- `references.rs`: a scan-only stub storage (search → `UnsupportedCapability`, optional scan) exercising `resolve_by_canonical_url`'s fallback: resolution by url and by pinned version without crossing types, no match → 404, no scan → 501 with a reason, a failing scan → 500 naming the canonical, and any other search error → 500.
- `state.rs`: the default `resource_scan()` hook is `None`.
- `s3/tests.rs` (runs under Test Rust with the `s3` feature, not measured by Codecov): the definition listing now also walks `_offset`, `has_previous` and a page past the end.

Still uncovered in `references.rs` are the lines of the successful-search path (pre-existing code); they are exercised by the SQLite integration tests, not by the lib unit tests.

smunini (approved):

